PR #1, opened 2020-04-24
libbpf-rs: Addressed API code review

danobi:


danobi:
cc @anakryiko .

One thing I'm wondering is if we want an `Object` being dropped to unload the `Program`s or if we want to instead refcount `LoadedMap` and unload the program when all `LoadedMap` instances are dropped. It could probably map cleanly to libbpf b/c there's already a `bpf_program__unload` function.

Not sure if that would be necessary for `LoadedMap`. I'm looking at libbpf code and it doesn't look like there's any closing of map FDs. I'm guessing that's b/c the kernel will hold a refcount as long as the prog is active.

libbpf-rs/src/object.rs
```diff
@@ -86,48 +155,65 @@ impl Map {
...
 
-    /// Same as [`Map::lookup()`] except this also deletes the key from the map.
+    /// Same as [`LoadedMap::lookup()`] except this also deletes the key from the map.
     ///
     /// `key` must have exactly [`Map::key_size()`] elements.
-    pub fn lookup_and_delete(&mut self, _key: &[u8], _opts: MapOptions) -> Result<Option<Vec<u8>>> {
+    pub fn lookup_and_delete(
+        &mut self,
+        _key: &[u8],
+        _flags: LoadedMapFlags,
```
anakryiko, line 164: this is not loaded map flags, it's that operation's flags... Do you intend to make LoadedMapFlags into a builder? Wouldn't plain int be easier to use though?

libbpf-rs/src/object.rs
```diff
@@ -7,57 +7,126 @@ use crate::*;
...
+/// file descriptor.
 ///
 /// Some methods require working with raw bytes. You may find libraries such as
 /// [`plain`](https://crates.io/crates/plain) helpful.
-pub struct Map {}
+#[derive(Clone)]
+pub struct LoadedMap {}
 
-impl Map {
+impl LoadedMap {
```
anakryiko, line 120: I'd imagine that LoadedMap and LoadedProg would be the most common types that user code has to interact with, so I'd try to keep the name as short and and simple, as possible: Map and Prog. What you have as Map and Program right now, it's more of a MapBuilder and ProgramBuilder, no?

libbpf-rs/src/object.rs
```diff
@@ -86,48 +155,65 @@ impl Map {
...
+        unimplemented!();
+    }
+
+    pub fn allow_multi(&mut self) -> &mut Self {
+        unimplemented!();
+    }
+
+    pub fn replace(&mut self) -> &mut Self {
+        unimplemented!();
+    }
```
anakryiko, line 211: this is cgroup attach-specific flags, they are not properties of a program. These would be specified when doing bpf_program__attach_cgroup() in libbpf.

libbpf-rs/src/object.rs
```diff
@@ -139,3 +225,46 @@ pub enum ProgramType {}
...
+    /// Returns a file descriptor to the underlying program.
+    pub fn fd(&self) -> i64 {
+        unimplemented!();
+    }
+
+    pub fn attach_type(&self) -> ProgramAttachType {
+        unimplemented!();
+    }
+
+    pub fn attach_cgroup<T: AsRef<str>>(&mut self, _cgroup: T) -> Result<Link> {
```
anakryiko, line 263: let's keep it simple and do cgroup_fd here? Using strings might be a restricting in some cases and just inconvenient. We can provide helpers to get cgroup_fd from cgroup path (string) or cgroup id (if there is a corresponding api in kernel, don't know).

libbpf-rs/src/object.rs
```diff
@@ -139,3 +225,46 @@ pub enum ProgramType {}
...
+        unimplemented!();
+    }
+
+    pub fn attach_perf_event(&mut self, _pfd: i64) -> Result<Link> {
```
anakryiko, line 267: fd is not going to be bigger than 32-bits, no need for i64 here

libbpf-rs/src/perf_buffer.rs
```diff
@@ -7,44 +7,39 @@ use crate::*;
...
-/// Options and callbacks to configure [`PerfBuffer`].
-///
-/// Some callbacks provide raw byte slices. You may find libraries such as
-/// [`plain`](https://crates.io/crates/plain) helpful.
-#[derive(Default)]
-pub struct PerfBufferOpts {}
-
-impl PerfBufferOpts {
-    pub fn new() -> Self {
+    pub fn new(_map: Map) -> Self {
```
anakryiko, line 10: same as above, I think using map_fd here is more flexible and easier in some cases for user code. Plus, I'd imagine you'd pass LoadedMap here, instead of not-yet-constructred Map, which doesn't yet even have fd

libbpf-rs/src/perf_buffer.rs
```diff
@@ -7,44 +7,39 @@ use crate::*;
...
-    /// Callback arguments are: (ctx, cpu, data).
-    pub fn sample_cb<F>(self, _cb: F) -> Self
+    /// This callback provides a raw byte slice. You may find libraries such as
+    /// [`plain`](https://crates.io/crates/plain) helpful.
+    ///
+    /// Callback arguments are: (cpu, data).
+    pub fn sample_cb<F>(&mut self, _cb: F) -> &mut Self
     where
-        F: FnMut(usize, u32, &[u8]),
+        F: FnMut(u32, &[u8]),
```
anakryiko, line 22: should callback also get a reference to PerfBuffer itself?

libbpf-rs/src/perf_buffer.rs
```diff
@@ -7,44 +7,39 @@ use crate::*;
...
+    pub fn lost_cb<F>(&mut self, _cb: F) -> &mut Self
     where
-        F: FnMut(usize, u32, u64),
+        F: FnMut(u32, u64),
     {
         unimplemented!();
     }
+
+    /// The number of pages to size the ring buffer.
+    pub fn pages(&mut self, _pages: usize) -> &mut Self {
```
anakryiko, line 38: this can actually fail, if pages is not power of two.. though we can just silently round it up, of course... something to consider

libbpf-rs/src/perf_buffer.rs
```diff
@@ -7,44 +7,39 @@ use crate::*;
...
+        unimplemented!();
+    }
+
+    pub fn poll(&mut self, _timeout: Duration) {
```
anakryiko, line 42: can fail, also doesn't need mutable reference, probably

anakryiko:
> cc @anakryiko .
> 
> One thing I'm wondering is if we want an `Object` being dropped to unload the `Program`s or if we want to instead refcount `LoadedMap` and unload the program when all `LoadedMap` instances are dropped. It could probably map cleanly to libbpf b/c there's already a `bpf_program__unload` function.
> 
I think unloading unconditionally. Map and Program are owned by Object, they can't leave outside of Object. libbpf doesn't do anything special here, when object is closed/destroyed, all map and program FDs are closed (see bpf_object__unload()). If map is referenced from some other program, it will be kept around by kernel.

> Not sure if that would be necessary for `LoadedMap`. I'm looking at libbpf code and it doesn't look like there's any closing of map FDs. I'm guessing that's b/c the kernel will hold a refcount as long as the prog is active.

libbpf-rs/src/link.rs
```diff
@@ -1,35 +1,16 @@
...
-    /// file descriptor.
-    pub fn attach_perf_event(_prog: Program, _pfd: i64) -> Result<Self> {
-        unimplemented!();
-    }
-
     /// Replace the underlying prog with `prog`.
-    pub fn update_prog(&mut self, _prog: Program) -> Result<()> {
+    ///
+    /// Returns the replaced [`LoadedProgram`] on success.
+    pub fn update_prog(&mut self, _prog: LoadedProgram) -> Result<LoadedProgram> {
```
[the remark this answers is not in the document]
danobi: sounds good

libbpf-rs/src/object.rs
```diff
@@ -7,57 +7,126 @@ use crate::*;
...
     }
 
-    pub fn with_memory(_name: &str, _mem: &[u8], _opts: ObjectOptions) -> Result<Self> {
+    pub fn from_memory(_name: &str, _mem: &[u8]) -> Result<Self> {
         unimplemented!();
     }
 
-    pub fn name(&self) -> &str {
+    /// Override the generated name that would have been inferred from the constructor.
+    pub fn name<T: AsRef<str>>(&mut self, _name: T) -> &mut Self {
```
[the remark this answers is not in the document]
danobi: ok

libbpf-rs/src/object.rs
```diff
@@ -7,57 +7,126 @@ use crate::*;
...
         unimplemented!();
     }
 
-    pub fn maps(&self) -> Vec<&Map> {
+    /// Option to parse map definitions non-strictly, allowing extra attributes/data
+    pub fn relaxed_maps(&mut self) -> &mut Self {
```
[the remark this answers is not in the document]
danobi: ok

libbpf-rs/src/object.rs
```diff
@@ -7,57 +7,126 @@ use crate::*;
...
+
+    pub fn btf_fd(&mut self, _fd: u32) -> &mut Self {
+        unimplemented!();
+    }
+
+    pub fn btf_key_type_id(&mut self, _id: u32) -> &mut Self {
+        unimplemented!();
+    }
+
+    pub fn btf_value_type_id(&mut self, _id: u32) -> &mut Self {
```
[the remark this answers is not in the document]
danobi: ok

libbpf-rs/src/object.rs
```diff
@@ -7,57 +7,126 @@ use crate::*;
...
+        unimplemented!();
+    }
+
+    pub fn no_prealloc(&mut self) -> &mut Self {
+        unimplemented!();
+    }
+
+    pub fn no_common_lru(&mut self) -> &mut Self {
+        unimplemented!();
+    }
```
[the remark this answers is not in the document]
danobi: let's use bitflags crate

libbpf-rs/src/object.rs
```diff
@@ -7,57 +7,126 @@ use crate::*;
...
+
+    // TODO: more flags here:
+    // https://github.com/torvalds/linux/blob/master/include/uapi/linux/bpf.h#L340-L341
+
+    pub fn load(&mut self) -> Result<LoadedMap> {
```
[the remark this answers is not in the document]
danobi: The intent is user should call this to load the map. User creates an `Object` struct and loads the maps and progs they want. I believe that's what we discussed but if I misremember we can change it.

libbpf-rs/src/object.rs
danobi: Yeah, let's use bitflags

libbpf-rs/src/object.rs
danobi: ok

libbpf-rs/src/object.rs
danobi: ok

libbpf-rs/src/object.rs
danobi: IIRC cgroup inode is used in some places. Not sure; never used it.

libbpf-rs/src/object.rs
danobi: ok

libbpf-rs/src/perf_buffer.rs
danobi: Let's go with `Map` (post-rename from `LoadedMap` -> `Map`) for now. We can add a `from_fd()` method when we have more FD based APIs.

libbpf-rs/src/perf_buffer.rs
danobi: Sure, why not. Might be handy later.

libbpf-rs/src/perf_buffer.rs
danobi: ok

libbpf-rs/src/perf_buffer.rs
danobi: If callbacks receive a mutable ref to PerfBuffer, I think it makes sense to keep this as mutable.

danobi:
> I think unloading unconditionally. Map and Program are owned by Object, they can't leave outside of Object.

Ok. We can make `Object` the guy that owns everything.

libbpf-rs/src/object.rs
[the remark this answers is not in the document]
anakryiko: Nothing beats A | B | C in terms of succinctness, but I'm fine with whatever Rust recommended solution :)

libbpf-rs/src/perf_buffer.rs
anakryiko: yeah, was thinking about Map::from_fd() constructor, you'll need to do GET_OBJ_INFO to get back map name, key/value sizes, etc, but it should work as well.

libbpf-rs/src/perf_buffer.rs
anakryiko: do callbacks need mutable reference? what for?

libbpf-rs/src/object.rs
```diff
@@ -16,108 +18,93 @@ impl Object {
...
         unimplemented!();
     }
 
-    pub fn btf_value_type_id(&mut self, _id: u32) -> &mut Self {
+    pub fn set_max_entries(&mut self, _entries: u32) -> &mut Self {
         unimplemented!();
     }
 
-    pub fn map_ifindex(&mut self, _idx: u32) -> &mut Self {
+    pub fn set_initial_value(&mut self, _data: &[u8]) -> &mut Self {
```
anakryiko, line 58: this can fail for many reasons: not supported map type, invalid data size, etc

libbpf-rs/src/perf_buffer.rs
```diff
@@ -16,30 +16,30 @@ impl PerfBuffer {
...
     where
-        F: FnMut(u32, u64),
+        F: FnMut(&mut PerfBuffer, u32, u64),
     {
         unimplemented!();
     }
 
     /// The number of pages to size the ring buffer.
-    pub fn pages(&mut self, _pages: usize) -> &mut Self {
+    pub fn pages(&mut self, _pages: usize) -> Result<&mut Self> {
```
anakryiko, line 38: set_pages? Also, should be go with the same PerfBufferBuilder + PerfBuffer split as with Map and Program?

libbpf-rs/src/object.rs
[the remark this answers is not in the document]
anakryiko: just checked bitflags usage examples, I think that's exactly what we need, thanks!

libbpf-rs/src/perf_buffer.rs
danobi: Do we need a reference back to the `PerfBuffer` at all? We don't currently have any properties that can really be changed after construction.

If we wanna add it later we can easily add a `poll_mut()` or `poll_ref()` method. But TBH I can't see a use case.

libbpf-rs/src/object.rs
danobi: Hmm but this is a builder method. Wouldn't it be more consistent for the interface to fail in `load()` later?

libbpf-rs/src/perf_buffer.rs
danobi: Yeah, split makes sense

libbpf-rs/src/perf_buffer.rs
anakryiko: yeah, I don't have specific use case to have any kind of references, but it's usually a good idea to have reference to an "object" that's calling a callback function. E.g., just in case someone has multiple PerfBuffers, but the same callback. Having reference to PerfBuffer would allow to differentiate between them, if necessary. 

But I think we are talking about different things here. PerfBuffer::poll needs &self -- I don't think PerfBuffer needs to be modified as a result of poll operation. What I was describing above is what should be passed to user-provided callback. There, I think, having PerfBuffer reference is useful.

libbpf-rs/src/object.rs
anakryiko: Ah, yeah, I guess for builder it's ok. You'd need to copy _data internally, though, right? Kind of sucks, but probably not a big deal. Having load() failing at the end will make it a bit harder for user to understand what specifically failed, but I think it's fine. Let's make only load() failing.

libbpf-rs/src/object.rs
```diff
@@ -89,45 +145,54 @@ impl Map {
     /// Same as [`Map::lookup()`] except this also deletes the key from the map.
     ///
     /// `key` must have exactly [`Map::key_size()`] elements.
-    pub fn lookup_and_delete(&mut self, _key: &[u8], _opts: MapOptions) -> Result<Option<Vec<u8>>> {
+    pub fn lookup_and_delete(&mut self, _key: &[u8], _flags: MapFlags) -> Result<Option<Vec<u8>>> {
```
anakryiko, line 148: btw, lookup can take BPF_F_LOCK flag, so I think lookup() needs to take flags as well

libbpf-rs/src/object.rs
```diff
@@ -139,3 +204,55 @@ pub enum ProgramType {}
...
+    pub fn section(&self) -> &str {
+        unimplemented!();
+    }
+
+    pub fn prog_type(&self) -> ProgramType {
+        unimplemented!();
+    }
+
+    /// Returns a file descriptor to the underlying program.
+    pub fn fd(&self) -> i64 {
```
anakryiko, line 234: fd is u32?

libbpf-rs/src/object.rs
```diff
@@ -1,63 +1,119 @@
...
+#[derive(Clone)]
 pub struct Map {}
 
 impl Map {
     pub fn name(&self) -> &str {
         unimplemented!();
     }
 
+    /// Returns a file descriptor to the underlying map.
+    pub fn fd(&self) -> i64 {
```
anakryiko, line 113: fd is u32?

libbpf-rs/src/object.rs
danobi: ok

libbpf-rs/src/object.rs
danobi: ok

libbpf-rs/src/object.rs
danobi: ok

libbpf-rs/src/perf_buffer.rs
danobi: Sounds reasonable

danobi:
I'm gonna merge this and start implementing everything. Feel free to leave more comments on here if you see anything later.

I'll put up a PR for the implementation as well.

anakryiko:
@danobi, sounds good, I think currently it looks good. You'll probably find some inconsistencies or missing things along the way, but nothing is set in stone and we can adjust and revisit as needed. Thanks for working on this!

PR #3, opened 2020-06-10
Refactor API to implement composite loading

danobi:
The previous model allowed individual loading of progs and maps (or at
least pretended to allow it). This doesn't really work in libbpf.

The new model will be the following:
```
            from_*()        load()
              |               |
              v               v
ObjectBuilder -> OpenedObject -> Object
                      ^            ^
                      |            |
          <pre-load modifications> |
                                   |
                            <post-load interactions>
```
All handles returned from `OpenedObject` and `Object` will be references
to ensure the handles can't outlive the lifetime of the containing
object.


libbpf-rs/src/object.rs
```diff
@@ -8,13 +8,13 @@ use std::path::Path;
...
 use bitflags::bitflags;
-use nix::{errno, fcntl, unistd};
+use nix::errno;
 use num_enum::TryFromPrimitive;
 
 use crate::util;
 use crate::*;
 
-/// Sets options for opening a [`Object`]
+/// Builder for creating an [`OpenedObject`]. Typically the entry point into libbpf-rs.
```
anakryiko, line 17: not a native English speaker, but I think it's not correct to use "opened" as an adjective... OpenObject is shorter as well :)

libbpf-rs/src/object.rs
```diff
@@ -152,17 +140,44 @@ impl Default for ObjectBuilder {
...
-/// Represents a BPF object file. An object may contain zero or more
-/// [`Program`]s and [`Map`]s.
-pub struct Object {
+fn find_map_in_object(
+    obj: *const libbpf_sys::bpf_object,
+    name: &str,
+) -> Result<Option<*mut libbpf_sys::bpf_map>> {
+    let c_name = util::str_to_cstring(name)?;
+    let ptr = unsafe { libbpf_sys::bpf_object__find_map_by_name(obj, c_name.as_ptr()) };
+    if ptr.is_null() {
```
anakryiko, line 149: Rust offtopic: I wonder if having generic Null->Option generic conversion routine would be possible?

libbpf-rs/src/object.rs
```diff
@@ -152,17 +140,44 @@ impl Default for ObjectBuilder {
...
+
+/// Represents an opened (but not yet loaded) BPF object file.
+///
+/// Use this object to access [`MapModifier`]s and [`ProgramModifier`]s.
+pub struct OpenedObject {
     ptr: *mut libbpf_sys::bpf_object,
-    maps: HashMap<String, MapBuilder>,
-    progs: HashMap<String, ProgramBuilder>,
+    maps: HashMap<String, MapModifier>,
+    progs: HashMap<String, ProgramModifier>,
```
anakryiko, line 175: wouldn't OpenMap and OpenProgram be consistent with OpenObject?

libbpf-rs/src/object.rs
```diff
@@ -183,40 +198,135 @@ impl Object {
...
+            Ok(None)
+        }
+    }
+
+    // Same as [`map`] except will panic if `Err` or `None` is encountered.
+    pub fn map_unwrap<T: AsRef<str>>(&mut self, name: T) -> &mut MapModifier {
+        self.map(name).unwrap().unwrap()
+    }
+
+    pub fn prog<T: AsRef<str>>(&mut self, name: T) -> Result<Option<&mut ProgramModifier>> {
```
anakryiko, line 218: For prog, lookup by name is a bit more ambiguous than for map. Maps only have name, which correspond to their respective variable name in C code.

But programs actually have section name, associated with them (which is what libbpf calls "title", see bpf_object__find_program_by_title) and then the name of C function (that's what bpf_object__find_program_by_name uses). Should we have both look up methods?

Name is probably good enough for starters... Don't have strong opinion. In libbpf/C land, historically most uses are lookups by title (section name), but I think overall lookup by name is better in the long run (I think eventually libbpf will allow to have multiple independent functions with the same section name). So just consider what would you do if someone requested lookup by section name.

PR #6, opened 2020-06-12
PerfBufferBuilder: use concrete types for ::new

mokomull:
This allows PerfBufferBuilder to build when the user does not specify
one of the callbacks, because there is no longer any ambiguity about
function type to infer if one is missing.

The mutators also now have to modify the type, as they replace the
callback function type on-demand -- they now take self by move, and
return a new PerfBufferBuilder with the right types.

### DIFF
--- a/example/src/main.rs
+++ b/example/src/main.rs
@@ -109,10 +109,11 @@ fn main() {
     println!("{:8} {:16} {:7} {:14}", "TIME", "COMM", "TID", "LAT(us)");
 
     let events = obj.map_unwrap("events");
-    let mut perf_builder = PerfBufferBuilder::new(events);
-    perf_builder.set_sample_cb(handle_event);
-    perf_builder.set_lost_cb(handle_lost_events);
-    let perf = perf_builder.build().unwrap();
+    let perf = PerfBufferBuilder::new(events)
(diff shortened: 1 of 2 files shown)


danobi:
Awesome, thanks

anakryiko:
I don't care much about set_xxx vs with_xxx naming convention for builders, but let's keep them consistent throughout library?

danobi:
I'm looking at https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344] and see:

> Note that this convention is about getters/setters on ordinary data types, not on builder objects.

about `set_foo()` for setting and `foo()` for getting.

Looking at https://doc.rust-lang.org/1.0.0/style/ownership/builders.html , the example they give they omit the `set_` and `with_` prefix. Maybe we should get rid of the prefix altogether.

What do you think?

anakryiko:
If there are no getters for builder, then yeah, having no set_ or with_ prefix is fine. So let's drop them then.

mokomull:
Yeah, I kind of assumed `with_` was the builder equivalent of `into_` but it appears that isn't really a commonly-used case.  I'll send a follow-up PR shortly and drop the prefix entirely.

PR #7, opened 2020-06-15
libbpf-rs: Remove (set|with)_* prefix on builder arg methods

danobi:


libbpf-rs/src/object.rs
```diff
@@ -73,7 +73,7 @@ impl ObjectBuilder {
         }
     }
 
-    pub fn from_path<P: AsRef<Path>>(&mut self, path: P) -> Result<OpenObject> {
+    pub fn path<P: AsRef<Path>>(&mut self, path: P) -> Result<OpenObject> {
```
danobi, line 76: Felt a bit weird to have the terminal methods be just `path` and `memory`, but it looks like the std library does this too: https://doc.rust-lang.org/std/process/struct.Command.html#method.status

anakryiko: it is indeed. The biggest problem is that it looks like just another setter, while it is really a build method. Looking at PerfBuffer, there is a clear .build() call. Here I'd say that it should either be `open_from_path` (I'd go with just `open_file`, actually, because that's what we are doing we do not expect to open from directory, pointed to by `path`, right?) or make `path()` purely a setter and have dedicated `build()`/`open()` call. The latter is a bit problematic for building from memory, because will require a potentially pretty big copy, maintained internally in a builder.

So with all that, I'd just do `open_file()` and `open_mem()` as two build methods.

PR #10, opened 2020-07-09
libbpf-rs: Implement query API

danobi:
This commit implements an iterator for prog, map, and btf information.
The code was all pretty similar so I used a `macro_rules!` to generate
it. This makes it easy to add new iterators if libbpf (the C library)
follows the same convention for future query APIs.

### DIFF
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -1,0 +1,86 @@
+//! Query the host about BPF
+//!
+//! For example, to list the name of every bpf program running on the system:
+//! ```
+//! use std::str::from_utf8;
+//! use libbpf_rs::query::ProgInfoIter;
+//!
+//! let mut iter = ProgInfoIter::default();
+//! for prog in iter {
+//!     let converted_arr: Vec<u8> = prog.name
+//!         .iter()
+//!         .map(|x| *x as u8)
+//!         .collect();
+//!
+//!     println!("{}", from_utf8(&converted_arr).unwrap());
+//! }
+//! ```
+
+use core::ffi::c_void;
+use std::mem::size_of;
+
+use nix::{errno, unistd::close};
+
+macro_rules! gen_info_impl {
+    // This magic here allows us to embed doc comments into macro expansions
+    ($(#[$attr:meta])*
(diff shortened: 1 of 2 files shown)


danobi:
Oh, in my rust to put this up I forgot to actually test it. Will test tomorrow (already have the doctest program)

libbpf-rs/src/query.rs
```diff
@@ -0,0 +1,86 @@
...
+
+            }
+        }
+    };
+}
+
+gen_info_impl!(
+    /// Iterator that returns information about bpf programs currently running on the system.
+    ProgInfoIter,
+    libbpf_sys::bpf_prog_info,
```
anakryiko, line 67: is exposing raw kernel UAPI struct the best idea here? Wouldn't Rust structs for each of bpf_prog_info/bpf_map_info/btf_btf_info provide a bit better user experience?

danobi: Yeah, you're right. Originally I thought that since most of the fields were numbers it'd be fine to expose the uapi. But when writing the example it was annoying to handle the strings and convert enums from int to actual rust enum type.

libbpf-rs/src/query.rs
```diff
@@ -0,0 +1,249 @@
...
+        impl Iterator for $name {
+            type Item = $info_ty;
+
+            fn next(&mut self) -> Option<Self::Item> {
+                if unsafe { $next_id(self.cur_id, &mut self.cur_id) } != 0 {
+                    return None;
+                }
+
+                let fd = unsafe { $fd_by_id(self.cur_id) };
+                if fd < 0 && errno::errno() == errno::Errno::ENOENT as i32 {
```
anakryiko, line 42: I don't think this is the right algorithm. Just because this ID is no more, doesn't mean there are no more IDs to iterate over. See https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/link.c#L273-L294 for how bpftool iterates, I think that one is correct.

danobi: Whoops, yeah. Copied it wrong

danobi:
Will squash when merge

PR #13, opened 2020-07-09
examples: Add query example

danobi:


danobi:
Blocked on #10 and #12

libbpf-rs/src/query.rs
```diff
@@ -0,0 +1,249 @@
+//! Query the host about BPF
```
anakryiko, line 1: this file is not supposed to be in this PR, right?

examples/query/src/main.rs
```diff
@@ -1,3 +1,63 @@
...
+        println!(
+            "name={:<16} type={:<15} run_count={:<2} runtime_ns={}",
+            prog.name,
+            prog.ty.to_string(),
+            prog.run_cnt,
+            prog.run_time_ns
+        );
+    }
+
+    0
```
anakryiko, line 29: why all these zeroes? look weird and it's the only value you can return here (btf, how can iterators communicate failure?) If it's an example, I'd keep it to bare minimum  and just return 0 in main().

libbpf-rs/src/query.rs
danobi: It's b/c this PR is stacked on top of the query PR which hasn't landed yet. Once the lower PRs land it'll just be the example code. I left it in draft mode b/c of that

examples/query/src/main.rs
danobi: Could communicate errors by having iterator return `Result` type. But I think for the libbpf query interface it should be ok to return bare values. I don't see any errors that don't terminate iteration.

anakryiko: my point was that when one of those kernel APIs does return real error, how does iterator communicate that iteration process stopped due to error, not because we finished iterating all of the items? It's a general question about Rust iteration, how to distinguish premature iteration end due to error vs completed iteration?

danobi: I'm not sure there's a great way. I guess one way to do that would have a constructor that returns a `Result<*InfoIter>`. But then you'd have to iterate everything ahead of time and store the data instead of get it on demand.

PR #16, opened 2020-07-21
Support links in query API

### LINKED ISSUE #14: Add query API for link
https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L221

danobi:
This closes #14 .

### DIFF
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -66,13 +66,14 @@ macro_rules! gen_info_impl {
                 let mut len = size_of::<$uapi_info_ty>() as u32;
 
                 let ret = unsafe { libbpf_sys::bpf_obj_get_info_by_fd(fd, item_ptr as *mut c_void, &mut len) };
-                let _ = close(fd);
-                if ret != 0 {
-                    return None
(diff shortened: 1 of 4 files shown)


libbpf-rs/src/query.rs
```diff
@@ -265,3 +266,105 @@ gen_info_impl!(
...
+    pub info: LinkTypeInfo,
+    pub id: u32,
+    pub prog_id: u32,
+}
+
+impl LinkInfo {
+    fn from_uapi(fd: i32, mut s: libbpf_sys::bpf_link_info) -> Option<Self> {
+        let type_info = match s.type_ {
+            libbpf_sys::BPF_LINK_TYPE_RAW_TRACEPOINT => {
+                if unsafe { s.__bindgen_anon_1.raw_tracepoint.tp_name } == 0 {
```
anakryiko, line 308: you can be pretty sure that this is always going to be true. bpftool uses this check to get out of the infinite loop, but they way you implemented this there is no possibility of a loop, so you can slightly simplify this logic, just always specify tp_name and do obj_get_info() again to get details.

danobi:
Removed the check

PR #17, opened 2020-08-06
Don't assume the width and signedness of C types

tchebb:
In several places, we use a Rust type of `u64` to represent a C type that is actually pointer sized. This causes compilation errors when targeting 32-bit platforms. Similarly, we assume that `std::os::raw::c_char` is an `i8`, when in fact it can be either an `i8` or a `u8` depending on the platform we're targeting. Fix this by adding casts and using the `usize` Rust type where appropriate.

I don't claim that this commit fixes all such issues, but it does at least fix the ones that were preventing us from building for 32-bit ARM with musl libc. I have also confirmed that it does not regress compilation of a simple demo program on x86_64.


danobi:
Thanks!

PR #19, opened 2020-08-10
libbpf-rs: Add pin/unpin API to Map/Program

danobi:
Pinning/unpinning maps/progs is useful for networking applications.


danobi:
Yeah, will do in another PR

PR #22, opened 2020-08-19
libbpf-rs: add attach_sockmap for Program

gdetal:
This is required to support verdict/parser of SK_SKB programs. The programs must be attached to SOCKMAP/SOCKHASH for them to work.

References:
 - https://lwn.net/Articles/731133/
 - https://blog.cloudflare.com/sockmap-tcp-splicing-of-the-future/
 - https://blog.bolcsfoldi.com/posts/ebpf-sockmap/

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -327,4 +327,15 @@ impl Program {
             Ok(Link::new(ptr))
         }
     }
+
+    /// Attach a verdict/parser to a [sockmap/sockhash](https://lwn.net/Articles/731133/)
+    pub fn attach_sockmap(&self, map_fd: i32) -> Result<()> {
+        let err =
+            unsafe { libbpf_sys::bpf_prog_attach(self.fd(), map_fd, self.attach_type() as u32, 0) };
+        if err != 0 {
+            Err(Error::System(err as i32))
+        } else {
+            Ok(())
+        }
+    }
 }


libbpf-rs/src/program.rs
```diff
@@ -327,4 +327,15 @@ impl Program {
             Ok(Link::new(ptr))
         }
     }
+
+    /// Attach a verdict/parser to a [sockmap/sockhash](https://lwn.net/Articles/731133/)
+    pub fn attach_sockmap(&self, map_fd: i32) -> Result<()> {
```
danobi, line 332: Does it work to pass a map? Would be nicer to do a Map object if it doesn't cause borrowing issues (need 2 mutable borrows from Object)
```suggestion
    pub fn attach_sockmap(&self, map: &Map) -> Result<()> {
```

gdetal: It would only work if `Map` implements `Clone` which is not the case now. I don't know enough about the project to see whether this is OK.

Here is a potential code snippet with `clone()`:

```rust
        let sockmap = obj.map("sock_map")?.ok_or("unknown map: sock_map")?.clone();
        obj.prog("handle__verdict")?
            .ok_or("unknown program: handle__verdict")?
            .attach_sockmap(&sockmap)?;
        obj.prog("handle__parser")?
            .ok_or("unknown program: handle__parser")?
            .attach_sockmap(&sockmap)?;
```

The previous one I used with the file descriptor:

```rust
        let sockmap = obj.map("sock_map")?.ok_or("unknown map: sock_map")?.fd();
        obj.prog("handle__verdict")?
            .ok_or("unknown program: handle__verdict")?
            .attach_sockmap(sockmap)?;
        obj.prog("handle__parser")?
            .ok_or("unknown program: handle__parser")?
            .attach_sockmap(sockmap)?;
```

Let me know which one is best and I'll adapt my PR.

danobi: `Clone` for `Map` seems dangerous b/c a cloned Map would alias the underlying pointer and potentially cause safety issues. Let's just go with the fd-based approach b/c I can't think of anything better right now

danobi:
cc @anakryiko ; will merge in a bit if no objections

PR #26, opened 2020-08-26
libbpf-rs: Return correct errno for attach_sockmap

danobi:
bpf_prog_attach() has the following code path:

    bpf_prog_attach(..)
      bpf_prog_attach_xattr(..)
        sys_bpf(..)
          syscall(__NR_bpf, ..)

syscall(2) returns -1 on error and stores errno in `errno`. So I think
we need to explicitly get errno() for attach_sockmap().


danobi:
cc @gdetal.

Gonna go ahead and merge b/c I'm pretty sure I'm right but if I'm not, let me know and we can change it again.

anakryiko:
Yes, this is the right approach. All low-level BPF APIs are just a thin wrappers around syscall(), which always set errno on error. So using errno as a return code is correct.

PR #29, opened 2020-11-20
Remove docs link for `libbpf-cago`

frewsxcv:
Rustdoc doesn't generate docs for binary crates


danobi (approved):
Darn, at one point they did.

PR #33, opened 2020-12-03
libbpf-rs: map: Add set_max_entries() helper

danobi:


libbpf-rs/src/map.rs
```diff
@@ -45,6 +45,17 @@ impl OpenMap {
         Ok(())
     }
 
+    pub fn set_max_entries(&mut self, max: u32) -> Result<()> {
```
anakryiko, line 48: nit: max -> nr_entries or cnt or something like that?

anakryiko (approved):
looks good, one naming nit, feel free to ignore

PR #34, opened 2020-12-03
cargo-libbpf: make: Add `make` subcommand 

danobi:
The `cargo libbpf make` subcommand runs:

* cargo libbpf build
* cargo libbpf gen
* cargo build

in sequence, bailing at the first error. The idea is to help prevent
users from forgetting a step. If a step is forgotten, the end results
are unpredictable.

### DIFF
--- a/libbpf-cargo/src/main.rs
+++ b/libbpf-cargo/src/main.rs
@@ -6,6 +6,7 @@ use structopt::StructOpt;
 #[doc(hidden)]
 mod build;
 mod gen;
+mod make;
 mod metadata;
 #[cfg(test)]
 mod test;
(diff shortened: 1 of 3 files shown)


libbpf-cargo/src/main.rs
```diff
@@ -81,6 +82,23 @@ enum Command {
         /// Path to top level Cargo.toml
         manifest_path: Option<PathBuf>,
     },
+    Make {
+        #[structopt(short, long)]
+        debug: bool,
+        #[structopt(long, parse(from_os_str))]
+        /// Path to top level Cargo.toml
+        manifest_path: Option<PathBuf>,
+        #[structopt(long, parse(from_os_str), default_value = "/bin/clang")]
```
anakryiko, line 91: if you specify default_value just "clang", would that work? Not all setups will have /bin/clang exactly there...

danobi: Yeah, it should search PATH. I can do it as a follow up. There's another place where it's /bin/clang

PR #35, opened 2020-12-04
libbpf-cargo: make: skel: Add tests for `gen` and `make`

danobi:
This PR adds a couple tests for skeleton generation and the `cargo
libbpf make` subcommand.

We test the skeleton by:

* generating a skeleton for basic progs
* generate a main.rs that uses the exposed APIs
* cargo-build the project

Note these tests don't require root b/c we're only compiling, not
running. This should be sufficient to test skeleton correctness.


danobi:
Stacked on top of #34

PR #39, opened 2020-12-10
libbpf-cargo: gen: Implement skeleton using C API

danobi:
This PR implements the rust skeleton using the libbpf C APIs.

This is necessary to obtain the `(struct bpf_map *)->mmaped` field.
It also simplifies a few things (at the cost of complicating a few others).

Generated code diff: http://ix.io/2HCt .

See individual commits for full details.


runqslower.skel.rs-DELETEME
```diff
@@ -0,0 +1,234 @@
...
+static mut OBJECT: *mut libbpf_sys::bpf_object = std::ptr::null_mut();
+static mut MAP_START: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
+static mut MAP_MIN_US: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
+static mut MAP_EVENTS: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
+static mut PROG_HANDLE__SCHED_WAKEUP: *mut libbpf_sys::bpf_program = std::ptr::null_mut();
+static mut PROG_HANDLE__SCHED_WAKEUP_LINK: *mut libbpf_sys::bpf_link = std::ptr::null_mut();
+static mut PROG_HANDLE__SCHED_WAKEUP_NEW: *mut libbpf_sys::bpf_program = std::ptr::null_mut();
+static mut PROG_HANDLE__SCHED_WAKEUP_NEW_LINK: *mut libbpf_sys::bpf_link = std::ptr::null_mut();
+static mut PROG_HANDLE__SCHED_SWITCH: *mut libbpf_sys::bpf_program = std::ptr::null_mut();
+static mut PROG_HANDLE__SCHED_SWITCH_LINK: *mut libbpf_sys::bpf_link = std::ptr::null_mut();
```
anakryiko, line 18: how badly you need to use statics?... That makes creating two skeletons in parallel impossible, no? Why this limitation? I.e., can't the skel config have enough storage for these pointers?

danobi: > That makes creating two skeletons in parallel impossible, no? 

Do you mean if the user tries to create two `{Open,}FooSkeleton` instances? If yeah, then this would cause issues.

I suppose it's possible to allocate pointers on the heap in `ObjectSkeletonConfig` and then access pointers in .skel.rs via checked array index. Would make things cleaner too.

anakryiko: Yes, you can create many independent skeletons with C skeleton, we should probably keep that property here. Not sure you need heap allocation, though. You know at the time of generation how many maps, progs (and links) you need, so code-generated skeleton could just have a fixed-size array of pointers, no? and then yes, just use indices to access them.

danobi: > could just have a fixed-size array of pointers, no?

I don't think that'll work. All rust structs must be `memmove`-ready -- that's how move semantics are implemented (unless compiler RVOs). It's possible to avoid a heap allocation, but then you'd need to [`Pin`](https://doc.rust-lang.org/stable/std/pin/struct.Pin.html) the `ObjectSkeletonConfig` to prevent instances from being moved. That'd be pretty unergonomic so I think allocating on heap is easier.

anakryiko: Oh, ok. Sure, whichever way works. I hoped arrays would be the simplest solution, but it's fine.

danobi: I guess you could allocate `ObjectSkeletonConfig`s on the heap to avoid the issue. But I still think you need some `Pin` stuff to avoid callers from moving `ObjectSkeletonConfig` out of the box (which could trigger a memmove).

danobi:
Updated to store pointers inside `ObjectSkeletonConfig`

libbpf-cargo/src/gen.rs
```diff
@@ -83,6 +107,74 @@ fn get_prog_name(prog: *const libbpf_sys::bpf_program) -> Result<String> {
...
+            "false"
+        };
+
+        write!(
+            skel,
+            r#"
+            .map("{raw_name}", {mmaped})
+            "#,
+            raw_name = raw_name,
+            mmaped = mmaped,
```
anakryiko, line 149: just curious, what write!() will write if you pass `bool` directly without "false"/"true" conversion? It capitalizes "True"/"False"?

anakryiko (approved):
looks good, thanks!

libbpf-cargo/src/gen.rs
danobi: Compile error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=06300b1a9b847065c1126116f4e7b92f

anakryiko: I think it's supposed to work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f26a3b1fd6c45d3a627280b579882a08 See https://doc.rust-lang.org/beta/std/macro.write.html

anakryiko: specifically for bool, it seems like it does what I expected it to: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=621f1d7191930c55efb2438adbb207ea

danobi: Interesting. A magic trait for everything...

PR #42, opened 2020-12-16
BPF programs will not run unless there is a map existing.

masmullin2000:
This change allows BPF programs without maps.


danobi (approved):
Oops, nice catch.

PR #43, opened 2020-12-18
libbpf-cargo: gen: Add datasec mmap support

danobi:
This PR adds support for mmap'd datasec fields. Userspace can now
share data with progs via simple variable reads/writes. See the final
commit for the user facing interface.

See individual commits for details.

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -1,6 +1,8 @@
 use std::cmp::{max, min};
+use std::collections::BTreeSet;
 use std::convert::TryFrom;
 use std::ffi::{c_void, CStr, CString};
+use std::fmt::Write;
 use std::mem::size_of;
 use std::slice;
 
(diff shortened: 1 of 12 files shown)


anakryiko (commented):
Looks good. I think there is no need to parse ELF in Rust, given libbpf handles all that and allows to get back raw BTF data in whatever endianness you need. I skimmed through the Rust parts, didn't go too deep into all the details, but the end-user UX looks great!

libbpf-cargo/src/btf/btf.rs
```diff
@@ -0,0 +1,737 @@
...
+impl<'a> Btf<'a> {
+    pub fn new(name: &str, object_file: &[u8]) -> Result<Option<Self>> {
+        let elf = File::parse(object_file)?;
+        ensure!(elf.format() == BinaryFormat::Elf, "Object file not ELF");
+
+        let endian = if elf.is_little_endian() {
+            scroll::LE
+        } else {
+            scroll::BE
+        };
```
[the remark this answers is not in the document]
danobi: Ah ok. Will use btf__pointer_size().

btf__set_endianness() was added in libbpf v0.2 and libbpf_sys is stil lon libbpf v0.1 so can't use that yet. I put up https://github.com/alexforster/libbpf-sys/pull/14 to update libbpf_sys. I'll get rid of ELF parsing when 0.2 update goes in.

danobi:
Updated with suggested changes

libbpf-cargo/src/btf/btf.rs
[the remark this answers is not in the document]
danobi: Ok, fixed

libbpf-cargo/src/btf/btf.rs
```diff
@@ -0,0 +1,739 @@
...
+        obj_opts.sz = std::mem::size_of::<libbpf_sys::bpf_object_open_opts>() as libbpf_sys::size_t;
+        obj_opts.object_name = cname.as_ptr();
+        let bpf_obj = unsafe {
+            libbpf_sys::bpf_object__open_mem(
+                object_file.as_ptr() as *const c_void,
+                object_file.len() as u64,
+                &obj_opts,
+            )
+        };
+        if bpf_obj.is_null() {
```
anakryiko, line 35: libbpf doesn't really return NULL on error, rather encoded error code as a pointer. You need to use libbpf_get_error() to check if it was an error or not.

libbpf-cargo/src/btf/btf.rs
```diff
@@ -0,0 +1,739 @@
...
+            bail!("Failed to bpf_object__open_mem()");
+        }
+
+        let bpf_obj_btf = unsafe { libbpf_sys::bpf_object__btf(bpf_obj) };
+        if bpf_obj_btf.is_null() {
```
anakryiko, line 40: similarly as for bpf_object, NULL is never returned on error (I want to change that in 1.0, but you need to use libbpf_get_error() for now)

anakryiko (approved):
apart from the seemingly incorrect error handling for bpf_object__open and btf__parse/btf__new, looks good. Please use libbpf_get_error().

libbpf-cargo/src/btf/btf.rs
danobi: Good catch, thanks

libbpf-cargo/src/btf/btf.rs
danobi: Does bpf_object__btf() return null on errors? https://github.com/libbpf/libbpf/blob/e05f9be4f4bed72cd2c4942bab972da8d693df08/src/libbpf.c#L8172-L8175

If we used libbpf_get_error() on a null, wouldn't it return no error?

anakryiko: oops, I misread bpf_object__btf as `btf__parse`/`btf__new` constructors. Ignore this one, it will return NULL, if there is no BTF.

PR #45, opened 2020-12-23
libbpf-cargo: gen: Add --object option for standalone skel generation

danobi:
Generates and prints the skeleton to stdout for a given object file.
This is useful for integrating with non-cargo build systems.

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -4,7 +4,7 @@ use std::ffi::{c_void, CStr, CString};
 use std::fmt::Write as fmt_write;
 use std::fs::File;
 use std::io::Write;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::ptr;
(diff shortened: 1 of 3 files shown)


libbpf-cargo/src/gen.rs
```diff
@@ -490,29 +496,31 @@ fn gen_skel_contents(_debug: bool, obj: &UnprocessedObj) -> Result<String> {
...
-        obj_name += &obj.name[1..];
+    // Unwrap is safe b/c already checked that `raw_obj_name` contains chars
+    obj_name += &raw_obj_name
+        .chars()
+        .next()
+        .unwrap()
+        .to_uppercase()
+        .to_string();
+    if raw_obj_name.len() > 1 {
+        obj_name += &raw_obj_name[1..];
```
anakryiko, line 517: you've added capitalize_first_letter in https://github.com/libbpf/libbpf-rs/pull/43, so once that one is merged you can use it here, right?

anakryiko (approved):

PR #50, opened 2021-01-05
libbpf-cargo: gen: Embed object file as a literal array

danobi:
More reliable in general, especially in environments where build
artifacts may be cached and absolute paths introduce nondeterminism.

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -626,14 +626,6 @@ fn gen_skel_contents(_debug: bool, raw_obj_name: &str, obj_file_path: &Path) -> 
         "#
     )?;
 
-    write!(
-        skel,
-        r#"
-        const DATA: &[u8] = include_bytes!("{}");
-        "#,
-        obj_file_path.display()
-    )?;
-
     // Open bpf_object so we can iterate over maps and progs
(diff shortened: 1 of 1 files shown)


anakryiko:
Why not just do the embedding of literal array of bytes always? Why all the extra options?

libbpf-cargo/src/gen.rs
```diff
@@ -741,6 +733,16 @@ fn gen_skel_contents(_debug: bool, raw_obj_name: &str, obj_file_path: &Path) ->
     gen_skel_attach(&mut skel, object, &obj_name)?;
     writeln!(skel, "}}")?;
 
+    // Coerce to &[u8] just to be safe, as we'll be using debug formatting
+    let bytes: &[u8] = &*mmap;
+    write!(
+        skel,
+        r#"
+        const DATA: &[u8] = &{:?};
```
anakryiko, line 741: it's probably a good idea to split array into multiple lines, if possible. Single very long line will cause troubles for a lot of editors, making it really hard to look at the generated skeleton file. So please consider doing that.

danobi: Rustfmt breaks it up into 100 char wide lines already

anakryiko (approved):

PR #51, opened 2021-01-14
Implement ringbuf support

### LINKED ISSUE #49: Support for ringbuf?
I noticed this crate seems to support the older perf buffer technique for passing per-event data to userspace but not the newer ringbuf map. Do you have plans to include this / would you be interested in a PR?

Thanks! Loving the crate so far.

willfindlay:
This PR adds support for the new ringbuf map from Linux 5.8.

Changes:
* Implement ringbuf support in ringbuf.rs
* Bump libbpf-sys version to add ringbuf bindings
* Update tests/tests.rs to add support for loading different test object files
* Add a test for the ringbuf map along with BPF source file

If this gets merged I can follow up with a patch adding a more realistic example.

Note that running the new test requires at least Linux 5.8. I wonder how easy it would be to gate tests based on kernel version?

Fixes #49.


willfindlay:
Oops, left some testing code in by accident. Should be good now.

danobi:
Not around a computer for a few days — will look after the weekend.

danobi (commented):
Can we split `RingBufferManager` into `RingBuffferManagerBuilder` and `RingBufferManager`? Two reasons:
* would be more consistent with `PerfBuffer` API
* it could encourage downstream applications to setup the ringbuf manager during setup phases of their application. then they'd receive "You must add at least one ring buffer before polling" errors during setup rather than during actual work

Looks reasonable overall, though.

libbpf-rs/src/ringbuf.rs
```diff
@@ -0,0 +1,146 @@
...
+{
+    /// Add a new ringbuf `map` and associated `callback` to this ring buffer
+    /// manager. The callback should take one argument, a slice of raw bytes,
+    /// and return an i32.
+    ///
+    /// Non-zero return values will stop ring buffer consumption early.
+    ///
+    /// The callback provides a raw byte slice. You may find libraries such as
+    /// [`plain`](https://crates.io/crates/plain) helpful.
+    pub fn add_ringbuf(&mut self, map: &Map, callback: F) -> Result<&mut Self> {
```
anakryiko, line 34: can we use just `add()` name to stay in line with libbpf's `ring_buffer__add()` API?

libbpf-rs/src/ringbuf.rs
```diff
@@ -0,0 +1,146 @@
...
+/// preferred over the `perf buffer`.
+pub struct RingBufferManager<F>
+where
+    F: FnMut(&[u8]) -> i32 + 'static,
+{
+    ptr: *mut libbpf_sys::ring_buffer,
+    ringbufs: Vec<Box<F>>,
+}
+
+impl<F> RingBufferManager<F>
```
anakryiko, line 22: we have `PerfBuffer`, so let's stick to `RingBuffer` here? It also matches libbpf naming.

willfindlay: Sure, agreed

libbpf-rs/src/ringbuf.rs
willfindlay: Sure

libbpf-rs/tests/test.rs
```diff
@@ -242,3 +246,89 @@ fn test_object_link_pin() {
...
+
+    mgr.consume()
+        .expect_err("This shouldn't work until we add a ringbuf");
+
+    let map1 = obj
+        .map_immut("ringbuf1")
+        .expect("Error getting ringbuf1 map")
+        .expect("Failed to get ringbuf1 map");
+
+    mgr.add_ringbuf(map1, callback1 as fn(&[u8]) -> i32)
```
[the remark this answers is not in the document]
willfindlay: Because functions in Rust always have unique types (regardless of signature), the cast is necessary to coerce the compiler to use a function pointer instead. The compiler isn't smart enough to figure out that it should be using a function pointer, so we need to do a cast here. See https://stackoverflow.com/questions/27895946/expected-fn-item-found-a-different-fn-item-when-working-with-function-pointer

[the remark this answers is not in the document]
willfindlay: If you can think of a better way to deal with this, I'm happy to discuss. If not, I'll update the documentation to explicitly mention this.

willfindlay:
> Can we split `RingBufferManager` into `RingBuffferManagerBuilder` and `RingBufferManager`? Two reasons:
> 
>     * would be more consistent with `PerfBuffer` API
>
>     * it could encourage downstream applications to setup the ringbuf manager during setup phases of their application. then they'd receive "You must add at least one ring buffer before polling" errors during setup rather than during actual work

The only reason I was hesitant to use a builder here is that I couldn't really get an API I was happy with and it started feeling a bit anti-pattern-ish to me. But I agree with your points, so I'll try refactoring in the next few days.

libbpf-rs/src/ringbuf.rs
```diff
@@ -6,82 +6,128 @@ use std::time::Duration;
...
+    /// Callback is defined as:
+    /// ```rust
+    /// type Callback = fn(&[u8]) -> i32;
+    /// ```
     ///
     /// The callback provides a raw byte slice. You may find libraries such as
     /// [`plain`](https://crates.io/crates/plain) helpful.
-    pub fn add_ringbuf(&mut self, map: &Map, callback: F) -> Result<&mut Self> {
-        let callback = Box::new(callback);
+    pub fn add(&mut self, map: &Map, callback: Callback) -> Result<&mut Self> {
```
danobi, line 41: The builder should probably store a reference to `map` to ensure that `map` doesn't go out of scope before `RingBuffer` is built.

libbpf-rs/src/ringbuf.rs
```diff
@@ -6,82 +6,128 @@ use std::time::Duration;
...
-                    map.fd(),
-                    c_sample_cb,
-                    sample_cb_ptr as *mut _,
-                )
-            };
-
-            // Handle errors
-            if err != 0 {
-                return Err(Error::System(err as i32));
+        for (fd, callback) in self.fd_callbacks.clone() {
```
danobi, line 55: Nit: maybe iterate over refs and clone the `Box` instead of cloning the `Vec`?

libbpf-rs/src/ringbuf.rs
```diff
@@ -6,82 +6,128 @@ use std::time::Duration;
...
+            let sample_cb_ptr = Box::into_raw(callback);
+            if ptr.is_null() {
```
danobi, line 57: Instead of adding a special case in the loop, could do:
```
if ringbugs.is_empty():
  return early

ptr = ring_buffer__new()

for cb in cbs:
  ...
```

libbpf-rs/src/ringbuf.rs
```diff
@@ -6,82 +6,128 @@ use std::time::Duration;
...
+{
     /// Poll from all open ring buffers, calling the registered callback for
     /// each one. Polls continually until we either run out of events to consume
     /// or `timeout` is reached.
     pub fn poll(&self, timeout: Duration) -> Result<()> {
         if self.ptr.is_null() {
-            return Err(Error::InvalidInput(
-                "You must add at least one ring buffer before polling".into(),
-            ));
+            return Err(Error::Internal("Null ringbuf pointer".into()));
```
danobi: We error checked this in the builder so this can be replaced with an `assert!` or removed entirely (I prefer the assert!())

libbpf-rs/src/ringbuf.rs
```diff
@@ -98,9 +144,7 @@ where
     /// to consume or one of the callbacks returns a non-zero integer.
     pub fn consume(&self) -> Result<()> {
         if self.ptr.is_null() {
-            return Err(Error::InvalidInput(
-                "You must add at least one ring buffer before consuming".into(),
-            ));
+            return Err(Error::Internal("Null ringbuf pointer".into()));
```
danobi: Same as above

libbpf-rs/tests/test.rs
```diff
@@ -277,35 +275,42 @@ fn test_object_ringbuf() {
...
-        .expect_err("This shouldn't work until we add a ringbuf");
+    // Test trying to build without adding any ringbufs
+    // Can't use expect_err here since RingBuffer does not implement Debug
+    let builder = libbpf_rs::RingBufferBuilder::new();
+    if let Ok(_) = builder.build() {
+        assert!(
+            false,
+            "Should not be able to build without adding at least one ringbuf"
+        );
+    }
```
danobi, line 292: Nit: could just do:
```
assert!(builder.build().is_err());
```

danobi (commented):
I think using ` type Callback = fn(&[u8]) -> i32; ` instead of `FnMut(&[u8]) -> i32 + 'static` prevents using capturing closures:

```diff --git a/libbpf-rs/tests/test.rs b/libbpf-rs/tests/test.rs
index c8d4c26..bae1332 100644
--- a/libbpf-rs/tests/test.rs
+++ b/libbpf-rs/tests/test.rs
@@ -270,7 +270,9 @@ fn test_object_ringbuf() {
         0
     }

-    fn callback2(data: &[u8]) -> i32 {
+    let mut foo = Vec::new();
+
+    let callback2 = move |data: &[u8]| -> i32 {
         let mut value: i32 = 0;
         plain::copy_from_bytes(&mut value, data).expect("Wrong size");

@@ -278,8 +280,10 @@ fn test_object_ringbuf() {
             V2 = value;
         }

+        foo.push("asdf");
+
         0
-    }
+    };

     // Test trying to build without adding any ringbufs
     // Can't use expect_err here since RingBuffer does not implement Debug
```

```
$ cargo b --tests
   Compiling libbpf-rs v0.6.2 (/home/daniel/dev/libbpf-rs/libbpf-rs)
error[E0308]: mismatched types
   --> libbpf-rs/tests/test.rs:315:23
    |
275 |       let callback2 = move |data: &[u8]| -> i32 {
    |  _____________________-
276 | |         let mut value: i32 = 0;
277 | |         plain::copy_from_bytes(&mut value, data).expect("Wrong size");
278 | |
...   |
285 | |         0
286 | |     };
    | |_____- the found closure
...
315 |       builder.add(map2, callback2).expect("Failed to add ringbuf");
    |                         ^^^^^^^^^ expected fn pointer, found closure
    |
    = note: expected fn pointer `for<'r> fn(&'r [u8]) -> i32`
                  found closure `[closure@libbpf-rs/tests/test.rs:275:21: 286:6]`
```

Capturing closures are really handy so I think we have to make that work.

willfindlay:
> Capturing closures are really handy so I think we have to make that work.

@danobi I don't think there's a way to both support closures and work around the weird function pointer cast I had to do in my original unit tests. We need to pick one or the other.

libbpf-rs/src/ringbuf.rs
willfindlay: I'm running into some borrow checker issues with this in the unit tests since `obj.map()` borrows obj as mutable. That's the reason I opted to use map fd rather than storing references to the maps.

libbpf-rs/src/ringbuf.rs
willfindlay: Sure, done

libbpf-rs/src/ringbuf.rs
willfindlay: Sure

libbpf-rs/src/ringbuf.rs
willfindlay: Sure

libbpf-rs/tests/test.rs
willfindlay: Done

libbpf-rs/src/ringbuf.rs
willfindlay: I think this would end up being uglier since ring_buffer__new() actually includes an implicit ring_buffer__add(). Would end up with the same amount of code in the end but with some extra logic to skip over the first callback in the vec.

danobi:
> Edit: Another issue is that capturing closures don't implement clone, so I don't really see a way to make this work with the new builder. Any ideas are appreciated.

Would changing `build(&self)` -> `build(self)` fix the issue? If we consume the builder then we don't need to clone, right?

libbpf-rs/src/ringbuf.rs
danobi: Ah ok. Missed that bit.

libbpf-rs/src/ringbuf.rs
danobi: Right, but the issue I'm worried about is that the owning `Object` could drop before `RingBufferBuilder::build()` is called, dropping the maps and invalidating the FDs before the kernel can increment the refcount on the map.

Maybe we could drop support for the multi-ring manager feature? Would that allow us to hold a ref to `Map` in `RingBufferBuilder`?

danobi:
> @danobi I don't think there's a way to both support closures and work around the weird function pointer cast (see #51 (comment)) I had to do in my original unit tests. We need to pick one or the other.

Not an expert on rust type system, but would trying to cast a closure to a function pointer even work? It shouldn't, right?

Maybe the alternative is to drop support for multi-ring manager (see above inline comment). And if we do that, we should probably drop the builder pattern too b/c it doesn't get us much anymore.

libbpf-rs/src/ringbuf.rs
anakryiko: Multi-ring manager is important, we shouldn't remove the support for it. If someone drops Object before instantiating RingBuffer, that's too bad for them, they are misusing the API. If we can prevent that -- that's great, if not, then it's fine, doesn't seem like a terrible problem to be concerned about.

danobi:
For storing multiple closures, maybe try this? https://stackoverflow.com/questions/39083375/expected-closure-found-a-different-closure

libbpf-rs/src/ringbuf.rs
willfindlay: Agreed that multi ring buffer is a killer feature of ringbuf over perf buffers. We should definitely support it if possible.

willfindlay:
> > Edit: Another issue is that capturing closures don't implement clone, so I don't really see a way to make this work with the new builder. Any ideas are appreciated.
> 
> Would changing `build(&self)` -> `build(self)` fix the issue? If we consume the builder then we don't need to clone, right?

That would fix it, but  it would also break the ability to have nice one liners like the following:
```rust
    let ringbuf = libbpf_rs::RingBufferBuilder::new()
        .add(map1, callback1)
        .expect("Failed to add ringbuuf")
        .build()
        .expect("Failed to build");
```
The borrow checker now complains that we are moving out of a mutable reference.

If that's not a dealbreaker for you, I'm happy to change build to consume self. This would also be in line with the way PerfBufferBuilder currently works. But I think it's nice to support one liners like this.

willfindlay:
> > @danobi I don't think there's a way to both support closures and work around the weird function pointer cast (see #51 (comment)) I had to do in my original unit tests. We need to pick one or the other.
> 
> Not an expert on rust type system, but would trying to cast a closure to a function pointer even work? It shouldn't, right?
> 
> Maybe the alternative is to drop support for multi-ring manager (see above inline comment). And if we do that, we should probably drop the builder pattern too b/c it doesn't get us much anymore.

Per the discussion above, I think dropping multi-ringbuf support is _not_ the way to go. I'll try to find some time in the next few days to keep messing around with this and try to come up with a solution.

willfindlay:
@danobi It took some serious wrangling, but I've actually ended up with something working that handles both capturing closures and generic functions. The StackOverflow link you provided ended up being quite helpful, so thanks for that. Let me know what you think.

libbpf-rs/tests/test.rs
```diff
@@ -369,3 +372,163 @@ fn test_object_reuse_pined_map() {
...
+    bump_rlimit_mlock();
+
+    let mut obj = get_test_object("ringbuf.bpf.o");
+    let prog = obj
+        .prog("handle__sys_enter_getpid")
+        .expect("error finding program")
+        .expect("failed to find program");
+    let _link = prog.attach().expect("failed to attach prog");
+
+    let mut v1 = vec![];
```
danobi, line 478: Maybe use a `channel` ( https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html ) and send something through from the callback to the test?

libbpf-rs/tests/test.rs
```diff
@@ -369,3 +372,163 @@ fn test_object_reuse_pined_map() {
...
+
+    builder.add(map2, callback2).expect("Failed to add ringbuf");
+
+    let mgr = builder.build().expect("Failed to build");
+
+    // Call getpid to ensure the BPF program runs
+    unsafe { libc::getpid() };
+
+    // This should result in both callbacks being called
+    mgr.consume().expect("Failed to consume ringbuf");
```
danobi, line 533: Then we can verify that the callbacks were indeed called

danobi (commented):
LGTM, thanks for iterating on this!

danobi:
> Note that running the new test requires at least Linux 5.8. I wonder how easy it would be to gate tests based on kernel version?

I ought to set up VM testing at some point. drgn does this for example: https://github.com/osandov/drgn/tree/master/vmtest . It's quite a bit of work though so maybe if libbpf-rs starts getting more heavy usage.

danobi:
> That would fix it, but it would also break the ability to have nice one liners like the following:

Unfortunate, but if that's what's necessary to support closures, I'd rather have closures. Without closures, then (I think) users would have to use some kind of static variable to communicate between callback and the rest of the code. And that's pretty un-ergnomic.

libbpf-rs/tests/test.rs
willfindlay: Sure, I'll look into this

willfindlay:
@danobi I've updated the closure unit test to use channels. I assume you'll squash all these commits on your end?

danobi (approved):
Yep, will squash

PR #55, opened 2021-01-20
libbpf-rs: Add Map::keys() iterator

### LINKED ISSUE #53: Map should implement Iterator
Currently there seems to be no way of iterating over the entries in a `Map`.

danobi:
Add iterator over keys in a map. This provides the user a way to iterate
over the keys in a map.

Note it was a deliberate decision to _not_ provide map values.  The
reason is that we can race between getting the key and the prog deleting
the entry in the map. It's difficult to propagate this error so let the
user handle it instead.

This closes #53.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -1,6 +1,7 @@
 use core::ffi::c_void;
 use std::convert::TryFrom;
 use std::path::Path;
+use std::ptr;
 
 use bitflags::bitflags;
 use nix::errno;
@@ -286,6 +287,15 @@ impl Map {
             Err(Error::System(errno::errno()))
         }
     }
+
+    /// Returns an iterator over keys in this map
+    ///
+    /// Note that if the map is not stable (stable meaning no updates or deletes) during iteration,
+    /// iteration can skip keys, restart from the beginning, or duplicate keys. In other words,
+    /// iteration becomes unpredictable.
+    pub fn keys(&self) -> MapKeyIter {
+        MapKeyIter::new(self, self.key_size())
+    }
 }
 
 #[rustfmt::skip]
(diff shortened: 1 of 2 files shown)


libbpf-rs/tests/test.rs
```diff
@@ -156,6 +157,57 @@ fn test_object_map_lookup_flags() {
         .is_err());
 }
 
+#[test]
+fn test_object_map_key_iter() {
+    bump_rlimit_mlock();
+
+    let mut obj = get_test_object();
```
willfindlay, line 164: Note that the changes introduced in #51 will require `get_test_object()` to be changed to `get_test_object("runqslower.bpf.o")`

libbpf-rs/tests/test.rs
```diff
@@ -156,6 +157,57 @@ fn test_object_map_lookup_flags() {
...
+    assert!(keys.contains(&key1));
+    assert!(keys.contains(&key2));
+    assert!(keys.contains(&key3));
+}
+
+#[test]
+fn test_object_map_key_iter_empty() {
+    bump_rlimit_mlock();
+
+    let mut obj = get_test_object();
```
willfindlay, line 198: Same comment here

libbpf-rs/src/map.rs
```diff
@@ -336,3 +342,37 @@ pub enum MapType {
...
+impl<'a> Iterator for MapKeyIter<'a> {
+    type Item = Vec<u8>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let prev = self.prev.as_ref().map_or(ptr::null(), |p| p.as_ptr());
+
+        if unsafe {
+            libbpf_sys::bpf_map_get_next_key(self.map.fd(), prev as _, self.next.as_mut_ptr() as _)
+        } != 0
+        {
```
anakryiko, line 371: nit: this is quite a convoluted formatting, why not have a local variable storing the returned error?

libbpf-rs/src/map.rs
```diff
@@ -286,6 +287,11 @@ impl Map {
             Err(Error::System(errno::errno()))
         }
     }
+
+    /// Returns an iterator over keys in this map
+    pub fn keys(&self) -> MapKeyIter {
```
anakryiko, line 292: it's probably worthwhile to mention that if BPF map is modified during iteration than iteration can effectively restart from the start, or it could skip some keys, or could duplicate some keys. Basically, reliably key iteration requires stable BPF map without deletes and updates.

anakryiko (approved):

PR #56, opened 2021-01-20
libbpf-rs: Add OpenMap::reuse_map() API

### LINKED ISSUE #52: Interacting with existing maps?
Long story short: I want to use libbf-rs to interact with globally pinned maps that are not created by loading an object with libbpf-rs, but I can not find a way to do that.

Some context: I am working on a user space program which communicates with a bpf object loaded as a tc direct-action via bpf maps. My bpf object in kernel creates the maps, and pins them globally. So, the task for the user-space part is just to manipulate these existing maps. I came up with this design since I have not find any option to load my bpf object as a direct-action w/o using tc. 

Libbpf has a straightforward way to interact with existing maps: bpf_obj_get() returns the fd of the map, then the returned fd can be used for map operations.  Is there any way to do this with libbpf-rs?

Thanks!

danobi:
Useful for interacting with pinned maps.

This closes #52.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -3,7 +3,7 @@ use std::convert::TryFrom;
 use std::path::Path;
 
 use bitflags::bitflags;
-use nix::errno;
+use nix::{errno, unistd};
 use num_enum::TryFromPrimitive;
 use strum_macros::Display;
 
@@ -58,6 +58,29 @@ impl OpenMap {
 
     pub fn set_inner_map_fd(&mut self, inner: &Map) {
         unsafe { libbpf_sys::bpf_map__set_inner_map_fd(self.ptr, inner.fd()) };
+    }
+
+    /// Reuse an already-pinned map for `self`.
+    pub fn reuse_pinned_map<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
+        let cstring = util::path_to_cstring(path)?;
+
+        let fd = unsafe { libbpf_sys::bpf_obj_get(cstring.as_ptr()) };
+        if fd < 0 {
+            return Err(Error::System(errno::errno()));
+        }
+
+        let ret = unsafe { libbpf_sys::bpf_map__reuse_fd(self.ptr, fd) };
+
+        // Always close `fd` regardless of if `bpf_map__reuse_fd` succeeded or failed
+        //
+        // Ignore errors b/c can't really recover from failure
+        let _ = unistd::close(fd);
+
+        if ret != 0 {
+            return Err(Error::System(-ret));
+        }
+
+        Ok(())
     }
 }
 
--- a/libbpf-rs/tests/test.rs
+++ b/libbpf-rs/tests/test.rs
@@ -1,6 +1,8 @@
 use std::fs;
 use std::path::{Path, PathBuf};
 
+use scopeguard::defer;
+
 use libbpf_rs::{MapFlags, Object, ObjectBuilder};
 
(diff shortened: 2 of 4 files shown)


libbpf-rs/src/map.rs
```diff
@@ -59,6 +59,23 @@ impl OpenMap {
     pub fn set_inner_map_fd(&mut self, inner: &Map) {
         unsafe { libbpf_sys::bpf_map__set_inner_map_fd(self.ptr, inner.fd()) };
     }
+
+    /// Reuse a already-pinned map for `self`.
```
anakryiko, line 63: typo: a -> an

libbpf-rs/src/map.rs
```diff
@@ -59,6 +59,23 @@ impl OpenMap {
...
+    pub fn reuse_map<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
```
anakryiko, line 64: I wonder if `reuse_pinned_map` would be a clearer name?

libbpf-rs/tests/test.rs
```diff
@@ -242,3 +242,57 @@ fn test_object_link_pin() {
...
+    let obj_path = get_test_object_path();
+    let mut builder = ObjectBuilder::default();
+    builder.debug(true);
+    let mut open_obj = builder.open_file(obj_path).expect("failed to open object");
+
+    let start = open_obj
+        .map("start")
+        .expect("error finding map")
+        .expect("failed to find map");
+    assert!(start.reuse_map("/asdf").is_err());
```
anakryiko, line 280: will assert!() panic? if yes, you won't unpin the map, wouldn't that be a problem?

libbpf-rs/src/map.rs
```diff
@@ -59,6 +59,23 @@ impl OpenMap {
...
+        let cstring = util::path_to_cstring(path)?;
+
+        let fd = unsafe { libbpf_sys::bpf_obj_get(cstring.as_ptr()) };
+        if fd < 0 {
+            return Err(Error::System(errno::errno()));
+        }
+
+        let ret = unsafe { libbpf_sys::bpf_map__reuse_fd(self.ptr, fd) };
```
anakryiko, line 72: looking at the code of bpf_map__reuse_fd(), I think you are supposed to close(fd) whether this call succeeded or failed.

anakryiko (changes requested):

libbpf-rs/src/map.rs
danobi: Funny story: I woke up in the middle of the night and thought "I forgot to close the fd" and when I woke up this morning I forgot until now

libbpf-rs/tests/test.rs
danobi: Good point; I'll fix it

libbpf-rs/src/map.rs
```diff
@@ -59,6 +59,35 @@ impl OpenMap {
...
+        // Always close `fd` regardless of if `bpf_map__reuse_fd` succeeded or failed
+        match unistd::close(fd) {
+            Ok(_) => (),
+            Err(e) => {
+                return Err(Error::Internal(format!(
+                    "Failed to close pinned map fd: {}",
+                    e
+                )))
+            }
+        }
```
anakryiko, line 83: there isn't much you can do if close() fails, and in practice it shouldn't ever fail, so I'd drop this error handling part

anakryiko (approved):
defer! {} is nice!

PR #59, opened 2021-01-24
runqslower: Don't redefine `struct event` in rust code

danobi:
It's better to define a type in one place so things don't get out of
sync.

### DIFF
--- a/examples/runqslower/src/bpf/runqslower.bpf.c
+++ b/examples/runqslower/src/bpf/runqslower.bpf.c
@@ -10,6 +10,9 @@ const volatile __u64 min_us = 0;
 const volatile pid_t targ_pid = 0;
 const volatile pid_t targ_tgid = 0;
 
+// Dummy instance to get skeleton to generate definition for `struct event`
+struct event _event = {0};
+
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 10240);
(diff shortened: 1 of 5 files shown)


anakryiko:
It is better to have one definition, but defining unnecessary global variable for that seems like a bad solution. Can you please remind me which C types are auto-generated as Rust definitions by skeleton generation code for Rust?

danobi:
> Can you please remind me which C types are auto-generated as Rust definitions by skeleton generation code for Rust?

Any type used in a global variable

examples/runqslower/src/bpf/runqslower.bpf.c
```diff
@@ -10,6 +10,9 @@ const volatile __u64 min_us = 0;
 const volatile pid_t targ_pid = 0;
 const volatile pid_t targ_tgid = 0;
 
+// Dummy instance to get skeleton to generate definition for `struct event`
+struct event _event = {0};
```
anakryiko, line 14: just curious, would it also work if you did any of these:

```
static volatile struct event *_event;
````

```
static volatile struct event _event[0];
```

??


also, nit: `= {0};` (first member is initialized to zero explicitly, all the rest are initialized to zero implicitly) is misleading, use just `= {};`

danobi: Nope, neither work (as it is right now).

> also, nit: = {0}; (first member is initialized to zero explicitly, all the rest are initialized to zero implicitly) is misleading, use just = {};

Oh didn't know that. Will keep in mind

anakryiko (approved):
oh well, C/Rust interop sacrifice, I suppose

PR #60, opened 2021-01-25
libbpf-cargo: gen.rs: Remove _bpf suffix from skeleton name

### LINKED ISSUE #57: Skeleton generation issues with short filename + global variables
This is something I noticed while I was working on my personal project based on libbpf-rs. When I have a very short BPF source file name and global variables, the resulting skeleton generates incorrectly, such that there is a name mismatch between the generated map name and the generated skeleton name.

The above may be a bit confusing, so I'll illustrate with an example:

**src/bpf/ebph.bpf.c:**
```c
/* ... */
volatile const int ebph_logging;
/* ... */
```
**src/bpf/ebph_skel.rs:**
```rust
fn build_skel_config() -> libbpf_rs::Result<libbpf_rs::skeleton::ObjectSkeletonConfig<'static>> {
    let mut builder = libbpf_rs::skeleton::ObjectSkeletonConfigBuilder::new(DATA);
    builder
        .name("ebph_bpf")
        .map("__profile_key_t__alloc", false)
        .map("__profile_key_t__temp", false)
        .map("profiles", false)
        .map("ebph.rodata", true)
        .prog("bprm_check_security");

    builder.build()
}
```

Note that `name` is `ebph_bpf` but we have `ebph.rodata`. In the libbpf debug logs, we can see the problem:
```
libbpf: map 'ebph_bpf.rodata' (global data): at sec_idx 6, offset 0, flags 480.
libbpf: map 3 is "ebph_bpf.rodata"
(...)
libbpf: failed to find skeleton map 'ebph.rodata'
```
If I go into the generated skeleton and manually update to either `.name("ebph")` or `.map("ebph_bpf.rodata")`, we get the expected result. With longer program names, the `_bpf` part gets cut off and we get the expected result as well. You can see this in runqslower for example:
```rust
fn build_skel_config() -> libbpf_rs::Result<libbpf_rs::skeleton::ObjectSkeletonConfig<'static>> {
    let mut builder = libbpf_rs::skeleton::ObjectSkeletonConfigBuilder::new(DATA);
    builder
        .name("runqslower_bpf")
        .map("start", false)
        .map("events", false)
        .map("runqslow.rodata", true)
        .prog("handle__sched_wakeup")
        .prog("handle__sched_wakeup_new")
        .prog("handle__sched_switch");

    builder.build()
}
```

I haven't had a chance to look at the skeleton generation code yet, but I suspect there is a trivial fix.

willfindlay:
The `_bpf` suffix in the generated skeleton name was causing naming conflicts for very short skeleton names. Remove the `_bpf` suffix in the generated skeleton to make names consistent. Fixes #57

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -165,7 +165,7 @@ fn gen_skel_c_skel_constructor(
         {{
             let mut builder = libbpf_rs::skeleton::ObjectSkeletonConfigBuilder::new(DATA);
             builder
-                .name("{name}_bpf")
+                .name("{name}")
         "#,
         name = name
     )?;


danobi:
@anakryiko IIRC we had a discussion about the object name affecting map names or something. Can't quite remember. This fix look right to you?

anakryiko:
This deviates from what C BPF skeleton is doing. There the object name does include _bpf suffix, so I'd prefer to keep it. I think the problem happens earlier, when BpfObject, from which Rust skeleton is generated is created with object name without _bpf. See gen_project() function.

willfindlay:
Don't really have time to pore over the skeleton generation code in detail at the moment. I'll convert this to a draft for now.

danobi:
@willfindlay can you give #61 a try? I think it should fix the issue.

willfindlay:
> @willfindlay can you give #61 a try? I think it should fix the issue.

@danobi It works, thanks! I'll close my PR.

PR #61, opened 2021-01-25
libbpf-cargo: gen: Fix skeletons with short object names

### LINKED ISSUE #57: Skeleton generation issues with short filename + global variables

danobi:
When we opened the object file to generate the skeleton, we used `{name}`.
The generated skeleton then opens the object with `{name}_bpf`. Since
libbpf uses the object name to derive the map/prog names, this skew can
cause issues.

The issue is sometimes covered up by the fact that libbpf truncates the
object name to a certain number of characters. So for longer object
names, this issue doesn't show up. For shorter object names, this issue
occurs.

This closes #57 .

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -165,7 +165,7 @@ fn gen_skel_c_skel_constructor(
         {{
             let mut builder = libbpf_rs::skeleton::ObjectSkeletonConfigBuilder::new(DATA);
             builder
-                .name("{name}_bpf")
+                .name("{name}")
         "#,
         name = name
     )?;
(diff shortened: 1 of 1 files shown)


anakryiko (approved):

willfindlay:
Works great on my end, thanks for the quick fix.

PR #62, opened 2021-02-04
libbpf-rs: Move BTF code into libbpf-rs

danobi:
So that other projects can reuse the BTF parsing code.


anakryiko (commented):
Will this API still make sense if/when we add support for libbpf's BTF writing APIs? I'm worried a bit that we are exposing these types a bit too early. And that they are sufficiently different from what libbpf. Do you think that will cause problems down the line?

anakryiko:
> > Will this API still make sense if/when we add support for libbpf's BTF writing APIs? I'm worried a bit that we are exposing these types a bit too early. And that they are sufficiently different from what libbpf.
> 
> I think so. I think we'll need to diverge from the C types to avoid users writing `unsafe` code anyways.
> 

yeah, I don't have a good answer as to how to keep it close to what libbpf is doing and still feel "Rusty".

> > Do you think that will cause problems down the line?
> 
> My current idea for the write API is something like:

Yeah, it's fine. It's more wasteful (string duplication, etc) and slower than libbpf's approach, but I don't think anyone would care enough about the difference, but this will be more usable. Under cover you can still use libbpf's APIs, if you'll need to "serialize" to raw bytes.

anakryiko (approved):

PR #65, opened 2021-03-03
Update README.md

rtkaratekid:
The versioning was wrong and caused me to curse at my machine too much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To use in your project, add into your `Cargo.toml`:
 
 ```toml
 [dependencies]
-libbpf-rs = "0.1"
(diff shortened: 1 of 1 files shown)


rtkaratekid:
Not sure why the check failed since I only changed the readme.

danobi:
> Not sure why the check failed since I only changed the readme.

Oh looks like the build env changed. I'll fix it in another PR

danobi (approved):

rtkaratekid:
Thanks! Hopefully that helps others avoid the same headaches! I was being told that ```Object``` didn't have an ```open_file``` method and was genuinely confused for a while.

PR #72, opened 2021-04-02
libbpf-rs: Retain type information for CbStruct

### LINKED ISSUE #71: Improper dropping of libbpf_rs::perf_buffer::CbStruct? Segfaulting
I've been messing with the library for a while. I have a few thoughts on things that could potentially be improved, but I'll probably just submit PRs for those (like looping over programs and maps in a BPF object using bpf_map__next and bpf_program__next so you don't need to know the program names, I'm not using any of the CO-RE functionality). But I've hit a segfaulting bug that I'm not completely sure how to fix so I thought I'd submit an issue.

Here's my code that opens the object, loads it, grabs the perf buffer, and then tries to create and populate an Array map with `create_syscall_filter_map`.
```rust
        let file = Path::new("syscall_snoop.bpf.o");

        let perf_name = "syscallsnoop_events";

        let open_obj = ObjectBuilder::default().open_file(file)?;
        let mut object = open_obj.load()?;

        let perf_map = object.map(perf_name)?
                       .ok_or_else(|| anyhow!("Can't fetch perf map '{}'", perf_name))?;

        let perf_buf = PerfBufferBuilder::new(perf_map)
                        .sample_cb(handle_syscall_event)
                        .lost_cb(handle_lost_events)
                        .build()?;

        create_syscall_filter_map(&mut object, syscalls_to_trace)?;
```

And here's the filter map function.
```rust
fn create_syscall_filter_map(object: &mut libbpf_rs::Object, syscalls: Vec<&str>) -> Result<()> {

    let sys_filter_map =  object.map("sys_filter_map")?
                        .ok_or_else(|| anyhow!("[!] Can't get pid map"))?;

    for syscall in syscalls {

        let pre_key = get_sys_id(syscall)?;
        let key = pre_key.to_le_bytes();

        let val = 1u32.to_le_bytes();

        sys_filter_map.update(&key, &val, MapFlags::ANY)?;
    }

    Ok(())
}
```
In this function I access a `HashMap`. The segfault occurs when I try to access an invalid key in the `HashMap`. In that case I've set my errors to back-propagate all the way to main, which then exits with an error.
Well, that's what I expected. What actually happens is a segfault. Here's the `gdb` trace of the segfault
```
Program received signal SIGSEGV, Segmentation fault.
__GI___libc_free (mem=0x1) at malloc.c:3102
3102	malloc.c: No such file or directory.
```
Instead of diving deeper into the `gdb` output, I decided to run valgrind since it was a memory issue. Here's the output
```
==43153== Invalid free() / delete / delete[] / realloc()
==43153==    at 0x483CA3F: free (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==43153==    by 0x19942D: alloc::alloc::dealloc (alloc.rs:104)
==43153==    by 0x19981C: <alloc::alloc::Global as core::alloc::Allocator>::deallocate (alloc.rs:239)
==43153==    by 0x19952B: alloc::alloc::box_free (alloc.rs:334)
==43153==    by 0x19ABA0: core::ptr::drop_in_place (mod.rs:179)
==43153==    by 0x19B917: <libbpf_rs::perf_buffer::CbStruct as core::ops::drop::Drop>::drop (/home/vagrant/.cargo/registry/src/github.com-1ecc6299db9ec823/libbpf-rs-0.7.0/src/perf_buffer.rs:22)
==43153==    by 0x1703AE: core::ptr::drop_in_place (mod.rs:179)
==43153==    by 0x1702FB: core::ptr::drop_in_place (mod.rs:179)
==43153==    by 0x1709A0: core::ptr::drop_in_place (mod.rs:179)
==43153==    by 0x1814F9: syscalls::sys_snoop::SyscallSnooper::new (src/sys_snoop.rs:66)
==43153==    by 0x139AFE: syscalls::main (src/main.rs:43)
==43153==    by 0x13848A: core::ops::function::FnOnce::call_once (function.rs:227)
==43153==  Address 0x1 is not stack'd, malloc'd or (recently) free'd
==43153==
```
This points to `libbpf_rs::perf_buffer::CbStruct::drop` not functioning correctly in this case. 
If I change the order of my code, then I can avoid the segfault.
```rust
        let file = Path::new("syscall_snoop.bpf.o");

        let perf_name = "syscallsnoop_events";

        let open_obj = ObjectBuilder::default().open_file(file)?;
        let mut object = open_obj.load()?;

        // now call this prior to getting the perf buffer
        create_syscall_filter_map(&mut object, syscalls_to_trace)?;

        let perf_map = object.map(perf_name)?
                       .ok_or_else(|| anyhow!("Can't fetch perf map '{}'", perf_name))?;

        let perf_buf = PerfBufferBuilder::new(perf_map)
                        .sample_cb(handle_syscall_event)
                        .lost_cb(handle_lost_events)
                        .build()?;
```
I took a look at how `Drop` is implemented for `CbStruct` and it's just casting the raw pointers to `Box` with the expectation that it will then be cleaned up properly I believe. I'm not completely sure why this segfault is occurring, but perhaps there is a safer way to deal with dropping/freeing the `CbStruct` such that it can contain the unsafe code better? I wish I had a solution to this to share, but in my own libbpf Rust library I've worked on I actually stored the functions in a `Box` initially and then coerced into `c_void` pointers when needed and that's how I avoided this issue.

danobi:
Before, we didn't retain callback type information in CbStruct. That
meant when we called Box::from_raw() anything could have happened
(segfault, memory leak, etc.).

This commit reworks the callback storing bits to store fat pointers
instead of templated types for all the callbacks. The fat pointers store
the type information such that rust can recover the type information at
runtime. This also simplifies the code a bit.

This fixes #71.

### DIFF
--- a/libbpf-rs/src/perf_buffer.rs
+++ b/libbpf-rs/src/perf_buffer.rs
@@ -1,6 +1,5 @@
 use core::ffi::c_void;
 use std::boxed::Box;
-use std::ptr;
 use std::slice;
 use std::time::Duration;
 
@@ -10,37 +9,29 @@ fn is_power_two(i: usize) -> bool {
     i > 0 && (!(i & (i - 1))) > 0
 }
 
+// Workaround for `trait_alias`
+// (https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html)
+// not being available yet. This is just a custom trait plus a blanket implementation.
+pub trait SampleCb: FnMut(i32, &[u8]) + 'static {}
+impl<T> SampleCb for T where T: FnMut(i32, &[u8]) + 'static {}
+
+pub trait LostCb: FnMut(i32, u64) + 'static {}
+impl<T> LostCb for T where T: FnMut(i32, u64) + 'static {}
+
 struct CbStruct {
-    // Both sample_cb and lost_cb are owning pointers to Box's
-    sample_cb: *mut c_void,
-    lost_cb: *mut c_void,
-}
-
-impl Drop for CbStruct {
-    fn drop(&mut self) {
-        if !self.sample_cb.is_null() {
-            let _ = unsafe { Box::from_raw(self.sample_cb) };
-        }
-
-        if !self.lost_cb.is_null() {
-            let _ = unsafe { Box::from_raw(self.lost_cb) };
-        }
-    }
+    sample_cb: Option<Box<dyn SampleCb>>,
+    lost_cb: Option<Box<dyn LostCb>>,
 }
 
 /// Builds [`PerfBuffer`] instances.
-pub struct PerfBufferBuilder<'a, F, G>
-where
-    F: FnMut(i32, &[u8]) + 'static,
-    G: FnMut(i32, u64) + 'static,
-{
+pub struct PerfBufferBuilder<'a> {
     map: &'a Map,
     pages: usize,
-    sample_cb: Option<Box<F>>,
-    lost_cb: Option<Box<G>>,
+    sample_cb: Option<Box<dyn SampleCb>>,
+    lost_cb: Option<Box<dyn LostCb>>,
 }
 
-impl<'a> PerfBufferBuilder<'a, fn(i32, &[u8]), fn(i32, u64)> {
+impl<'a> PerfBufferBuilder<'a> {
     pub fn new(map: &'a Map) -> Self {
         Self {
             map,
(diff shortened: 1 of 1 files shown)


anakryiko:
Does ring_buffer API need the same fix?

danobi:
> Does ring_buffer API need the same fix?

No, ring_buffer API does it better, actually. I've reworked this PR to use ring_buffer API's approach. The code is simpler now IMO.

danobi:
https://github.com/danobi/libbpf-rs/tree/cbstruct_drop_type-v2 is my alternate approach but I think I like this one better.

anakryiko (approved):
LGTM (as far as I can understand what's going on ;) )

PR #73, opened 2021-04-06
libbpf-rs: Fix is_power_two implementation

chenhengqi:
The function `fn is_power_two` in `perf_buffer.rs` don't work as expected (It works in C++ though).
This PR fixes that and add a small test.

Signed-off-by: Hengqi Chen <chenhengqi@outlook.com>


danobi (approved):
Oops, forgot `!` in rust is bitwise not.

danobi:
Thanks for the fix!

PR #75, opened 2021-04-16
libbpf-rs: Add bpf iterator support

brianc118:
Adds support for bpf iterators introduced in Linux 5.8.

If this is good, I'll make another PR with a very basic example.

Changes:

- Implement support for bpf iterator.
- Add test using task iterator.
- Set RLIMIT_MEMLOCK for all tests. Tests were flaky and would fail depending on order of execution.

Note. This is opinionated and `Iter` must be associated with a `Link`. Perhaps we could add support for directly reading from a path or fd instead, though not sure what the use case would be.


brianc118:
Didn't realize tests were being ignored in CI which makes sense. Failures before were because the new test is being run. VM testing would be nice but is a bit of work, so just ignoring this test as well for now.

libbpf-rs/src/iter.rs
```diff
@@ -0,0 +1,62 @@
...
+    ///
+    /// The data is read in chunks to an intermediate buffer of `BUF_SIZE`
+    /// elements of type `T`.
+    ///
+    /// # Safety
+    ///
+    /// The type layout of `T` must match the layout of the data structure being
+    /// dumped by the bpf program to guarantee that the returned objects are
+    /// valid.
+    pub unsafe fn read<T: Copy, const BUF_SIZE: usize>(&self) -> Result<Vec<T>> {
```
danobi, line 28: Couple things:

* We want almost all APIs in libbpf-rs to be safe -- we'll generally make an API unsafe if it takes an unsafe parameter (eg pointer, FD)
* Taking in a type `T` and a constant `BUF_SIZE` seems unnecessarily complex. How about we do `pub fn read(&self, &mut [u8) -> Result<usize>`. No need to reinterpret bytes here -- we can let the user handle it just like w/ the map functions: https://docs.rs/libbpf-rs/0.7.2/libbpf_rs/struct.Map.html#method.lookup

libbpf-rs/src/link.rs
```diff
@@ -61,6 +61,15 @@ impl Link {
             Ok(())
         }
     }
+
+    /// Get file descriptor of link.
+    ///
+    /// # Safety
+    ///
+    /// It is not safe to use the file descriptor after this link is dropped.
+    pub unsafe fn get_fd(&self) -> i32 {
```
danobi, line 70: Should be ok to leave this safe just like map API: https://docs.rs/libbpf-rs/0.7.2/libbpf_rs/struct.Map.html#method.fd. Safe is more in line with rest of rust. For example, you can take pointer of anything safely but when you try to deref it it'll involve unsafety. See my note above about how we'll generally make an API unsafe if we take in a FD.

libbpf-rs/tests/bin/src/taskiter.bpf.c
```diff
@@ -0,0 +1,31 @@
...
+
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_core_read.h>
+
+struct index_pid_pair {
+  uint32_t i;
+  pid_t pid;
+}
+
+uint32_t i = 0;
```
danobi, line 11: use `static uint32_t i = 0;` to not export this type to userspace (`libbpf-cargo gen` will generate accessors for this otherwise)

danobi (changes requested):

libbpf-rs/src/iter.rs
brianc118: Makes sense -- leave it to the user to use `plain` or similar. Maybe we can get it to implement `std::io::Read`: https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read. That way the user can just use `read_to_end(&self, &mut Vec[i8])` to read into a vector rather than having to think about how what to set as size of the buffer. What do you think?

danobi: Yeah that's a great idea. `impl Read for Link` sounds pretty ergonomic.

brianc118: Ok done. I've introduced an `OpenIter` so that the fd doesn't need to be closed in the read implementation (i.e. we don't have to read the entire file in one `read()` call, as expected by `std::io::Read`.

libbpf-rs/tests/bin/src/taskiter.bpf.c
```diff
@@ -0,0 +1,31 @@
...
+SEC("iter/task")
+int dump_pid(struct bpf_iter__task *ctx)
+{
+  struct seq_file *seq = ctx->meta->seq;
+  struct task_struct *task = ctx->task;
+
+  if (!task)
+    return 0;
+
+  struct index_pid_pair p;
```
anakryiko, line 22: please stick to C89 conventions: all variables are declared at the top of a function or lexical block

libbpf-rs/src/iter.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// Open the iterator for reading.
+    pub fn open(&self) -> Result<OpenIter> {
+        let link_fd = self.link.get_fd();
+        let fd = unsafe { libbpf_sys::bpf_iter_create(link_fd) };
+        if fd < 0 {
+            return Err(Error::System(-fd));
+        }
+        Ok(OpenIter { fd })
+    }
```
danobi, line 25: Any reason we can't do the contents of `open()` inside `Iter::new()`? When `Iter` is dropped we can close the fd there.

danobi: I'm trying to imagine why someone would create an `Iter` but not call `iter.open()` on it but I can't think of any reason.

brianc118: Right. Yeah I guess there isn't really a use case for calling `open()` multiple times either. Let's remove it.

libbpf-rs/src/iter.rs
```diff
@@ -0,0 +1,41 @@
...
+pub struct Iter {
+    fd: i32,
+}
+
+impl Iter {
+    pub fn new(link: &Link) -> Result<Self> {
+        let link_fd = link.get_fd();
+        let fd = unsafe { libbpf_sys::bpf_iter_create(link_fd) };
+        if fd < 0 {
+            return Err(Error::System(-fd));
```
anakryiko, line 21: you use last_os_error() in read, shouldn't that be used here as well?

brianc118: We could do that but we'd need to return `std::result::Result<usize, std::io::Error>` rather than `std::result::Result<usize, crate::Error>` and I don't think that's any better? Ideally, everything in this library should be using `crate::Error`. We don't in read since we have to `impl io::Read`.

anakryiko: as it is right now, -fd will be -1 which is -EPERM, which is a lie.

brianc118: oh right, my bad. Should be fixed now.

anakryiko:
thanks, looks good to me now!

danobi:
Awesome, LGTM too. Thanks for working on this!

2 final things:
* Can you squash the commits into something appropriate for the commit log? Ideally with a short description of the changes too
* Re:  "make another PR with a very basic example.": let's do that in https://github.com/libbpf/libbpf-bootstrap . The example would go in `examples/rust`

brianc118:
Thanks. Squashed the commits now. Sure I can add something to libbpf-boostrap.

danobi (approved):
Thanks.

In the future you probably wanna split the test changes/fixes into a separate commit. Ideally each commit should be doing 1 logical thing.

PR #79, opened 2021-05-05
Populate and return list of progs and maps in Object

rtkaratekid:
This PR is an attempt to allow users to iterate over maps and programs in a BPF Object without knowing names ahead of time.
Adjustments and changes are welcome. This code currently does compile and runs well.

### DIFF
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -300,46 +300,7 @@ fn object_map_next(
 }
 
 
-fn get_maps(ptr: *mut libbpf_sys::bpf_object) -> Result<HashMap<String, Map>> {
-    let mut maps: HashMap<String, Map> = HashMap::new();
-    let mut map: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
-
-    loop {
-        // while there's another map to grab
-        if let Some(next_map_ptr) = object_map_next(ptr, map).unwrap() {
-
-            // get map name
-            let mname = unsafe { libbpf_sys::bpf_map__name(next_map_ptr) };
-            let err = unsafe { libbpf_sys::libbpf_get_error(mname as *const _) };
-            if err != 0 {
-                std::process::exit(err as i32);
-            }
-
-            let fd = unsafe { libbpf_sys::bpf_map__fd(next_map_ptr) };
-            if fd < 0 {
-                std::process::exit(errno::errno());
-            } else {
-                let def = unsafe { ptr::read(libbpf_sys::bpf_map__def(next_map_ptr)) };
-
-                // convert to strings
-                let name = util::c_ptr_to_string(mname).unwrap();
-
-                maps.insert(
-                    name.clone(),
-                    Map::new(fd, name, def.type_, def.key_size, def.value_size, next_map_ptr),
-                );
-                map = next_map_ptr;
-            }
-
-        } else {
-            break;
-        }
-    }
-
-    Ok(maps)
-
-}
-     
+    
 
 impl Object {
     fn new(ptr: *mut libbpf_sys::bpf_object) -> Self {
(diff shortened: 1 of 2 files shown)


rtkaratekid:
I'm not sure why `rust_fmt` failed.

libbpf-rs/src/object.rs
```diff
@@ -283,6 +283,25 @@ pub struct Object {
...
+    Ok(util::ptr_to_option(next_prog_ptr))
+}
+
+fn object_map_next(
+    obj: *const libbpf_sys::bpf_object,
+    map: *mut libbpf_sys::bpf_map, // needed to know what the next map is
+) -> Result<Option<*mut libbpf_sys::bpf_map>> {
+    let next_map_ptr: *mut libbpf_sys::bpf_map = unsafe { libbpf_sys::bpf_map__next(map, obj) };
+    Ok(util::ptr_to_option(next_map_ptr))
+}
```
danobi, line 300: I don't think we need helpers for this. First, this always returns `Ok(..)` so the `Result<..>` doesn't do anything. And second, we can just do:
```
loop {
  ...
  let next_map_ptr: ... = unsafe { libbpf_sys::bpf_map__next(...) };
  if next_map_ptr.is_null() {
    break;
  }
  ...
}
```
in the code below

libbpf-rs/src/object.rs
```diff
@@ -334,6 +353,58 @@ impl Object {
         self.map(name).unwrap().unwrap()
     }
 
+
+    pub fn get_maps(&mut self) -> Result<&HashMap<String, Map>> {
```
danobi, line 357: I think we should hide `HashMap` implementation detail from the user in case we ever want to change the implementation without breaking API. What do you think about returning a struct that implements `std::iter::Iterator`? Like what https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.iter does.

```suggestion
    pub fn get_maps(&mut self) -> Result<Iter<(String, Map)>> {
```

libbpf-rs/src/object.rs
```diff
@@ -334,6 +353,58 @@ impl Object {
...
+
+            loop {
+                // while there's another map to grab
+                if let Some(next_map_ptr) = object_map_next(self.ptr, map).unwrap() {
+
+                    // get map name
+                    let mname = unsafe { libbpf_sys::bpf_map__name(next_map_ptr) };
+                    let err = unsafe { libbpf_sys::libbpf_get_error(mname as *const _) };
+                    if err != 0 {
+                        std::process::exit(err as i32);
```
danobi, line 376: As a library, we should never cause applications to exit. This should be

```suggestion
                        return Err(Error::System(err as i32));

File: libbpf-rs/src/object.rs [hunk 4]
```diff
@@ -334,6 +353,58 @@ impl Object {
...
+            Ok(&self.maps)
+
+        } else {
+     
+            let mut maps: HashMap<String, Map> = HashMap::new();
+            let mut map: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
+
+            loop {
+                // while there's another map to grab
+                if let Some(next_map_ptr) = object_map_next(self.ptr, map).unwrap() {
```
danobi, line 370: See above, we don't need to return a `Ok(..)` and `unwrap()`. No reason to do extra work

libbpf-rs/src/object.rs
```diff
@@ -334,6 +353,58 @@ impl Object {
...
+                    // get map name
+                    let mname = unsafe { libbpf_sys::bpf_map__name(next_map_ptr) };
+                    let err = unsafe { libbpf_sys::libbpf_get_error(mname as *const _) };
+                    if err != 0 {
+                        std::process::exit(err as i32);
+                    }
+
+                    let fd = unsafe { libbpf_sys::bpf_map__fd(next_map_ptr) };
+                    if fd < 0 {
+                        std::process::exit(errno::errno());
```
danobi, line 381: See above

libbpf-rs/src/object.rs
danobi, line 357: Also this should probably be named `maps(..)` and not `get_maps(..)`. See https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

danobi (commented):
Couple other things:
* Similar comments on the `get_progs(..)` code as the `get_maps(..)` code
* I think we should also implement similar functions for `OpenObject` too and not just `Object`
* Please run `cargo fmt` to rustfmt your code. The CI is complaining that the style does not match

libbpf-rs/src/object.rs
rtkaratekid: Hm, for my personal project (not published yet) I get a lot of use from iterating and getting both the map name and the `Map` itself. I suppose I could just do `Map.name()` instead of unpacking the `HashMap` tuple.

Are you suggesting we just return `self.maps.iter()`? It might have to be `iter_mut()` if you want to do something with the maps, but same point I believe.

libbpf-rs/src/object.rs
rtkaratekid: This is my fault for not reviewing my code as much as I should've before making a PR.

libbpf-rs/src/object.rs
danobi: Yeah, the iterator element type could be `(&str, &mut Map)`. And probably  not exactly `self.maps.iter()` b/c that would return a HashMap type. Reason would be in the event of us wanting to move from HashMap to something else, we wouldn't want to create a temporary HashMap just so we don't break API.

I'm suggesting we create our own type, `[Open]MapIter` and `[Open]ProgIter` (can use a macro to generate these) and return those instead.

rtkaratekid: I'm afraid my greeness in Rust will show a bit here because I don't know how to use a macro to generate these types. I'm not sure even what to research about it. I'll have a look around when I get some time though. It would be nifty to use a macro instead of rolling it by hand.

rtkaratekid: I will probably have time to get to this tomorrow.

danobi:
Here's what I had in mind: http://ix.io/3mwu

It's not the most beautiful macro ever but it generates the right code so whatever :p . Should probably add more tests too

libbpf-rs/src/object.rs
```diff
@@ -225,6 +225,40 @@ impl OpenObject {
         self.map(name).unwrap().unwrap()
     }
 
+    /// Function that returns an Iterator over all available maps
+    /// in a given BPF [`Object`]
+    pub fn maps(&mut self) -> Result<&HashMap<String, OpenMap>> {
+        if !self.maps.is_empty() {
+            Ok(&self.maps)
```
danobi, line 232: Not sure this logic is right. If user only accessed 1 map by name, then the iterator will only ever return 1 map even if there's more.

libbpf-rs/src/object.rs
rtkaratekid: Ugh sorry about the long turnaround time. I'm in the middle of real life stuff (moving). I **will** get to this though.

willfindlay:
@rtkaratekid Perhaps you could check out #90 and see if that accomplishes the same behaviour you're looking for?

rtkaratekid:
> @rtkaratekid Perhaps you could check out #90 and see if that accomplishes the same behaviour you're looking for?

Commented over there.

PR #82, opened 2021-05-09
Allow usage of anonymous structs and unions

mimullin-bbry:
Anonymous structs and unions cannot be used in a libbpf-rs project as the structs and unions are not named.

Name anonymous structs as "anon_struct_X" where X is an increasing integer, and anonymous unions as "anon_union_X"

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -584,9 +584,8 @@ impl<'a> Btf<'a> {
     fn load_struct(&mut self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
         let name = match self.get_btf_str(t.name_off as usize)? {
             "" => {
-                let n = format!("{}{}", ANON_AGGREGATE_PREFIX, self.anon_agg_count);
                 self.anon_agg_count += 1;
-                n
+                format!("{}{}", ANON_AGGREGATE_PREFIX, self.anon_agg_count)
             }
             n => n.to_string(),
         };
@@ -601,9 +600,8 @@ impl<'a> Btf<'a> {
     fn load_union(&mut self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
         let name = match self.get_btf_str(t.name_off as usize)? {
             "" => {
-                let n = format!("{}{}", ANON_AGGREGATE_PREFIX, self.anon_agg_count);
                 self.anon_agg_count += 1;
-                n
+                format!("{}{}", ANON_AGGREGATE_PREFIX, self.anon_agg_count)
             }
             n => n.to_string(),
         };
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -5,6 +5,7 @@ use std::ffi::{c_void, CStr, CString};
 use std::fmt::Write;
 use std::mem::size_of;
 use std::slice;
+use std::sync::Mutex;
 
 use anyhow::{bail, ensure, Result};
 use scroll::Pread;
@@ -17,6 +18,8 @@ pub struct Btf<'a> {
     ptr_size: u32,
     string_table: &'a [u8],
     bpf_obj: *mut libbpf_sys::bpf_object,
+    anon_union_count: Mutex<u32>,
+    anon_struct_count: Mutex<u32>,
 }
 
 impl<'a> Btf<'a> {
@@ -92,6 +95,8 @@ impl<'a> Btf<'a> {
             ptr_size: ptr_size as u32,
             string_table: str_data,
             bpf_obj,
+            anon_union_count: Mutex::new(0),
+            anon_struct_count: Mutex::new(0),
         };
 
         // Load all types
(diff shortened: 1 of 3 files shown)


libbpf-cargo/src/btf/btf.rs
```diff
@@ -17,6 +18,8 @@ pub struct Btf<'a> {
     ptr_size: u32,
     string_table: &'a [u8],
     bpf_obj: *mut libbpf_sys::bpf_object,
+    anon_union_count: Mutex<u32>,
+    anon_struct_count: Mutex<u32>,
```
danobi, line 22: I'm assuming you're using a `Mutex` to avoid the borrowing issues and to not change method signatures from `&self` -> `&mut self`. Mutex is kinda weird semantically. `AtomicU32` is better but it's still weird. I think we should just change the signatures to `&mut self`, use a regular `u32`, and skip the helper (it doesn't do much anyways). See below for my suggestion

libbpf-cargo/src/btf/btf.rs
```diff
@@ -578,17 +600,27 @@ impl<'a> Btf<'a> {
...
             members: self.load_members(t, extra)?,
         }))
     }
 
     fn load_union(&self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+        let name = self.get_name_or_anonymous_string(
+            t.name_off as usize,
+            &self.anon_union_count,
+            "anon_union_",
+        )?;
```
danobi, line 621: Similar to above

libbpf-cargo/src/btf/btf.rs
```diff
@@ -578,17 +600,27 @@ impl<'a> Btf<'a> {
     }
 
     fn load_struct(&self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+        let name = self.get_name_or_anonymous_string(
+            t.name_off as usize,
+            &self.anon_struct_count,
+            "anon_struct_",
+        )?;
```
danobi, line 607: How about just:
```suggestion
        let name = match self.get_btf_str(t.name_off as usize)? {
            "" => {
                let n = format!("anon_struct_{}", self.anon_struct_count);
                self.anon_struct_count += 1;
                n
            }
            n => n.to_string(),
        };
```

danobi (commented):
Thanks for the PR! Couple minor things

libbpf-cargo/src/btf/btf.rs
mimullin-bbry: Yes, I was hesitant to change the method signatures.  Fixed.

libbpf-cargo/src/btf/btf.rs
mimullin-bbry: updated.

libbpf-cargo/src/btf/btf.rs
```diff
@@ -577,18 +581,34 @@ impl<'a> Btf<'a> {
         }))
     }
 
-    fn load_struct(&self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+    fn load_struct(&mut self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+        let name = match self.get_btf_str(t.name_off as usize)? {
+            "" => {
+                let n = format!("anon_struct_{}", self.anon_struct_count);
```
danobi, line 587: Nit: let's do `__anon_struct_{}` instead

We already have `__pad_{}` so it'd match

libbpf-cargo/src/btf/btf.rs
```diff
@@ -577,18 +581,34 @@ impl<'a> Btf<'a> {
...
             size: t.type_id,
             members: self.load_members(t, extra)?,
         }))
     }
 
-    fn load_union(&self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+    fn load_union(&mut self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+        let name = match self.get_btf_str(t.name_off as usize)? {
+            "" => {
+                let n = format!("anon_union_{}", self.anon_union_count);
```
danobi, line 604: Same as above

danobi:
ping @anakryiko

libbpf-cargo/src/btf/btf.rs
```diff
@@ -577,18 +581,34 @@ impl<'a> Btf<'a> {
         }))
     }
 
-    fn load_struct(&self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+    fn load_struct(&mut self, t: &btf_type, extra: &'a [u8]) -> Result<BtfType<'a>> {
+        let name = match self.get_btf_str(t.name_off as usize)? {
+            "" => {
+                let n = format!("{}{}", ANON_AGGREGATE_PREFIX, self.anon_agg_count);
+                self.anon_agg_count += 1;
+                n
```
anakryiko, line 589: why start with zero index? I think __anon_1 is a better starting point and allows to make this a bit nicer:

```
self.anon_agg_count += 1;
format!("{}{}", ANON_AGGREGATE_PREFIX, self.anon_agg_count)
```

anakryiko (approved):
looks ok, but I'd do __anon_1 as the first type, it's not array index, so totally makes sense to have it 1-based.

libbpf-cargo/src/btf/btf.rs
mimullin-bbry: +1.  I wanted to do this originally :)  Will address.

PR #87, opened 2021-05-12
Generate impl Default functionality for structures which cannot use #[derive(Default)]

mimullin-bbry:
If an array has length > 32, #[derive(Default)] will fail to compile.

This PR modifies libbpf-cargo so that it no longer generates skeleton code which will fail to compile
due to the array size limitation of #[derive(Default)].

Determine if a structure contains an array longer than 32.  If it does, output both the structure definition
and also an impl Default for the structure.  If a structure does not contain an array longer than 32 use #[derive(Default)]
for output

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -247,7 +247,7 @@ impl<'a> Btf<'a> {
 
         Ok(match ty {
             BtfType::Void => "std::ffi::c_void::default()".to_string(),
-            BtfType::Int(_) => format!("{}::default()", self.type_declaration(type_id)?),
+            BtfType::Int(_) => format!("{}::default()", self.type_declaration(stripped_type_id)?),
             BtfType::Ptr(_) => "std::ptr::null_mut()".to_string(),
             BtfType::Array(t) => {
                 let val_ty = self.type_declaration(t.val_type_id)?;
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -233,6 +233,50 @@ impl<'a> Btf<'a> {
         })
     }
 
+    pub fn type_default(&self, type_id: u32) -> Result<String> {
+        let stripped_type_id = self.skip_mods_and_typedefs(type_id)?;
+        let ty = self.type_by_id(stripped_type_id)?;
+
+        Ok(match ty {
+            BtfType::Void => "std::ffi::c_void::default()".to_string(),
+            BtfType::Int(t) => {
+                let width = match (t.bits + 7) / 8 {
+                    1 => "8",
+                    2 => "16",
+                    4 => "32",
+                    8 => "64",
+                    16 => "128",
+                    _ => bail!("Invalid integer width"),
+                };
+
+                if t.encoding == btf::BtfIntEncoding::Signed {
+                    format!("i{}::default()", width)
+                } else {
+                    format!("u{}::default()", width)
+                }
+            }
+            BtfType::Ptr(_) => "std::ptr::null_mut()".to_string(),
+            BtfType::Array(t) => {
+                let val_ty = self.type_declaration(t.val_type_id)?;
+
+                format!("[{}::default(); {}]", val_ty, t.nelems)
+            }
+            BtfType::Struct(t) | BtfType::Union(t) => format!("{}::default()", t.name),
+            BtfType::Enum(t) => format!("{}::default()", t.name),
+            BtfType::Var(t) => format!("{}::default()", self.type_declaration(t.type_id)?),
+            BtfType::Func(_)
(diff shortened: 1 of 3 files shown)


libbpf-cargo/src/btf/btf.rs
```diff
@@ -227,6 +227,50 @@ impl<'a> Btf<'a> {
...
+                    16 => "128",
+                    _ => bail!("Invalid integer width"),
+                };
+
+                if t.encoding == btf::BtfIntEncoding::Signed {
+                    format!("i{}::default()", width)
+                } else {
+                    format!("u{}::default()", width)
+                }
+            }
```
danobi, line 251: Can't we do `format!("{}::default()", self.type_declaration()?)` here? Instead of duplicating parts of self.type_declaration()

libbpf-cargo/src/btf/btf.rs
```diff
@@ -227,6 +227,50 @@ impl<'a> Btf<'a> {
         })
     }
 
+    pub fn type_default(&self, type_id: u32) -> Result<String> {
```
danobi, line 230: Can you leave a doc comment here too? Maybe something along the lines of "Returns an expression that evaluates to the default value of type `typeid` in string form"

libbpf-cargo/src/btf/btf.rs
```diff
@@ -381,26 +412,85 @@ impl<'a> Btf<'a> {
...
+                                _ => {},
+                            }
+
+                            let field_ty_def = match self.type_default(field_ty_id) {
+                                Ok(def) => def,
+                                Err(_) => {
+                                    can_default_impl = false;
+                                    String::from("")
+                                }
+                            };
```
danobi: Doesn't an error from `self.type_default()` imply malformed BTF? I think this can just be:
```
let field_ty_def = self.type_default(field_ty_id)?;
```
and we can have one less variable (`can_default_impl`)

danobi (commented):
Thanks for writing tests :)

libbpf-cargo/src/btf/btf.rs
mimullin-bbry: Not with the current field structure of a BtfInt.

IMHO a refactor could be done here to break the type_declaration and type_definition to be functions impl by the various types themselves.  I'll upload a specific 'idea' for BtfInt and we can discuss once I get it done.

danobi: Hmm, why is it not possible? Unclear what you mean by "current field structure"

mimullin-bbry: For some reason my brain was fixated on only using the information from t to call type_definition.  I guess there is no reason not to re-use the type_id already available.

NM.  I'll fix.

libbpf-cargo/src/btf/btf.rs
```diff
@@ -227,6 +227,43 @@ impl<'a> Btf<'a> {
...
+    /// Rule of thumb is `ty` must be a type a variable can have.
+    ///
+    /// Type qualifiers are discarded (eg `const`, `volatile`, etc).
+    pub fn type_default(&self, type_id: u32) -> Result<String> {
+        let stripped_type_id = self.skip_mods_and_typedefs(type_id)?;
+        let ty = self.type_by_id(stripped_type_id)?;
+
+        Ok(match ty {
+            BtfType::Void => "std::ffi::c_void::default()".to_string(),
+            BtfType::Int(_) => format!("{}::default()", self.type_declaration(type_id)?),
```
danobi, line 244: Nit: can save a few cycles 
```suggestion
            BtfType::Int(_) => format!("{}::default()", self.type_declaration(stripped_type_id)?),
```

libbpf-cargo/src/btf/btf.rs
```diff
@@ -380,27 +408,88 @@ impl<'a> Btf<'a> {
...
+                                Ok(def) => {
+                                    impl_default.push(format!(
+                                        r#"            {field_name}: {field_ty_str}"#,
+                                        field_name = member.name,
+                                        field_ty_str = def
+                                    ));
+                                }
+                                Err(_) => {
+                                    if gen_impl_default {
+                                        bail!("Could not construct a necessary Default Impl");
```
danobi, line 443: More helpful error message
```suggestion
                                Err(e) => {
                                    if gen_impl_default {
                                        bail!("Could not construct a necessary Default Impl: {}", e);
```

danobi (approved):
LGTM

danobi:
Can you rebase?

danobi:
It still says it has merge conflicts -- I think you need to `git rebase upstream/master` your branch, where `upstream` is the remote for this repo.

mimullin-bbry:
My branch has become unusable.  I cannot rebase properly with it.  I've opened #101 to re-try a pull request.

PR #89, opened 2021-05-12
libbpf-cargo: Add SkeletonBuilder library for use in build scripts

### LINKED ISSUE #63: Feature Request: Expose libbpf-cargo BPF object and skeleton generation code as a library
Crates like [bindgen](https://rust-lang.github.io/rust-bindgen/) often expose code generation functionality via a builder pattern which can then be easily integrated into a project's build.rs file. I have found myself wanting to do something similar with libbpf-rs in one of my large projects, so it would be nice if this crate offered something similar.

Currently, as a workaround, I am just invoking the libbpf-cargo CLI from my build.rs, but obviously this is not ideal.

danobi:
Build scripts are pretty standard for projects that have a codegen step
as part of the build. The current method is kind of strange (took a
while to realize) where we have the user run a custom cargo tool. It'd
be nicer to give users the option to encapsulate everything in a build
script such that `cargo build` "just works".

See 4th commit in series for how runqslower example looks with the
build script.

This closes #63.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -78,61 +78,65 @@ fn check_clang(debug: bool, clang: &Path, skip_version_checks: bool) -> Result<(
 
 /// We're essentially going to run:
 ///
-///     clang -g -O2 -target bpf -c -D__TARGET_ARCH_$(ARCH) runqslower.bpf.c -o runqslower.bpf.o
+///   clang -g -O2 -target bpf -c -D__TARGET_ARCH_$(ARCH) runqslower.bpf.c -o runqslower.bpf.o
 ///
 /// for each prog.
-fn compile(debug: bool, objs: &[UnprocessedObj], clang: &Path) -> Result<()> {
+fn compile_one(debug: bool, source: &Path, out: &Path, clang: &Path) -> Result<()> {
     let arch = if std::env::consts::ARCH == "x86_64" {
         "x86"
     } else {
         std::env::consts::ARCH
     };
 
+    if debug {
+        println!("Building {}", source.display());
+    }
+
+    let output = Command::new(clang.as_os_str())
+        .arg("-g")
+        .arg("-O2")
+        .arg("-target")
+        .arg("bpf")
+        .arg("-c")
+        .arg(format!("-D__TARGET_ARCH_{}", arch))
+        .arg(source.as_os_str())
+        .arg("-o")
+        .arg(out)
+        .output()?;
+
+    if !output.status.success() {
+        bail!(
+            "Failed to compile obj={} with status={}\n \
+            stdout=\n \
+            {}\n \
+            stderr=\n \
+            {}\n",
+            out.to_string_lossy(),
+            output.status,
+            String::from_utf8(output.stdout).unwrap(),
+            String::from_utf8(output.stderr).unwrap()
+        );
+    }
+
+    Ok(())
+}
+
+fn compile(debug: bool, objs: &[UnprocessedObj], clang: &Path) -> Result<()> {
     for obj in objs {
-        let dest_name = if let Some(f) = obj.path.as_path().file_stem() {
+        let dest_name = if let Some(f) = obj.path.file_stem() {
             let mut stem = f.to_os_string();
             stem.push(".o");
             stem
         } else {
             bail!(
                 "Could not calculate destination name for obj={}",
-                obj.path.as_path().display()
+                obj.path.display()
             );
         };
-        let mut dest_path = obj.out.clone();
+        let mut dest_path = obj.out.to_path_buf();
         dest_path.push(&dest_name);
-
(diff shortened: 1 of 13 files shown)


libbpf-cargo/src/main.rs
```diff
@@ -101,7 +101,7 @@ enum Command {
 fn main() {
```
mimullin-bbry, line 99: What would the implications be to move main to return Result<()> ?

mimullin-bbry (commented):
+1

mimullin-bbry (commented):
+1

mimullin-bbry (commented):
+1

mimullin-bbry (commented):
+1

willfindlay:
Thanks so much for addressing this. Was about to finally take a crack at it myself, but looks like that is no longer necessary :)

libbpf-cargo/src/build.rs
```diff
@@ -78,61 +78,65 @@ fn check_clang(debug: bool, clang: &Path, skip_version_checks: bool) -> Result<(
 
 /// We're essentially going to run:
 ///
-///     clang -g -O2 -target bpf -c -D__TARGET_ARCH_$(ARCH) runqslower.bpf.c -o runqslower.bpf.o
+///   clang -g -O2 -target bpf -c -D__TARGET_ARCH_$(ARCH) runqslower.bpf.c -o runqslower.bpf.o
 ///
 /// for each prog.
-fn compile(debug: bool, objs: &[UnprocessedObj], clang: &Path) -> Result<()> {
+fn compile_one(debug: bool, source: &Path, out: &Path, clang: &Path) -> Result<()> {
```
willfindlay, line 84: Would you consider adding an API to support custom compiler args as well? I have a use case where I'd like to pass -D and -I flags to clang when compiling my BPF object.

libbpf-cargo/src/main.rs
danobi: Ah yeah good idea

libbpf-cargo/src/build.rs
danobi: Yeah, for sure. I added it as the 5th commit

willfindlay (approved):
LGTM

libbpf-cargo/src/lib.rs
```diff
@@ -46,5 +55,164 @@
...
+    ///
+    /// ```no_run
+    /// use libbpf_cargo::SkeletonBuilder;
+    ///
+    /// SkeletonBuilder::new("myobject.bpf.c")
+    ///     .options("-DMACRO=value -I/some/include/dir")
+    ///     .generate("/output/path")
+    ///     .unwrap();
+    /// ```
+    pub fn options<S: AsRef<str>>(&mut self, opts: S) -> &mut SkeletonBuilder {
```
anakryiko, line 157: am I only one confused whose options it is? {compiler,clang}_{options,args,extra_args} seems more self-describing?

libbpf-cargo/src/test.rs
```diff
@@ -300,15 +269,13 @@ fn test_build_workspace_collision() {
...
-            Some(&workspace_cargo_toml),
-            Path::new("/bin/clang"),
-            true
-        ),
-        0
-    );
+    build(
+        true,
+        Some(&workspace_cargo_toml),
+        Path::new("/bin/clang"),
```
anakryiko, line 275: just curious, if you specified "clang", would it try to find it from PATH? Just wondering if it's possible that clang will be in /usr/local/bin/clang, not in /bin/clang.

anakryiko (commented):
Looks good, I just think plain .options() is bad name, because eventually we might have skeleton generator-specific options, which will be very different from clang options.

libbpf-cargo/src/lib.rs
willfindlay: Maybe `clang_args` would make the most sense

libbpf-cargo/src/test.rs
danobi: Yes, just "clang" is a path lookup

danobi:
Updated

libbpf-cargo/src/test.rs
anakryiko: if that's the case, why did you use /bin/clang instead of clang? Sorry if we already discussed this.

anakryiko (approved):
LGTM.

libbpf-cargo/src/test.rs
danobi: I suspect it's b/c I put the symlink there: https://github.com/libbpf/libbpf-rs/blob/476712481ddaa2792f27d8bf2a2a8567172d8bb4/.github/workflows/rust.yml#L17

But now that we have more people sending code in it's probably wise to do a PATH search. I'll send a follow up PR

PR #90, opened 2021-05-13
Implement progs_iter and maps_iter for Object and OpenObject

### LINKED ISSUE #77: Feature request: looping over programs in loaded Object
I think it would be really handy for non-CO-RE programs to have something like the libbpf function and macro that allows for looping over all the available programs in a BPF object. Currently I don't think the `progs` Vec is populated when an `Object` is created, so you _have_ to search for the program(s) by name. Because you have to know the program names ahead of time, you can't just load and attach programs generically.

I would guess that you could implement an iterator such that you could do the following code
```rust
for program in bpf_object.programs {
    // do something
 }
 ```
The iterator would just call `bpf_program__next` instead of trying to grab the next from the `progs` vec.

There might be some naming issues that would need to be addressed if you assume that the `progs` vec is empty by default though.

I thought about just implementing this and making a PR, but I want to see the receptiveness (since this library is "opinionated") and if any of the actual maintainers want to do this themselves.

### PR DESCRIPTION
**Pull request by willfindlay**
I want to be able to easily iterate over all BPF programs and maps in my libbpf_rs project. My use case is for easy pinning/unpinning but I imagine this capability could be useful for other use cases as well.

This PR implements `progs_iter` and `maps_iter` for `OpenObject` and `Object` to address this use case. It also adds a simple smoke test and some documentation for previously undocumented getters.

Fixes #77 Closes #79

### DIFF
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -137,24 +137,6 @@ impl Default for ObjectBuilder {
     }
 }
 
-fn find_map_in_object(
-    obj: *const libbpf_sys::bpf_object,
-    name: &str,
-) -> Result<Option<*mut libbpf_sys::bpf_map>> {
-    let c_name = util::str_to_cstring(name)?;
-    let ptr = unsafe { libbpf_sys::bpf_object__find_map_by_name(obj, c_name.as_ptr()) };
-    Ok(util::ptr_to_option(ptr))
-}
-
-fn find_prog_in_object(
-    obj: *const libbpf_sys::bpf_object,
-    name: &str,
-) -> Result<Option<*mut libbpf_sys::bpf_program>> {
-    let c_name = util::str_to_cstring(name)?;
-    let ptr = unsafe { libbpf_sys::bpf_object__find_program_by_name(obj, c_name.as_ptr()) };
-    Ok(util::ptr_to_option(ptr))
-}
-
 /// Represents an opened (but not yet loaded) BPF object file.
 ///
 /// Use this object to access [`OpenMap`]s and [`OpenProgram`]s.
@@ -248,21 +230,13 @@ impl OpenObject {
     }
 
     /// Get a mutable reference to `OpenMap` with the name `name`, if one exists.
-    pub fn map<T: AsRef<str>>(&mut self, name: T) -> Result<Option<&mut OpenMap>> {
-        if self.maps.contains_key(name.as_ref()) {
-            Ok(self.maps.get_mut(name.as_ref()))
-        } else if let Some(ptr) = find_map_in_object(self.ptr, name.as_ref())? {
-            self.maps
-                .insert(name.as_ref().to_owned(), OpenMap::new(ptr));
-            Ok(self.maps.get_mut(name.as_ref()))
-        } else {
-            Ok(None)
-        }
+    pub fn map<T: AsRef<str>>(&mut self, name: T) -> Option<&mut OpenMap> {
+        self.maps.get_mut(name.as_ref())
     }
 
     /// Same as [`OpenObject::map`] except will panic if `Err` or `None` is encountered.
     pub fn map_unwrap<T: AsRef<str>>(&mut self, name: T) -> &mut OpenMap {
-        self.map(name).unwrap().unwrap()
+        self.map(name).unwrap()
     }
 
     /// Get an iterator over references to all `OpenMap`s.
@@ -273,21 +247,13 @@ impl OpenObject {
     }
 
     /// Get a mutable reference to `OpenProgram` with the name `name`, if one exists.
-    pub fn prog<T: AsRef<str>>(&mut self, name: T) -> Result<Option<&mut OpenProgram>> {
-        if self.progs.contains_key(name.as_ref()) {
-            Ok(self.progs.get_mut(name.as_ref()))
-        } else if let Some(ptr) = find_prog_in_object(self.ptr, name.as_ref())? {
-            let owned_name = name.as_ref().to_owned();
-            self.progs.insert(owned_name, OpenProgram::new(ptr));
-            Ok(self.progs.get_mut(name.as_ref()))
-        } else {
-            Ok(None)
-        }
+    pub fn prog<T: AsRef<str>>(&mut self, name: T) -> Option<&mut OpenProgram> {
+        self.progs.get_mut(name.as_ref())
     }
 
     /// Same as [`OpenObject::prog`] except will panic if `Err` or `None` is encountered.
     pub fn prog_unwrap<T: AsRef<str>>(&mut self, name: T) -> &mut OpenProgram {
-        self.prog(name).unwrap().unwrap()
+        self.prog(name).unwrap()
     }
 
     /// Get an iterator over references to all `OpenProgram`s.
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -6,8 +6,6 @@ use std::os::raw::c_char;
 use std::path::Path;
 use std::ptr;
 
-use nix::errno;
-
 use crate::util;
 use crate::*;
 
@@ -98,7 +96,7 @@ impl ObjectBuilder {
             return Err(Error::System(err as i32));
         }
 
-        Ok(OpenObject::new(obj))
+        OpenObject::new(obj)
     }
 
     pub fn open_memory<T: AsRef<str>>(&mut self, name: T, mem: &[u8]) -> Result<OpenObject> {
@@ -126,7 +124,7 @@ impl ObjectBuilder {
             return Err(Error::System(err as i32));
         }
 
-        Ok(OpenObject::new(obj))
+        OpenObject::new(obj)
     }
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by willfindlay** [comment]
Oops, looks like this was already attempted in #79 but this solution seems much simpler.

**Comment by rtkaratekid** [comment]
@willfindlay This looks good. The only issue is that the maps and progs HashMaps are not populated on `new()`. So if you build and load an object and then call your functions, you'll get an empty iterator. Unless I'm missing something?

So I do actually like the return type approach, but if you took my PR and included the `libbpf_sys` code that fetches all the programs and maps and puts them in the `HashMap` and **then** returned your values, I think that would be the best of the two PRs. Feel free to do that if you would like, I'm unfortunately short on free time for the next... couple weeks ish.

@danobi what do you think? I like this approach more than the macro just because the code is a little more readable.

**Comment by willfindlay** [comment]
@rtkaratekid Ah, I see the problem now. The `map()` and `prog()` methods are populating the hashmap lazily. Seems like the best approach would be combining our efforts somehow. Since you're short on free time, I'm happy to take the lead on this. I'll convert this to a draft for now since it's not actually working as intended.

**Comment by rtkaratekid** [comment]
I'm happy to review and test

**Review by danobi — COMMENTED** [review]
I don't think this works. `self.maps` and `self.progs` are lazily populated by `map()` and `prog()`. So for example, if `map()` isn't ever called, then `maps_iter_mut()` won't return anythinng

**Comment by danobi** [comment]
Ah we all commented at the same time.

> @danobi what do you think? I like this approach more than the macro just because the code is a little more readable.

Yeah I think this approach is better than the macro

**Comment by willfindlay** [comment]
@danobi I'll find some time tomorrow to refactor this into something that actually works. Will update this PR when it's ready.

**Comment by willfindlay** [comment]
FWIW is there some technical reason for this lazy population? Would you be opposed to removing it entirely?

**Comment by willfindlay** [comment]
@danobi What would you think of something like this (available on my fork: https://github.com/willfindlay/libbpf-rs/tree/progs-maps-iter-populate-in-new). Removes lazy map population and moves it into Object::new() and OpenObject::new().

The alternative is keeping this lazy population and doing all the bpf_xxxx__next() calls every time the user calls xxxx_iter_mut(). I think the former option is better, but I'm open to discussion.

**Comment by danobi** [comment]
> FWIW is there some technical reason for this lazy population? Would you be opposed to removing it entirely?

No reason. It's just what made sense at the time. We can change it.

> @danobi What would you think of something like this (available on my fork: https://github.com/willfindlay/libbpf-rs/tree/progs-maps-iter-populate-in-new). Removes lazy map population and moves it into Object::new() and OpenObject::new().

Looks totally reasonable. It's an API break but I think that's acceptable since this library has been pretty up front about having an unstable API.

**Comment by willfindlay** [comment]
> > FWIW is there some technical reason for this lazy population? Would you be opposed to removing it entirely?
> 
> No reason. It's just what made sense at the time. We can change it.
> 
> > @danobi What would you think of something like this (available on my fork: https://github.com/willfindlay/libbpf-rs/tree/progs-maps-iter-populate-in-new). Removes lazy map population and moves it into Object::new() and OpenObject::new().
> 
> Looks totally reasonable. It's an API break but I think that's acceptable since this library has been pretty up front about having an unstable API.

Sounds good to me. I'll push the changes over to this branch and re-open the PR.

**Comment by rtkaratekid** [comment]
I haven't tested it but the code does look good.

File: libbpf-rs/src/object.rs [hunk 1]
```diff
@@ -284,12 +344,87 @@ pub struct Object {
...
+
+            // Get the map def
+            // bpf_map__def can return null but only if it's passed a null.
+            // We already know next_ptr is not null.
+            let def = unsafe { ptr::read(libbpf_sys::bpf_map__def(next_ptr)) };
+
+            // Get the map fd
+            let fd = unsafe { libbpf_sys::bpf_map__fd(next_ptr) };
+            if fd < 0 {
+                return Err(Error::System(errno::errno()));
```
danobi, line 379: I think the existing code is wrong. libbpf returns the negative errno on errors: https://github.com/libbpf/libbpf/blob/c5389a965bc3f19e07b1ee161092fc227e364e94/src/libbpf.c#L9382-L9385

```suggestion
                return Err(Error::System(-fd));
```

libbpf-rs/src/object.rs
```diff
@@ -284,12 +344,87 @@ pub struct Object {
...
+            let err = unsafe { libbpf_sys::libbpf_get_error(title as *const _) };
+            if err != 0 {
+                return Err(Error::System(err as i32));
+            }
+            let section = util::c_ptr_to_string(title)?;
+
+            // Get the program fd
+            let fd = unsafe { libbpf_sys::bpf_program__fd(next_ptr) };
+            if fd < 0 {
+                return Err(Error::System(errno::errno()));
```
danobi, line 418: Same as above

libbpf-rs/src/object.rs
```diff
@@ -300,65 +435,40 @@ impl Object {
...
+        self.maps.get_mut(name.as_ref())
     }
 
-    /// Same as [`Object::map`] except will panic if `Err` or `None` is encountered.
-    pub fn map_unwrap<T: AsRef<str>>(&mut self, name: T) -> &mut Map {
-        self.map(name).unwrap().unwrap()
+    /// Get an iterator over references to all `Map`s.
+    /// Note that this will include automatically generated .data, .rodata, .bss, and
+    /// .kconfig maps. You may wish to filter these out depending on your use case.
+    pub fn maps_iter_mut(&mut self) -> impl Iterator<Item = &mut Map> {
```
danobi, line 450: `map()` and `prog()` already return mutable references without `mut` in their name. Maybe rename to `maps_iter()`? However, the API guidelines say otherwise: https://rust-lang.github.io/api-guidelines/naming.html#methods-on-collections-that-produce-iterators-follow-iter-iter_mut-into_iter-c-iter . Don't have a strong opinion here, wondering what others think

willfindlay: I was a bit torn on this for this reason. I do think we should try and be as consistent as possible with the API guidelines. When I see something like `maps_iter` I think I am going to get something immutable back.

willfindlay:
I've fixed the fd error handling. A quick grep reveals a few other instances in the existing code where it is being handled incorrectly. I'll make a follow-up PR to address those.

libbpf-rs/src/object.rs
willfindlay: Something else to consider would be whether it's worth supporting the ability to grab immutable references to maps.

rtkaratekid: For maps specifically, it might be worthwhile to have an immutable reference because there are some useful functions that can be used with just a reference. For programs though, it seems much less useful to get an immutable reference. Just my two cents.

I would be okay with the explicit naming, but if there isn't explicit naming I think the library should probably just try to be consistent.

willfindlay: If we do end up adding immutable versions, perhaps that should be in a separate PR (possibly with changes to generated skeleton file as well?). I do agree that immutable map references would be useful for many cases.

willfindlay: For now, I'll revert my current naming convention back to `maps_iter` and `progs_iter` and then we can address immutability in a future PR.

danobi (approved):
LGTM

libbpf-rs/src/object.rs
```diff
@@ -167,12 +165,58 @@ pub struct OpenObject {
...
+        let mut map: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
+        loop {
+            // Get the pointer to the next BPF map
+            let next_ptr = unsafe { libbpf_sys::bpf_map__next(map, obj.ptr) };
+            if next_ptr.is_null() {
+                break;
+            }
+
+            // Get the map name
+            let name = unsafe { libbpf_sys::bpf_map__name(next_ptr) };
```
anakryiko, line 185: bpf_map__name never returns NULL if you pass non-NULL map pointer. Even when you pass NULL it returns NULL, so there is no need to use libbpf_get_error. So please just simplify all this.

libbpf-rs/src/object.rs
```diff
@@ -167,12 +165,58 @@ pub struct OpenObject {
...
+        let mut prog: *mut libbpf_sys::bpf_program = std::ptr::null_mut();
+        loop {
+            // Get the pointer to the next BPF program
+            let next_ptr = unsafe { libbpf_sys::bpf_program__next(prog, obj.ptr) };
+            if next_ptr.is_null() {
+                break;
+            }
+
+            // Get the program name
+            let name = unsafe { libbpf_sys::bpf_program__name(next_ptr) };
```
anakryiko, line 207: it's even stricter for bpf_program__name, it doesn't expect NULL and never returns NULL

libbpf-rs/src/object.rs
```diff
@@ -225,6 +270,14 @@ impl OpenObject {
         self.map(name).unwrap().unwrap()
     }
 
+    /// Get an iterator over references to all `OpenMap`s.
+    /// Note that this will include automatically generated .data, .rodata, .bss, and
+    /// .kconfig maps.
+    pub fn maps_iter(&mut self) -> impl Iterator<Item = &mut OpenMap> {
```
anakryiko, line 276: I'm wondering how mutable and read-only iterators are generally implemented? E.g., if I just want to print out all map and prog names and I don't have mutable reference to OpenObject, how can I do that?

anakryiko (changes requested):

libbpf-rs/src/object.rs
willfindlay: That would be with an iterator that returns an immutable reference. We discussed this above somewhere. Right now I'm thinking it makes more sense to include such functionality in a second PR.

willfindlay: Signature would be like this: `pub fn maps_iter(&self) -> impl Iterator<Item = &OpenMap>` and then the body would just return `self.maps.values()`.

willfindlay: My plan is to eventually propose changing all existing map and prog getters to have a `_mut` suffix and writing an immutable version for each. This goes a bit beyond the scope of this PR, which is why I wanted to do it in a follow-up.

libbpf-rs/src/object.rs
willfindlay: Fixed in latest commit

libbpf-rs/src/object.rs
willfindlay: Fixed in latest commit

libbpf-rs/src/object.rs
anakryiko: I was more concerned with naming, because I think Rust doesn't allow method overrides. But I guess we are still at the point where it's not too bad to change APIs and rename them, so ok.

libbpf-rs/src/object.rs
```diff
@@ -284,12 +337,82 @@ pub struct Object {
...
+            }
+
+            // Get the program name
+            // bpf_program__name never returns NULL, so no need to check the pointer.
+            let name = unsafe { libbpf_sys::bpf_program__name(next_ptr) };
+            let name = util::c_ptr_to_string(name)?;
+
+            // Get the program section
+            let title = unsafe { libbpf_sys::bpf_program__title(next_ptr, false) };
+            let err = unsafe { libbpf_sys::libbpf_get_error(title as *const _) };
```
anakryiko, line 397: missed this one?

anakryiko (approved):
One more place with unnecessray libbpf_get_error() use, but otherwise LGTM, so I approved pro-actively.

willfindlay:
> I was more concerned with naming, because I think Rust doesn't allow method overrides. But I guess we are still at the point where it's not too bad to change APIs and rename them, so ok.

I originally named them with `_mut` suffixes but we agreed to change it in one of the threads above ;)

willfindlay:
@danobi I fixed one last nit from Andrii and rebased into one commit. Should be good to merge when you're ready.

willfindlay:
I realized I forgot to simplify the getter logic for OpenObject in the same way that I did for Object. Added another commit that fixes this. Ended up deleting a few helper functions from object.rs which are no longer needed. The `ptr_to_option` helper in utils.rs is technically dead code now, but I left it in since it might be useful in the future.

libbpf-rs/src/util.rs
```diff
@@ -28,6 +28,7 @@ pub fn c_ptr_to_string(p: *const c_char) -> Result<String> {
         .to_owned())
 }
 
+#[allow(dead_code)] // This might be useful in the future
```
danobi, line 31: Let's just delete this. The logic is not complicated and can easily be added later.

willfindlay:
@danobi Updated to remove `ptr_to_option`

danobi:
Thank you!

PR #91, opened 2021-05-14
libbpf-rs: Fix handling of cstrings in several functions

willfindlay:
Several functions in program.rs were not handling cstrings correctly. This caused some headaches in my project due to strings not being null-terminated. I ended up null terminating them myself as a workaround, but this is obviously far from ideal. To fix this, I've updated all the offending functions to use `xxx_to_cstring` helper functions exposed in util.rs.

I also noticed that `T` in `attach_uprobe` really should be an `AsRef<Path>`, not an `AsRef<str>`. Fixed this as well.

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -1,5 +1,4 @@
 use std::convert::TryFrom;
-use std::os::raw::c_char;
 use std::path::Path;
 
 use nix::errno;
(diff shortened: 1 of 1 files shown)


masmullin2000 (commented):
+1

danobi (approved):
Ah, nice. Thanks for the fixes

PR #93, opened 2021-05-15
libbpf-rs: Add immutable getters

willfindlay:
This is a draft proposal to refactor existing Object and OpenObject getters into mutable and immutable variants. It is based on top of my PR for maps and program iterators and should not be landed until that is merged. Existing getters are renamed with the `_mut` suffix and new getters are created for the immutable variants, using the original names. For now, libbpf-cargo/gen.rs has been changed to emit `xxx_unwrap_mut()` instead of `xxx_unwrap()`.

I have a couple of points that I'd like to discuss before making any further progress on this.

1. We should probably add mutable/immutable variants to the generated skeleton in addition to Object and OpenObject.
    - As I see it, this could be done in one of two ways. Either add an extra struct like MapsMut that would be accessed using something like `maps_mut()` or add additional getters to Maps with the `_mut` suffix.
2. Perhaps it's also worth adding `xxx_iter()` and `xxx_iter_mut()` to the skeleton as a thin wrapper around the methods exposed by Object and OpenObject.
3. Is it even worth keeping the `xxx_unwrap()` convenience functions? Since the getters now only return an `Option<T>` instead of a `Result<Option<T>>`, it may make more sense to just have the consumer of the API call unwrap manually.

Let me know if there's anything I missed/overlooked here.

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -159,12 +159,8 @@ fn map_is_mmapable(map: *const libbpf_sys::bpf_map) -> bool {
 fn map_is_datasec(map: *const libbpf_sys::bpf_map) -> bool {
     let internal = unsafe { libbpf_sys::bpf_map__is_internal(map) };
     let mmapable = map_is_mmapable(map);
-    let has_datasec_name = match get_map_name(map).unwrap().as_ref().map(String::as_str) {
-        Some("rodata") | Some("data") | Some("bss") | Some("kconfig") => true,
-        _ => false,
-    };
 
-    internal && mmapable && has_datasec_name
+    internal && mmapable
 }
 
 fn map_is_readonly(map: *const libbpf_sys::bpf_map) -> bool {
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -261,6 +261,76 @@ fn gen_skel_map_defs(
         skel,
         r#"
         pub struct {struct_name}<'a> {{
+            inner: &'a  {inner_ty},
+        }}
+
+        impl<'a> {struct_name}<'a> {{
+        "#,
+        inner_ty = inner_ty,
+        struct_name = struct_name
+    )?;
+
+    for map in MapIter::new(object) {
+        if map_is_mmapable(map) {
+            continue;
+        }
+
+        let map_name = match get_map_name(map)? {
+            Some(n) => n,
+            None => continue,
+        };
+
+        write!(
+            skel,
+            r#"
+            pub fn {map_name}(&self) -> &{return_ty} {{
+                self.inner.map("{raw_map_name}").unwrap()
+            }}
+            "#,
+            map_name = map_name,
+            raw_map_name = get_raw_map_name(map)?,
+            return_ty = return_ty,
+        )?;
+    }
+
+    writeln!(skel, "}}")?;
+
+    Ok(())
+}
+
+fn gen_skel_map_defs_mut(
+    skel: &mut String,
+    object: *mut libbpf_sys::bpf_object,
+    obj_name: &str,
+    open: bool,
+) -> Result<()> {
+    // If no non-datasec maps, return early
+    if MapIter::new(object)
+        .filter(|map| !map_is_mmapable(*map))
+        .count()
+        == 0
+    {
+        return Ok(());
+    }
+
+    let (struct_name, inner_ty, return_ty) = if open {
+        (
+            format!("Open{}MapsMut", obj_name),
+            "libbpf_rs::OpenObject",
+            "libbpf_rs::OpenMap",
+        )
+    } else {
+        (
+            format!("{}MapsMut", obj_name),
+            "libbpf_rs::Object",
+            "libbpf_rs::Map",
+        )
+    };
+
+    write!(
+        skel,
+        r#"
+        pub struct {struct_name}<'a> {{
             inner: &'a mut {inner_ty},
         }}
 
@@ -326,7 +396,7 @@ fn gen_skel_prog_defs(
         skel,
         r#"
         pub struct {struct_name}<'a> {{
-            inner: &'a mut {inner_ty},
+            inner: &'a {inner_ty},
         }}
 
         impl<'a> {struct_name}<'a> {{
@@ -335,6 +405,60 @@ fn gen_skel_prog_defs(
         struct_name = struct_name
     )?;
 
+    for prog in ProgIter::new(object) {
+        write!(
+            skel,
+            r#"
+            pub fn {prog_name}(&self) -> &{return_ty} {{
+                self.inner.prog("{prog_name}").unwrap()
+            }}
+            "#,
+            prog_name = get_prog_name(prog)?,
+            return_ty = return_ty,
+        )?;
+    }
+
+    writeln!(skel, "}}")?;
+
+    Ok(())
+}
+
+fn gen_skel_prog_defs_mut(
+    skel: &mut String,
+    object: *mut libbpf_sys::bpf_object,
+    obj_name: &str,
+    open: bool,
+) -> Result<()> {
+    if ProgIter::new(object).next().is_none() {
+        return Ok(());
+    }
+
+    let (struct_name, inner_ty, return_ty) = if open {
+        (
+            format!("Open{}ProgsMut", obj_name),
+            "libbpf_rs::OpenObject",
+            "libbpf_rs::OpenProgram",
+        )
+    } else {
+        (
+            format!("{}ProgsMut", obj_name),
+            "libbpf_rs::Object",
+            "libbpf_rs::Program",
+        )
+    };
+
+    write!(
+        skel,
+        r#"
+        pub struct {struct_name}<'a> {{
+            inner: &'a mut {inner_ty},
+        }}
+
+        impl<'a> {struct_name}<'a> {{
+        "#,
+        inner_ty = inner_ty,
+        struct_name = struct_name
+    )?;
     for prog in ProgIter::new(object) {
         write!(
             skel,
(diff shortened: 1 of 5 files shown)


libbpf-cargo/src/gen.rs
```diff
@@ -271,7 +271,7 @@ fn gen_skel_map_defs(
             skel,
             r#"
             pub fn {map_name}(&mut self) -> &mut {return_ty} {{
```
danobi, line 273: I wonder if the skeleton should have immutable/mutable getters too. Gut feeling is yeah, for consistency.

willfindlay: I'm thinking it should. Have you had a chance to consider the two options I proposed above?

anakryiko:
>     1. We should probably add mutable/immutable variants to the generated skeleton in addition to Object and OpenObject.
>        
>        * As I see it, this could be done in one of two ways. Either add an extra struct like MapsMut that would be accessed using something like `maps_mut()` or add additional getters to Maps with the `_mut` suffix.

As in `skel.maps().events()`/`skel.maps_mut().events()` vs `skel.maps().events()`/`skel.maps().events_mut()`? Feels like `skel.maps().events_mut()` is a better way to go (and probably will result in nicer skeleton)?

> 
>     2. Perhaps it's also worth adding `xxx_iter()` and `xxx_iter_mut()` to the skeleton as a thin wrapper around the methods exposed by Object and OpenObject.

Not sure about this. If you can access Object and/or OpenObject from skeleton easily, why duplicate all the APIs? What other bpf_object APIs are we going to expose through skeleton next?

> 
>     3. Is it even worth keeping the `xxx_unwrap()` convenience functions? 

_unwrap() variants look weird to me, given they don't save much typing but 2x number of exposed API. And it's a generally bad API that panics, no? :)

willfindlay:
> > ```
> > 1. We should probably add mutable/immutable variants to the generated skeleton in addition to Object and OpenObject.
> >    
> >    * As I see it, this could be done in one of two ways. Either add an extra struct like MapsMut that would be accessed using something like `maps_mut()` or add additional getters to Maps with the `_mut` suffix.
> > ```
> 
> As in `skel.maps().events()`/`skel.maps_mut().events()` vs `skel.maps().events()`/`skel.maps().events_mut()`? Feels like `skel.maps().events_mut()` is a better way to go (and probably will result in nicer skeleton)?

Right exactly. I think both are certainly viable, although I personally prefer the former rather than the latter.
 
> > ```
> > 2. Perhaps it's also worth adding `xxx_iter()` and `xxx_iter_mut()` to the skeleton as a thin wrapper around the methods exposed by Object and OpenObject.
> > ```
>
> Not sure about this. If you can access Object and/or OpenObject from skeleton easily, why duplicate all the APIs? What other bpf_object APIs are we going to expose through skeleton next?

Good point. Let's just leave it in `Object` and `OpenObject` then.

> > ```
> > 3. Is it even worth keeping the `xxx_unwrap()` convenience functions? 
> > ```
> 
> _unwrap() variants look weird to me, given they don't save much typing but 2x number of exposed API. And it's a generally bad API that panics, no? :)

This was my thinking too. It made more sense _originally_ as it abstracted away two calls to `unwrap()` but even then it seems quite dubious to me. If nobody has any objections I will remove those along with this PR.

libbpf-cargo/src/gen.rs
danobi: Oh whoops, I think I just immediately clicked over to "files changed" without reading

danobi:
> > > ```
> > > 1. We should probably add mutable/immutable variants to the generated skeleton in addition to Object and OpenObject.
> > >    
> > >    * As I see it, this could be done in one of two ways. Either add an extra struct like MapsMut that would be accessed using something like `maps_mut()` or add additional getters to Maps with the `_mut` suffix.
> > > ```
> > 
> > 
> > As in `skel.maps().events()`/`skel.maps_mut().events()` vs `skel.maps().events()`/`skel.maps().events_mut()`? Feels like `skel.maps().events_mut()` is a better way to go (and probably will result in nicer skeleton)?
> 
> Right exactly. I think both are certainly viable, although I personally prefer the former rather than the latter.

I prefer the former as well. An exact identifier match is useful for people poking around in unfamilar code or use the `*` vim command that does a `\<TOKEN\>` search. vim's `*` will not match for `TOKEN_suffix`

> 
> > > ```
> > > 2. Perhaps it's also worth adding `xxx_iter()` and `xxx_iter_mut()` to the skeleton as a thin wrapper around the methods exposed by Object and OpenObject.
> > > ```
> > 
> > 
> > Not sure about this. If you can access Object and/or OpenObject from skeleton easily, why duplicate all the APIs? What other bpf_object APIs are we going to expose through skeleton next?
> 
> Good point. Let's just leave it in `Object` and `OpenObject` then.

:+1: 

> 
> > > ```
> > > 3. Is it even worth keeping the `xxx_unwrap()` convenience functions? 
> > > ```
> > 
> > 
> > _unwrap() variants look weird to me, given they don't save much typing but 2x number of exposed API. And it's a generally bad API that panics, no? :)
> 
> This was my thinking too. It made more sense _originally_ as it abstracted away two calls to `unwrap()` but even then it seems quite dubious to me. If nobody has any objections I will remove those along with this PR.

Yeah, agreed. Please remove

anakryiko:
> > > As in `skel.maps().events()`/`skel.maps_mut().events()` vs `skel.maps().events()`/`skel.maps().events_mut()`? Feels like `skel.maps().events_mut()` is a better way to go (and probably will result in nicer skeleton)?
> > 
> > 
> > Right exactly. I think both are certainly viable, although I personally prefer the former rather than the latter.
> 
> I prefer the former as well. An exact identifier match is useful for people poking around in unfamilar code or use the `*` vim command that does a `\<TOKEN\>` search. vim's `*` will not match for `TOKEN_suffix`

Ok, sounds good. I didn't have strong preferences anyways.

willfindlay:
Alright, I'll make those changes and remove the draft tag when it's ready for review.

willfindlay:
Alright, should be all set to review now.

libbpf-cargo/src/gen.rs
```diff
@@ -248,6 +248,76 @@ fn gen_skel_map_defs(
...
+        impl<'a> {struct_name}<'a> {{
+        "#,
+        inner_ty = inner_ty,
+        struct_name = struct_name
+    )?;
+
+    for map in MapIter::new(object) {
+        if map_is_mmapable(map) {
+            continue;
+        }
```
anakryiko, line 267: hm... why this check? the fact that map is mmapable has nothing to do with whether we should have accessor for it or not. @danobi, I see similar check above, why are we skipping mmapable maps. They might be .data/.rodata/.bss, in which case it's still generally useful to have map accessor for them. But BPF_F_MMAPABLE can be specified by users explicitly, in which case omitting such maps would be bad as well.

willfindlay: This is how it was done in the original skeleton generation logic. What do you think about replacing this with a check against map name (i.e. does it end in a `.bss`, `.data`, `.rodata`)?

anakryiko: This is original skeleton generation logic: https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/gen.c#L380-L404 :) let's keep them in sync

willfindlay: > This is original skeleton generation logic: https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/gen.c#L380-L404 :) let's keep them in sync

Haha, that's a fair point! Just meant this wasn't anything new I added :)

willfindlay: @anakryiko So you think we should just remove the check entirely then? So the .bss map would be included in both skel.maps().bss() and skel.bss() for example.

anakryiko: I vote yes, I don't like when tool make arbitrary decisions on what should or shouldn't be available to users. Those internal maps are still maps, so should be represented in maps().

willfindlay: This makes sense to me. @danobi any thoughts?

danobi: Sure, yeah. Add them. Don't remember why I originally did this but it was probably b/c I thought it was being mmap'd so no need to expose the actual maps.

willfindlay:
Updated to remove old logic for filtering out datasec maps. Also added a new helper `map_is_datasec` to differentiate datasec maps from mmapable maps. Finally, updated tests to verify that the new `_mut()` getters work.

libbpf-cargo/src/gen.rs
```diff
@@ -143,12 +143,19 @@ fn get_prog_name(prog: *const libbpf_sys::bpf_program) -> Result<String> {
...
+fn map_is_datasec(map: *const libbpf_sys::bpf_map) -> bool {
+    let internal = unsafe { libbpf_sys::bpf_map__is_internal(map) };
+    let mmapable = map_is_mmapable(map);
+    let has_datasec_name = match get_map_name(map).unwrap().as_ref().map(String::as_str) {
+        Some("rodata") | Some("data") | Some("bss") | Some("kconfig") => true,
+        _ => false,
+    };
 
-    internal && (mmapable > 0) && (name.is_ok() && name.unwrap().is_some())
+    internal && mmapable && has_datasec_name
```
danobi, line 158: Let's not duplicate the hardcoded list in more than 1 place.

I think this can be simplified to:
```
let has_name = get_map_name(map).is_some();

internal && mmapable && has_name
```

Because if `internal` is true then a `Some(..)` returned by `get_map_name()` is already a valid internal map name

danobi (commented):
For commit 3 (libbpf-cargo: Split skeleton map and prog getters into mutable and immutable), instead of having largely duplicated code, can we add a `mutable` flag to the generator function? Like how the `open` flag works.

willfindlay:
> For commit 3 (libbpf-cargo: Split skeleton map and prog getters into mutable and immutable), instead of having largely duplicated code, can we add a `mutable` flag to the generator function? Like how the `open` flag works.

This is actually what I did originally, but I ended up not being happy with it so I decided to refactor into this. I'm happy to switch it back though.

libbpf-cargo/src/gen.rs
willfindlay: Simplified

willfindlay:
@danobi Done refactoring to use mutable flag. I ended coming up with something I like much better.

libbpf-cargo/src/gen.rs
```diff
@@ -303,20 +303,27 @@ fn gen_skel_prog_defs(
     object: *mut libbpf_sys::bpf_object,
     obj_name: &str,
     open: bool,
+    mutable: bool,
 ) -> Result<()> {
     if ProgIter::new(object).next().is_none() {
         return Ok(());
     }
```
willfindlay, line 310: I should probably add something like this to `gen_skel_map_defs`. I'll do it now.

willfindlay:
Oops, CI is failing somehow, let me fix.

libbpf-cargo/src/gen.rs
```diff
@@ -152,12 +152,15 @@ fn get_prog_name(prog: *const libbpf_sys::bpf_program) -> Result<String> {
...
-    let name = get_map_name(map);
+    (unsafe { (*def).map_flags } & libbpf_sys::BPF_F_MMAPABLE) > 0
+}
 
-    internal && (mmapable > 0) && (name.is_ok() && name.unwrap().is_some())
+fn map_is_datasec(map: *const libbpf_sys::bpf_map) -> bool {
+    let internal = unsafe { libbpf_sys::bpf_map__is_internal(map) };
+    let mmapable = map_is_mmapable(map);
+
+    internal && mmapable
```
danobi, line 163: Ah yeah, I think this is better

danobi (approved):
LGTM

PR #94, opened 2021-05-17
libbpf-cargo: tests: Use clang from $PATH

danobi:
More likely to work on different systems. I think the tests were
originally hardcoded b/c the CI had a fixed symlink at /bin/clang.

### DIFF
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -141,17 +141,17 @@ fn test_build_default() {
     let (_dir, proj_dir, cargo_toml) = setup_temp_project();
 
     // No bpf progs yet
-    build(true, Some(&cargo_toml), Path::new("/bin/clang"), true).unwrap_err();
+    build(true, Some(&cargo_toml), None, true).unwrap_err();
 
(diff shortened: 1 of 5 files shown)


libbpf-cargo/src/test.rs
```diff
@@ -702,7 +684,7 @@ fn test_skeleton_builder_basic() {
     let skel = NamedTempFile::new().unwrap();
     SkeletonBuilder::new(proj_dir.join("src/bpf/prog.bpf.c"))
         .debug(true)
-        .clang("/bin/clang")
+        .clang("clang")
```
anakryiko, line 687: this should probably be a default value and shouldn't be required to be specified explicitly

danobi: Yeah, it should. I'll change it

danobi:
Updated

libbpf-cargo/src/test.rs
anakryiko: Maybe drop `.clang("clang")` in one of the cases to make sure that still works?

PR #96, opened 2021-05-18
runqslower: Place generated skeleton in .output directory

danobi:
In line with how libbpf-bootstrap and libbpf-tools does it.

### DIFF
--- a/examples/runqslower/src/main.rs
+++ b/examples/runqslower/src/main.rs
@@ -8,8 +8,9 @@ use libbpf_rs::PerfBufferBuilder;
 use plain::Plain;
 use structopt::StructOpt;
 
-mod bpf;
-use bpf::*;
+#[path = "bpf/.output/runqslower.skel.rs"]
+mod runqslower;
+use runqslower::*;
 
 /// Trace high run queue latency
 #[derive(Debug, StructOpt)]
--- a/examples/runqslower/.gitignore
+++ b/examples/runqslower/.gitignore
@@ -1,1 +1,1 @@
-src/bpf/mod.rs
+src/bpf/.output
--- a/examples/runqslower/build.rs
+++ b/examples/runqslower/build.rs
@@ -1,3 +1,4 @@
+use std::fs::create_dir_all;
 use std::path::Path;
 
 use libbpf_cargo::SkeletonBuilder;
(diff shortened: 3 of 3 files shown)


examples/runqslower/src/main.rs
```diff
@@ -8,6 +8,7 @@ use libbpf_rs::PerfBufferBuilder;
 use plain::Plain;
 use structopt::StructOpt;
 
+#[path = "bpf/.output/mod.rs"]
```
anakryiko, line 11: does it have to be called generic "mod.rs"? What will happen if application is using two different skeletons? Ideally each skeleton should be called with its name, something like `runqslower.skel.rs` and `some_other.skel.rs`

danobi: That is controlled in the build script: https://github.com/libbpf/libbpf-rs/blob/699e2a7e6b3e49e212902ce95314f850e2ffdf4e/examples/runqslower/build.rs#L16-L17

If there's multiple skeletons, the user is free to name each one whatever they want. That's the nice thing about the `#[path=..]` attribute

anakryiko: cool, can we please use `runqslower.skel.rs` to set a good example?

danobi: Done

examples/runqslower/src/main.rs
```diff
@@ -8,6 +8,7 @@ use libbpf_rs::PerfBufferBuilder;
 use plain::Plain;
 use structopt::StructOpt;
 
+#[path = "bpf/.output/runqslower.skel.rs"]
 mod bpf;
```
anakryiko, line 12: hm.. I hope user can also control`mod bpf` name, but I'm obnoxious enough already, so let's land this for now.

anakryiko (approved):

examples/runqslower/src/main.rs
willfindlay: They should be able to. Here the `#[path]` decorator here is overloading the path for `mod bpf`, but in principle, you could name the generated skeleton whatever you like and the module name would correspond.

danobi:
Updated to use `runqslower` instead of `bpf` for module name

PR #98, opened 2021-05-19
libbpf-rs/program.rs: Fix lifetime issue with CString pointers

willfindlay:
My original CStrings fix had a really bad lifetime issue that caused a use-after-free error.
The details are outlined here: https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_ptr
Update affected functions to store the CString and raw pointer as separate bindings to prevent the early free.

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -235,13 +235,14 @@ impl Program {
         binary_path: T,
         func_offset: usize,
     ) -> Result<Link> {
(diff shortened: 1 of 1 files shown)


willfindlay:
I think we should test these functions to prevent regressions. I can work on a PR to add tests when I have some more free time. For now, this PR works on my project without errors.

danobi:
Ouch. Thanks for the fix.

> I can work on a PR to add tests when I have some more free time

That'd be great

danobi (approved):

PR #104, opened 2021-06-01
libbpf-rs/link.rs: Add Link::disconnect()

willfindlay:
Add the ability to "disconnect" a BPF Link to prevent the underlying program from being detached when the Link is dropped.


danobi:
Seems reasonable to me. I'm not very familiar with this API though so I'll defer to @anakryiko

anakryiko (approved):
LGTM.

PR #106, opened 2021-06-02
libbpf-cargo: bump cargo_metadata to 0.12

davide125:
Needed to package `libbpf-cargo` in Fedora

### DIFF
--- a/libbpf-cargo/Cargo.toml
+++ b/libbpf-cargo/Cargo.toml
@@ -25,7 +25,7 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = "1.0"
-cargo_metadata = "0.9"
+cargo_metadata = "0.12"
 libbpf-sys = { version = "0.3.0-2" }
 memmap = "0.7"
 num_enum = "0.5"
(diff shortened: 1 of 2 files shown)


libbpf-cargo/Cargo.toml
```diff
@@ -25,7 +25,7 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = "1.0"
-cargo_metadata = "0.9"
+cargo_metadata = "0.12.3"
```
danobi, line 28: Does it have to be exactly 0.12.3? I would usually prefer to leave it at 0.12 so we can pick up bug fix patches.

danobi:
Hmm I think `Cargo.lock` is usually updated with this kind of change too. Maybe I'm mistaken, but try running `cargo build` to see if Cargo.lock is updated

libbpf-cargo/Cargo.toml
davide125: Oh yeah that's fine too

danobi (approved):

PR #109, opened 2021-06-02
libbpf-cargo: Fix compilation on non-x86_64 archs

### LINKED ISSUE #108: libbpf-cargo fails to build for non-x86_64 architectures
See https://koji.fedoraproject.org/koji/taskinfo?taskID=69201098 for examples (look at `build.log`)

willfindlay:
libbpf-cargo was making assumptions about the width of c_ulong and signedness of c_char in some places. This was causing compilation to fail on some other archs. Fixes #108

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -4,6 +4,7 @@ use std::convert::TryFrom;
 use std::ffi::{c_void, CStr, CString};
 use std::fmt::Write;
 use std::mem::size_of;
+use std::os::raw::{c_char, c_ulong};
 use std::slice;
 
 use anyhow::{bail, ensure, Result};
@@ -33,7 +34,7 @@ impl<'a> Btf<'a> {
         let bpf_obj = unsafe {
             libbpf_sys::bpf_object__open_mem(
                 object_file.as_ptr() as *const c_void,
-                object_file.len() as u64,
+                object_file.len() as c_ulong,
                 &obj_opts,
             )
         };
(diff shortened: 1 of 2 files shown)


willfindlay:
@davide125 Can you see if this fixes your issue?

willfindlay:
Perhaps we should run tests on other architectures via Github actions to prevent regressions?

danobi:
> Perhaps we should run tests on other architectures via Github actions to prevent regressions?

Yeah, totally, if there's a way. Lemme google around

danobi:
Don't think there's any native runners we can ues. We could use run-on-arch-action but b/c it's fully emulated it'll be super slow ( https://github.com/uraimo/run-on-arch-action/issues/4 ). Maybe we can just rely on bug reports from fedora lol

danobi (approved):

willfindlay:
> Don't think there's any native runners we can ues. We could use run-on-arch-action but b/c it's fully emulated it'll be super slow ( [uraimo/run-on-arch-action#4](https://github.com/uraimo/run-on-arch-action/issues/4) ). Maybe we can just rely on bug reports from fedora lol

That's too bad :(

davide125:
Testing a build with this here: https://koji.fedoraproject.org/koji/taskinfo?taskID=69203691

willfindlay:
> Testing a build with this here: https://koji.fedoraproject.org/koji/taskinfo?taskID=69203691

Looks like that did the trick :)

willfindlay:
@danobi Should be good to merge now

PR #110, opened 2021-06-03
Move .cargo/config to project root

willfindlay:
Running tests from the project root with `cargo test` causes libbpf-rs tests to fail since they are not being run with `sudo`. This PR moves libbpf-rs's `.cargo/config` into the project root to fix this. It has the nice added benefit of allowing examples to be run with `sudo` via a simple `cargo run`.


willfindlay:
Not sure why this is causing CI to fail, will investigate.

danobi:
The reason it's split like this is b/c I didn't wanna run the libbpf-cargo tests as root b/c it doesn't need root. If there's a way to only run libbpf-rs tests as root and still make `cargo test` work correctly from repository root that'd be great.

willfindlay:
> The reason it's split like this is b/c I didn't wanna run the libbpf-cargo tests as root b/c it doesn't need root. If there's a way to only run libbpf-rs tests as root and still make `cargo test` work correctly from repository root that'd be great.

I'll look into it. Closing this for now. I'll open a new PR if what you described is feasible.

PR #113, opened 2021-06-13
Lookup all values for per-cpu maps 

skyfmmf:
A successful lookup of a key in a per-cpu map does not return just one value but one value per CPU. These values written to the buffer aligned to 8 bytes. 

Currently, the `Vec` returned by `lookup` would just contain the first value. In addition, libbpf would write to memory areas outside the allocated buffer. This may cause crashes.

This PR makes `lookup` return a `Vec` with all values and proper alignment when called on per-cpu maps.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -143,14 +143,25 @@ impl Map {
         self.value_size
     }
 
-    fn all_values_size(&self) -> Result<u32> {
-        let ret = unsafe { libbpf_sys::libbpf_num_possible_cpus() };
-        if ret < 0 {
-            // Error code is returned negative, flip to positive to match errno
-            Err(Error::System(-ret))
+    /// Size of the value for lookups in bytes. Respects the space needed when looking up values in
+    /// per-cpu maps.
+    fn lookup_value_size(&self) -> Result<usize> {
+        let val_size = self.value_size() as usize;
+        if self.map_type() == MapType::PercpuArray
+            || self.map_type() == MapType::PercpuHash
+            || self.map_type() == MapType::LruPercpuHash
+            || self.map_type() == MapType::PercpuCgroupStorage
+        {
+            let ret = unsafe { libbpf_sys::libbpf_num_possible_cpus() };
+            if ret < 0 {
+                // Error code is returned negative, flip to positive to match errno
+                Err(Error::System(-ret))
+            } else {
+                let ncpu = ret as usize;
+                Ok(ncpu * util::roundup(val_size, 8))
+            }
         } else {
-            let ncpu = ret as u32;
-            Ok(ncpu * self.value_size())
+            Ok(val_size)
         }
     }
 
@@ -187,6 +198,9 @@ impl Map {
     /// Returns map value as `Vec` of `u8`.
     ///
     /// `key` must have exactly [`Map::key_size()`] elements.
+    ///
+    /// If the map is one of the per-cpu data structures, the returned `Vec` contains one value per
+    /// CPU in the system. The size of each value is rounded to the next multiple of 8 bytes.
     pub fn lookup(&self, key: &[u8], flags: MapFlags) -> Result<Option<Vec<u8>>> {
         if key.len() != self.key_size() as usize {
             return Err(Error::InvalidInput(format!(
@@ -196,12 +210,7 @@ impl Map {
             )));
         };
 
-        let out_size =
-            if self.map_type() == MapType::PercpuArray || self.map_type() == MapType::PercpuHash {
-                self.all_values_size()?
-            } else {
-                self.value_size()
-            } as usize;
+        let out_size = self.lookup_value_size()?;
         let mut out: Vec<u8> = Vec::with_capacity(out_size);
 
         let ret = unsafe {
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -143,6 +143,17 @@ impl Map {
         self.value_size
     }
 
+    fn all_values_size(&self) -> Result<u32> {
+        let ret = unsafe { libbpf_sys::libbpf_num_possible_cpus() };
+        if ret < 0 {
+            // Error code is returned negative, flip to positive to match errno
+            Err(Error::System(-ret))
+        } else {
+            let ncpu = ret as u32;
+            Ok(ncpu * self.value_size())
+        }
+    }
+
     /// [Pin](https://facebookmicrosites.github.io/bpf/blog/2018/08/31/object-lifetime.html#bpffs)
     /// this map to bpffs.
     pub fn pin<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
@@ -185,7 +196,13 @@ impl Map {
             )));
         };
 
-        let mut out: Vec<u8> = Vec::with_capacity(self.value_size() as usize);
+        let out_size =
+            if self.map_type() == MapType::PercpuArray || self.map_type() == MapType::PercpuHash {
+                self.all_values_size()?
+            } else {
+                self.value_size()
+            } as usize;
+        let mut out: Vec<u8> = Vec::with_capacity(out_size);
 
         let ret = unsafe {
             libbpf_sys::bpf_map_lookup_elem_flags(
@@ -198,7 +215,7 @@ impl Map {
 
         if ret == 0 {
             unsafe {
-                out.set_len(self.value_size() as usize);
+                out.set_len(out_size);
             }
             Ok(Some(out))
         } else {
--- a/libbpf-rs/src/util.rs
+++ b/libbpf-rs/src/util.rs
@@ -27,3 +27,8 @@ pub fn c_ptr_to_string(p: *const c_char) -> Result<String> {
         .map_err(|e| Error::Internal(e.to_string()))?
         .to_owned())
 }
+
+/// Round up a number to the next multiple of `r`
+pub fn roundup(num: usize, r: usize) -> usize {
+    ((num + (r - 1)) / r) * r
+}


libbpf-rs/src/map.rs
```diff
@@ -143,6 +143,28 @@ impl Map {
         self.value_size
     }
 
+    /// Size of the value for lookups in bytes. Respects the space needed when looking up values in
+    /// per-cpu maps.
+    fn lookup_value_size(&self) -> Result<usize> {
```
danobi, line 148: Can you add a comment above `MapType` to remind editors to update this function too?

libbpf-rs/src/util.rs
```diff
@@ -27,3 +27,8 @@ pub fn c_ptr_to_string(p: *const c_char) -> Result<String> {
         .map_err(|e| Error::Internal(e.to_string()))?
         .to_owned())
 }
+
+/// Round up a number to the next multiple of `r`
+pub fn roundup(num: usize, r: usize) -> usize {
+    ((num + (r - 1)) / r) * r
+}
```
danobi, line 34: Can you add some tests for this? Just some simple ones to be safe

danobi (changes requested):
LGTM overall.

Can you add a test for per-cpu maps too? See libbpf-rs/tests/ for existing examples

libbpf-rs/src/map.rs
skyfmmf: Done. I added the comment as a normal, non-doc comment so this implementation note does not show up in the user documentation.

skyfmmf:
While writing the requested tests, I noticed, that the update on per-cpu maps also needs the bigger buffer, so I made the modifications there too and made the `value_buffer_size` part of the public API.

anakryiko:
Wouldn't it be cleaner to have lookup_percpu() that would return a vector of vectors? That way one can actually sanely iterate each CPU's value. Internally we can error out in lookup() if it's used on per-CPU maps, and vice versa, lookup_percpu() would only work for per-CPU maps and would error out for everything else. I think that's much saner interface than concatenated blob of values for all CPUs? WDYT?

willfindlay:
> Wouldn't it be cleaner to have lookup_percpu() that would return a vector of vectors? That way one can actually sanely iterate each CPU's value. Internally we can error out in lookup() if it's used on per-CPU maps, and vice versa, lookup_percpu() would only work for per-CPU maps and would error out for everything else. I think that's much saner interface than concatenated blob of values for all CPUs? WDYT?

I agree with this. This way the burden is not on the user to manually fiddle with coercing the byte slice into the correct collection. Erroring out for non-percpu maps and vice versa sounds like an idiomatic approach.

skyfmmf:
I agree as well. It would be the way cleaner interface to return a vector of vectors and hide the 8 byte alignment from the users. Then we should probably also have update_percpu, which also works on a vector of the values (maybe like `Vec<&[u8]>`?). I guess, the additional copying of the data to/from the bytes the kernel expects/returns is acceptable.

Actually, I think it might be nicer to reflect this difference in the behavior also in the type system, i.e., have something like a PercpuMap. But that comes with a lot of additional work.

danobi:
Agreed that lookup_percpu()-style API is better.

> Then we should probably also have update_percpu, which also works on a vector of the values (maybe like Vec<&[u8]>

I think the kernel automatically sets the value on each cpu:
```
bpf()
  map_update_elem()
    bpf_percpu_hash_update()  // similar to other percpu map types
      __htab_percpu_map_update_elem(.., /* onallcpus=*/ true)
        pcpu_copy_value(.., /* oncallcpus=*/ true)
          for_each_possible_cpu()
            bpf_long_memcpy()
        
```

> I guess, the additional copying of the data to/from the bytes the kernel expects/returns is acceptable.

Yeah, there's an extra copy but it's probably worth it for the ergonomics.

> Actually, I think it might be nicer to reflect this difference in the behavior also in the type system, i.e., have something like a PercpuMap. But that comes with a lot of additional work.

Yes, it'd be extra work. We'd also need to consider the complexity it adds to the API. For example, map definitions are stored in the bpf object file. And we cannot return different types based on runtime state. So we'd need a separate `Object::map()`, `Object::map_mut()`, `Object::maps_iter()`, etc. for each map type. You _could_ return a trait object or a `impl T` but now we're back to runtime checks again.

skyfmmf:
Alright, I'll update the PR to the lookup_percpu() API then. 

> > Then we should probably also have update_percpu, which also works on a vector of the values (maybe like Vec<&[u8]>
> 
> I think the kernel automatically sets the value on each cpu:
> 
> ```
> bpf()
>   map_update_elem()
>     bpf_percpu_hash_update()  // similar to other percpu map types
>       __htab_percpu_map_update_elem(.., /* onallcpus=*/ true)
>         pcpu_copy_value(.., /* oncallcpus=*/ true)
>           for_each_possible_cpu()
>             bpf_long_memcpy()
>         
> ```

Oh, that makes the update API more complicated. The syscall actually allocates the 8 byte aligned array of values and fills it with data from user space (https://elixir.bootlin.com/linux/v5.12.11/source/kernel/bpf/syscall.c#L1116). And while an update syscall for the hash map uses just the first value and sets it on all CPUs, the update syscalls for arrays and cgroup-storage use all of the values (https://elixir.bootlin.com/linux/v5.12.11/source/kernel/bpf/arraymap.c#L349, https://elixir.bootlin.com/linux/v5.12.11/source/kernel/bpf/local_storage.c#L230). 

This means, that an update_percpu() with as many values as cores would make sense for per-cpu arrays and per-cpu cgroup storage, but would do unexpected things on per-cpu (lru) hash maps (updating everything with the first value). Should we have update_percpu() for arrays and cgroup storage and continue to update() for per-cpu hashes and error out in the updates accordingly, despite that being potentially confusing to users? And of course document that properly...

anakryiko:
> And while an update syscall for the hash map uses just the first value and sets it on all CPUs

Have you tested that? From looking at https://github.com/torvalds/linux/blob/master/kernel/bpf/hashtab.c#L834-L841 what you are saying isn't true.

skyfmmf:
> Have you tested that? From looking at https://github.com/torvalds/linux/blob/master/kernel/bpf/hashtab.c#L834-L841 what you are saying isn't true.

No, I was just assuming it from @danobi's comment. But yes, it seems to use one value per CPU. I'll test that for the implementation.

Then we are back at also needing the update_percpu() API. Should that take a `&Vec<&[u8]>` so it somehow matches the existing API of update?

danobi:
Oh I think I was reading the `pcpu_copy_value()` code wrong

danobi:
Ok, so how about this?

* add a `lookup_percpu()` API
  * instead of returning `Vec<Vec<u8>>`, we could return a custom struct that holds the aggregate value in one `Vec` but provides an iterator interface for each CPU's values. Don't have to do this but it's another option to save on some allocations
* add a `update_percpu()` API
* also add `lookup_percpu_and_delete()` for completeness? Not sure about this

anakryiko:
custom type that represents per-cpu array of values might be a nice API. You'd remember a blob of memory and number of cpus internally, and let either iterate over all CPUs or just fetch &[u8] references to a specified CPU, right? Add some casting/reinterpretation (probably as unsafe APIs), and you'll avoid any extra allocation for a bunch of cases.

The question is what do we do for update? Accept the same custom type or Vec<Vec<u8>>? Probably consistent use would be good, but we'll need to provide convenient constructors from different representations.

I'm not familiar with _and_delete() operations, but if they are supported in kernel, then it's probably good to have API for them.

danobi:
> The question is what do we do for update? Accept the same custom type or Vec<Vec>? Probably consistent use would be good, but we'll need to provide convenient constructors from different representations.

We could probably use that for updates too. Maybe the following constructors (didn't think too hard yet):
```
struct PercpuValue { ... }

impl PercpuValue {
    pub fn new(ncpus: usize) -> Self
    pub fn new_with_val(ncpus: usize, val: &[u8]) -> Self
    pub fn new_with_vals(vals: &[&[u8]]) -> Self
}

**Comment by anakryiko** [comment]
For number of cpus. I'd determine it internally. It's number of possible CPUs as configured for a given kernel. So it doesn't change, and it's not up to user application. So there is no point in taking it from users. It's probably better to use that internal value to enforce the correct sizing of per-cpu vectors.

**Comment by danobi** [comment]
> For number of cpus. I'd determine it internally. It's number of possible CPUs as configured for a given kernel. So it doesn't change, and it's not up to user application. 

Ah yeah good point. So then does it even make sense to add a per-cpu setting API? User app cannot assume any number other than 1. So even if they wanted to set per-cpu values they'd need a giant if/else block for # cpus. The only realistic case I can think of is same value for all CPUs. So maybe we could just have the current `Map::update(..)` set the same value for all CPUs?

**Comment by anakryiko** [comment]
I don't understand. For per-cpu update user needs to specify an array of values, where each element corresponds to distinct CPU. 

> User app cannot assume any number other than 1.

This parts makes little sense to me... I'm not sure what problem we are discussing anymore :)

**Comment by skyfmmf** [comment]
I think that the app needs to be able to set different values per cpu, i.e., know about the number of CPUs. libbpf has libbpf_num_possible_cpus() for that and I would just expose this function in a safe way from libbpf-rs. Of course, if we go with something like `PercpuValue`, this should check the length where it is necessary to get it from the user.

As for the lookup_and_delete(): I think right now it is not necessary to implement that for percpu as well because it is only supported for a few map types and none of them is per-cpu (https://elixir.bootlin.com/linux/v5.12.12/source/kernel/bpf/syscall.c#L1510).

I get the feeling that the `PercpuValue` is a bit too much just to save on a few allocations for map types that were apparently not commonly accessed in the past. Therefore, I suggest to keep it simple and potentially add the iterator API later in case there is an issue with the allocations. 

I suggest an API something like this: 
```rust
fn lookup_percpu(key: &[u8]) -> Vec<Vec<u8>>;
fn update_percpu(key: &[u8], values: &Vec<Vec<u8>>);
```
Then users can do whatever they need to do with the returned, owned data. They can also modify the returned Vec<Vec> and use it again to update the map (this lookup-update-cycle is why I suggest `&Vec<Vec<u8>>` for the values instead of `&Vec<&Vec<u8>>` or similar things).

danobi:
Yeah, that makes sense to me.

skyfmmf:
I have updated the PR to match the API discussion we had a while ago. I ended up not putting the number of cpus into the `MapType` enum (as @AlexZhao's patch suggested) to keep the enum matching the kernel definition more closely. 

Please have a look and suggest improvements where you see fit.

jfernandez:
@skyfmmf I've been using your branch for a project and it's been working great so far.

libbpf-rs/src/map.rs
```diff
@@ -133,6 +133,17 @@ impl Map {
...
+    /// Returns if the map is of one of the per-cpu types.
+    pub fn is_percpu(&self) -> bool {
+        match self.map_type() {
+            MapType::PercpuArray
+            | MapType::PercpuHash
+            | MapType::LruPercpuHash
+            | MapType::PercpuCgroupStorage => true,
+            _ => false,
+        }
+    }
```
danobi, line 145: I wonder if:
* this should be a method on `MapType` 
* we should add a `map_type()` method on `OpenMap` as well

so the OpenMap and Map are symmetric.

Thoughts?

libbpf-rs/src/map.rs
```diff
@@ -295,6 +364,72 @@ impl Map {
...
+                "update() must be used for maps that are not per-cpu (type of the map is {})",
+                self.map_type(),
+            )));
+        }
+
+        if values.len() != num_possible_cpus()? {
+            return Err(Error::InvalidInput(format!(
+                "number of values {} != number of cpus {}",
+                values.len(),
+                num_possible_cpus()?
```
danobi, line 394: Cache value of `num_possible_cpus()`?

danobi (commented):
Sorry about the slow review (conference season).

Mostly looks good to me. One other thing: should we add a percpu version of `OpenMap::set_initial_value()` as well?

libbpf-rs/src/map.rs
skyfmmf: I agree that it looks a bit expensive to call `num_possible_cpus()` everytime. But I just took a look at the implementation of `libbpf_num_possible_cpus()` in libbpf itself and it already caches the result (https://github.com/libbpf/libbpf/blob/e671a47bc279c6a73d1fee5aa522b0979fd1a33a/src/libbpf.c#L10905). So I think another layer of caching is not necessary.

libbpf-rs/src/map.rs
skyfmmf: Making this a method on MapType sounds like a good idea. I'll move it over there. 

I am not so sure about implementing `map_type()` on `OpenMap`. It seems like we don't have a lot of information available at that stage and it feels a bit out of scope for this PR. But if `is_percpu()` is on `MapType`, it is immediately usable on `OpenMap` if someone implements `map_type()` there.

skyfmmf:
> One other thing: should we add a percpu version of `OpenMap::set_initial_value()` as well?

Without looking much into it, I don't think there is a per-cpu version of `set_initial_value()`. Also, I can not really think of a reason why one would like different initial values on different cores.

PR #114, opened 2021-06-15
libbpf-cargo: Switch memmap dependency to memmap2

willfindlay:
According to [this security advisory](https://rustsec.org/advisories/RUSTSEC-2020-0077), the `memmap` crate is no longer maintained and should be replaced with the `memmap2` fork. This causes `cargo audit` to fail on any project that depends on `libbpf-cargo`. This PR simply replaces `memmap` with `memmap2`.

As an aside, I would like to propose adding `cargo audit` and `cargo clippy` to the CI workflow to reduce the risk of vulnerable dependencies and enforce good code hygiene. Any thoughts? I can file a separate PR for this if you're interested.

### DIFF
--- a/libbpf-cargo/Cargo.toml
+++ b/libbpf-cargo/Cargo.toml
@@ -27,7 +27,6 @@ path = "src/lib.rs"
 anyhow = "1.0"
 cargo_metadata = "0.12"
 libbpf-sys = { version = "0.4.0-1" }
-memmap = "0.7"
 num_enum = "0.5"
 regex = "1.4"
 scroll = "0.10"
@@ -38,6 +37,7 @@ structopt = "0.3"
 semver = "0.9"
 tempfile = "3.1"
 thiserror = "1.0"
+memmap2 = "0.3.0"
 
 [dev-dependencies]
 goblin = "0.2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,7 +188,7 @@ dependencies = [
  "cargo_metadata",
  "goblin",
  "libbpf-sys",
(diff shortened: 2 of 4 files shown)


danobi:
> As an aside, I would like to propose adding cargo audit and cargo clippy to the CI workflow to reduce the risk of vulnerable dependencies and enforce good code hygiene. Any thoughts? I can file a separate PR for this if you're interested.

Sure, sounds good. Ideally both audit and clippy are warnings and not hard errors. I think I tried for another project to have GH actions post a message to the PR but I don't believe it's possible (or at least it wasn't at the time). You might need to use a github app (or whatever it's called).

If there's stuff we can pull off the shelf that'd be ideal but I wouldn't go crazy over it if there isn't a preexisting thing.

libbpf-cargo/Cargo.toml
```diff
@@ -38,6 +38,7 @@ structopt = "0.3"
 semver = "0.9"
 tempfile = "3.1"
 thiserror = "1.0"
+memmap2 = "0.3.0"
```
danobi, line 41: Shouldn't the old `memmap = 0.7.0` dep be removed too?

danobi (changes requested):

libbpf-cargo/Cargo.toml
willfindlay: > Shouldn't the old `memmap = 0.7.0` dep be removed too?

Oops, I think this got added back when I was rebasing. I'll remove it again!

danobi (approved):
Thanks!

PR #115, opened 2021-06-16
libbpf-cargo: Show better error message if clang isn't found

danobi:


libbpf-cargo/src/build.rs
```diff
@@ -42,7 +42,10 @@ fn extract_version(output: &str) -> Result<&str> {
 }
 
 fn check_clang(debug: bool, clang: &Path, skip_version_checks: bool) -> Result<()> {
-    let output = Command::new(clang.as_os_str()).arg("--version").output()?;
+    let output = Command::new(clang.as_os_str())
+        .arg("--version")
+        .output()
+        .context("Failed to execute clang")?;
```
anakryiko, line 48: if this fails, it's more of a "failed to find clang"?

anakryiko (approved):

libbpf-cargo/src/build.rs
willfindlay: `.context` would bubble up the No such file or directory error

danobi: Yeah, what @willfindlay  said

PR #116, opened 2021-06-17
Bump strum_macros to 0.21

eclipseo:
Hi,

I'd like to update strum_macros to 0.21, as we strive to keep package in sync to the latest version during Fedora Rust packaging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -616,9 +616,9 @@ dependencies = [
 
 [[package]]
 name = "strum_macros"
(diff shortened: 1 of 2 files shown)


danobi (changes requested):
I think Cargo.lock needs to be updated too (just run `cargo build` in repo root).

BTW, would be good to have a justification for update. Even if it's just "newer version is out".

eclipseo:
Thanks, it is done.

PR #118, opened 2021-07-09
Use vendored headers when building progs 

danobi:
Currently we expect the progs to be built with system provided libbpf
headers. This is problematic b/c they may not match the libbpf (C
library) that libbpf-rs uses.

https://github.com/alexforster/libbpf-sys/commit/f966e1e620b exposed 
vendored headers libbpf-sys users. This PR has libbpf-cargo build
use the vendored headers.


libbpf-cargo/src/build.rs
```diff
@@ -128,6 +147,9 @@ fn compile_one(debug: bool, source: &Path, out: &Path, clang: &Path, options: &s
 }
 
 fn compile(debug: bool, objs: &[UnprocessedObj], clang: &Path) -> Result<()> {
+    let header_dir = extract_libbpf_headers_to_disk()?;
```
mimullin-bbry, line 150: Is this going to force users to use the libbpf-sys provided bpf headers?

mimullin-bbry:
Change LGTM, would just like to know if this will force users to use libbpf-sys' headers.

libbpf-cargo/src/build.rs
danobi: > libbpf-sys provided bpf headers

To clarify, this PR only affects libbpf API headers. Kernel bpf headers remain unaffected.

By default, it will have includes like `#include <bpf/bpf_helpers.h>` resolve to libbpf-sys provided helpers. But because we the libbpf-sys `-I` at the end of the clang args, the user can provide their own search path with `-I` and it should get resolved before libbpf-sys's path.

But I can't think of any case where the user would want to use their own libbpf headers. Kernel bpf headers, sure. But libbpf headers, no.

mimullin-bbry: +1

anakryiko (approved):
LGTM.

Issue #119, opened 2021-07-11
Rust Cannot Derive Debug for unions


mimullin-bbry:
```
#[derive(Debug)]
pub union test {
    pid: u8,
    b: u64,
}

fn main() {
    println!("Hello, world!");
}
```

The above code will not compile in rust.  As such in libbpf-cargo https://github.com/libbpf/libbpf-rs/blob/master/libbpf-cargo/src/btf/btf.rs#L469
this code which puts a #[derive(Debug...] on a union will create invalid code.

mimullin-bbry:
This issue is pernicious.

If a union cannot be Debug, then a struct which has said union as a field cannot be debug.  We can test for this when building the struct.

However, if a union is a field of a struct, and said struct is a field of another struct, that "second level" struct also cannot be Debug.  The libbpf-cargo code generation doesn't look multiple levels deep.

Choice is the remove #[derive(Debug)] from all code generation, or to impl Debug for unions.  Note I think we need to impl Default for all unions too (investigating this before creating an issue).

PR #120, opened 2021-07-11
Add a Debug impl for all union types

mimullin-bbry:
Fix for 
https://github.com/libbpf/libbpf-rs/issues/119

Rust cannot #[derive(Debug)] for unions.  Structures must have Debuggable unions in order to #[derive(Debug)] for them.

To solve issue #119 we must either remove all #[derive(Debug)] or impl Debug for unions.  This PR suggests doing the latter.

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -488,7 +488,6 @@ impl<'a> Btf<'a> {
                     writeln!(def, "}}")?;
 
                     // if required write a Default implementation for this struct
-                    #[rustfmt::skip]
                     if gen_impl_default {
                         writeln!(def, r#"impl Default for {} {{"#, t.name)?;
                         writeln!(def, r#"    fn default() -> Self {{"#)?;
@@ -502,7 +501,10 @@ impl<'a> Btf<'a> {
                     } else if !t.is_struct {
                         // write a Debug implementation for a union
                         writeln!(def, r#"impl std::fmt::Debug for {} {{"#, t.name)?;
-                        writeln!(def, r#"    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {{"#)?;
+                        writeln!(
+                            def,
+                            r#"    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {{"#
+                        )?;
                         writeln!(def, r#"        write!(f, "unimplemented debug for union")"#)?;
                         writeln!(def, r#"    }}"#)?;
                         writeln!(def, r#"}}"#)?;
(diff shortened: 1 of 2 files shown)


mimullin-bbry:
Note: I was a bit premature on my submission as the test cases need to be updated.

danobi:
Hmm, maybe it's better to remove `Debug` impl instead of adding an unimplemented stub for union. None of the code I'm aware of is taking advantage of `Debug` trait. Thoughts?

mimullin-bbry:
> Hmm, maybe it's better to remove Debug impl instead of adding an unimplemented stub for union.

It's a coin flip to me.  I rarely use the Debug trait in my code as I find it's output difficult to read.

I'll argue in favour of the current change, but I'm certainly not married to the idea of retaining Debug traits throughout the codebase.

So three arguments in favour of retaining the Debug trait.

1) the Debug trait already exists in the code; it's there if anyone wants to use it, they can w/o any fuss.  Removing would mean that users would have to implement the Debug trait by hand if they ever wanted it, which can get really tedious for nested types.

If there was a way to #[derive(X)] on a previously created aggregate/enum then I think it would be a simple "remove it" answer, as users could easily put it back in if they wanted it, and whether the user wants to use this is really on the user themselves, and shouldn't really be on the library.  But things as they are with rust, that's not possible AFAIK.

2) IMHO the code as written does not require a versioning bump, while the removal does.

3) The code overhead that impl Debug for union brings is rather minor considering my other PR which uses the same if statement to impl Default for unions.  IE it's not really a code cleanliness issue as IMHO we must impl Default for unions so this code only adds 5 writeln! overhead of that code.

danobi:
Sure, I buy that argument. Let's go w/ the current implementation, then.

libbpf-cargo/src/btf/btf.rs
```diff
@@ -496,6 +498,16 @@ impl<'a> Btf<'a> {
...
                         writeln!(def, r#"    }}"#)?;
                         writeln!(def, r#"}}"#)?;
+                    } else if !t.is_struct {
+                        // write a Debug implementation for a union
+                        writeln!(def, r#"impl std::fmt::Debug for {} {{"#, t.name)?;
+                        writeln!(
+                            def,
+                            r#"    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {{"#
+                        )?;
+                        writeln!(def, r#"        write!(f, "unimplemented debug for union")"#)?;
```
danobi, line 508: Maybe this, to keep it more terse? 
```suggestion
                        writeln!(def, r#"        write!(f, "(???)"#)?;
```

mimullin-bbry: +1

danobi (approved):

PR #126, opened 2021-08-05
impl Default for unions

mimullin-bbry:
impl Default for unions so that they can be placed inside of other aggregates

re-implement the test_btf_dump_definition_anon_enum as it was
accidentally overwritten in commit:
3e083dd


danobi (approved):
LGTM, thanks!

PR #130, opened 2021-08-16
libbpf-cargo: btf: Support bool type

### LINKED ISSUE #129: bool in bpf code becomes u8 in generated skeleton code
Hi,

The current skel generation does this. I was looking up how to convert the bool properly to u8 when I found [this](https://stackoverflow.com/a/47705543/2277013) . Any thoughts on why the skel generation represents bool in C as u8 in Rust? Also, how would you convert rust bool to rust u8 so that the bool value does get passed to bpf code?

Thanks in advance

danobi:
Before, forgot to check int encoding for bools. Rust bools appear to be
1 byte wide just like C bool.

This fixes #129.

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -197,10 +197,13 @@ impl<'a> Btf<'a> {
                     _ => bail!("Invalid integer width"),
                 };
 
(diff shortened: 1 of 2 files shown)


anakryiko (approved):
lgtm

PR #131, opened 2021-08-18
Capable using libbpf 

DevasiaThomas:
Hi Guys,
I'm new to a few things - Rust and ebpfs. @anakryiko's your blog posts a good starting point resource when approaching ebpfs from a high level language. I had a few issues with my dev env that @danobi helped with- ty for that. I took the suggestion in the blogposts and partially converted [capable(8)](https://github.com/iovisor/bcc/blob/master/tools/capable.py) to contribute as an example here and possibly contribute more to this repo, if I can. I say partially because:

* [filter_by_containers()](https://github.com/iovisor/bcc/blob/0120770f3cdcd0540861e2f92c4f796fe09c3ff6/src/python/bcc/containers.py#L73) is a cool helper function in bcc. I thought of implementing it myself, coz it seems to be doable. But I thought, it would be better if it was implemented in someway through ``bpf_helpers.h``. I also noticed all libbpf converted tools that had this function used in their bcc versions just decided to ignore it and not implement the ``--cgroupmap`` and 
``--mntnsmap`` options
* [These](https://github.com/iovisor/bcc/blob/0120770f3cdcd0540861e2f92c4f796fe09c3ff6/tools/capable.py#L261) 4 lines of code hide a whole lot of code that is provided by BCC to do stacktrace stuff. I noticed that libbpf-tools that dealt with stack traces used @ethercflow's(your blog posts helped as well) code in ``trace_helpers``. I was wondering why it's not part of libbpf and by extension libbpf-rs, since a lot of the converted tools do use that and bcc provides that. Hence the stacktrace options are a no-op right now
* The blogposts talk about considering using ``raw_tp/fentry/fexit`` instead of ``tracepoint/kprobe/kretprobe``.  Bcc had a few helper functions around detecting the presence of these features. I thought I could use ``LINUX_KERNEL_VERSION >= KERNEL_VERSION(x, y, z)`` to get around this. But then I saw how a few libbpf-tools did it by using ``fentry_exists(name, mod)`` which made me think its not that simple. However, that function is also part of ``trace_helpers`` - not libbpf.  After detection of those features, libbpf lets you disable autoload. But, I don't think thats possible in libbf-rs right now. The disable autoload exists in ``libbpf_sys`` and is exposed through ``libbpf_rs`` - but the ``bpf_program`` is a private member of the generated code and hence can't be passed over. So for now I just have the ``kprobe`` implemented. Till this detection and autoload disabling  is possible, the ``fentry`` version probably stays commented.

Apart from the request to absorb ``trace_helpers``, I thought it would be cool to actually distribute the latest ``vmlinux.h`` within libbpf and libbpf-rs/cargo(I have a feeling rustaceans might oppose this) to be used in development because I just noticed that I have the same ``vmlinux.h`` as the ``runqslower`` example. 

Also, if you noticed bcc capable  has a lot of ``#ifdefs`` in structs that are controlled by user args, probably to reduce struct size if those options aren't used. I know thats not possible as-is with libbpf, any suggestions on how to achieve this in a smart manner? 

All comments, thoughts and code reviews are highly appreciated :)


DevasiaThomas:
Oh another thing, I think the ``runqslower`` example forgot to swap ``pid`` and ``tgid`` when setting the configuration for bpf code. It had me confused for a while, coz I thought libbpf-rs intelligently handled that somehow

danobi:
Thanks for the PR! Unfortunately I'll be out the rest of the week but I'll take a look the next week.

In the meantime, @anakryiko , should this tool go into libbpf-bootstrap?

anakryiko:
> In the meantime, @anakryiko , should this tool go into libbpf-bootstrap?

@DevasiaThomas this seems to be a Rust implementation of BCC's capable.py, is that right? How complete is it and how compatible with BCC's version (in terms of CLI arguments, output format, features supported, etc) is it?

If it's a complete tool, I think it might make sense to start a discussion over in BCC repo about adding Rust-based tool implementations. We'd need to figure out the best setup, of course, but I think the discussion should start in https://github.com/iovisor/bcc/.

DevasiaThomas:
> @DevasiaThomas this seems to be a Rust implementation of BCC's capable.py, is that right? How complete is it and how compatible with BCC's version (in terms of CLI arguments, output format, features supported, etc) is it?

@anakryiko Yes it is. **Completeness**: No stack traces because of the the symbolization and ``trace_helpers`` discussion above. The ``--cgroupmap`` and ``--mntnsmap`` that tie into BCC's should filter container substitutions aren't there either - I noticed none of the libbpf conversions done in C implement those either. So, I left it unimplemented here as well. I don't think its hard to implement it, but its a handy helper function across multiple tools. As for the ``--unique`` option, I didn't quite understand the how it differentiated between cgroups and pids(I don't think that part is implemented or works). So I extended this option to explicitly ask for pid or cgroup uniqueness.
 
> If it's a complete tool, I think it might make sense to start a discussion over in BCC repo about adding Rust-based tool implementations. We'd need to figure out the best setup, of course, but I think the discussion should start in https://github.com/iovisor/bcc/.

Based on what I said above, compared to other ``libbpf-tools`` conversions,  I lack stack traces. So, not complete with respect to the C conversions but as complete as I could get with Rust.

DevasiaThomas:
> Couple things:
> 
> * could you please remove all the commented out code?
> * please `cargo fmt` to format the rust code

@danobi Done. I had rustfmt working automatically when i saved the files. Idk why cargo fmt and rustfmt have different outputs. I'll look it up later.

examples/capable/src/bpf/capable.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+#include <bpf/bpf_core_read.h>
+
+extern int LINUX_KERNEL_VERSION __kconfig;
+
+const volatile struct {
+    gid_t tgid; //PID to filter
+    bool verbose; // Include non audit logs
+    enum Unique unique_type; // Only unique info traces for same pid or cgroup
+
+} tool_config = {0};
```
danobi, line 18: Nit
```suggestion
} tool_config = {};
```

examples/capable/src/bpf/capable.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+
+
+struct event _event = {0}; //Dummy instance for skeleton to generate definition
```
danobi, line 21:
```suggestion
struct event _event = {}; //Dummy instance for skeleton to generate definition
```

examples/capable/src/bpf/capable.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    if (!tool_config.verbose && audit == 0) {
+        return 0;
+    }
+
+    uid_t uid = bpf_get_current_uid_gid();
+
+    struct event event = {.tgid = tgid, .pid = pid, .uid = uid, .cap = cap, .audit = audit,
+            .insetid = insetid};
```
danobi, line 72:
```suggestion
    struct event event = {
            .tgid = tgid,
            .pid = pid,
            .uid = uid,
            .cap = cap,
            .audit = audit,
            .insetid = insetid
    };
```

examples/capable/src/bpf/capable.bpf.c
```diff
@@ -0,0 +1,103 @@
...
+
+    if (tool_config.unique_type) {
+        struct unique_key key = {0};
+        key.cap = cap;
```
danobi, line 76: Can we do this in one line? Or does verifier complain?
```suggestion
        struct unique_key key = {.cap = cap};
```

examples/capable/src/bpf/capable.h
```diff
@@ -0,0 +1,23 @@
...
+    uid_t uid;
+    int cap;
+    int audit;
+    int insetid;
+    u8 comm[TASK_COMM_LEN];
+    int kernel_stack_id;
+    int user_stack_id;
+};
+
+enum Unique {UnqOff, UnqPid, UnqCgroup};
```
danobi, line 21: Nit: kernel naming style
```suggestion
enum uniqueness {UNQ_OFF, UNQ_PID, UNQ_CGROUP};
```

examples/capable/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+    verbose: bool,
+    /// only trace <pid>
+    #[structopt(short, long, default_value = "0")]
+    pid: u32,
+    /// extra fields: Show TID and INSETID columns
+    #[structopt(short = "x", long = "extra")]
+    extra_fields: bool,
+    /// don't repeat same info for the same pid<1> or cgroup<2>
+    #[structopt(long = "unique", default_value = "0")]
+    unique_type: Unique,
```
danobi, line 93: Is the bcc tool using integers as part of cmd line options? If not, would be better to do something more easily human readable

danobi (commented):
Mostly lgtm, just some nits.

Btw, the prog code should follow linux kernel coding style. I pointed out some small stuff I noticed, but the entire file should be using tabs.

examples/capable/src/main.rs
DevasiaThomas: I don't think the --unique worked as advertised in the bcc tool. Thats why I did this. I can definitely change it to parse strings. no probs.

examples/capable/src/bpf/capable.h
DevasiaThomas: @danobi I had it like this first, but when I ``cargo build``, I got warnings for [enum naming conventions](https://doc.rust-lang.org/rust-by-example/custom_types/enum.html). When I thought this was to be a rust project, I thought it might be easier to have it coded rust style in the c code as well so that no compile time warnings happened. I wasn't sure if code-gen should be changed to emit rust convention code or do [this](https://stackoverflow.com/a/64363119/2277013) instead. I'm totally fine changing it back and ignoring that compile time warning. Lmk.

examples/capable/src/bpf/capable.bpf.c
DevasiaThomas: nope no issues there.

examples/capable/src/bpf/capable.h
danobi: Oh, maybe we can add more stuff to https://github.com/libbpf/libbpf-rs/blob/3ff6c9377e4d98c4af700347a99e303e9abff319/libbpf-cargo/src/gen.rs#L666-L670 . Having C code look like C ought to take precedence here. B/c the rust code knows it's using C types but we cannot say the same about the other way around.

DevasiaThomas:
@danobi a side question, does the callback for lost events exist because the rust code was not able to consume events fast enough? or is it lost way before?

danobi:
> @danobi a side question, does the callback for lost events exist because the rust code was not able to consume events fast enough? or is it lost way before?

I believe the lost even callback exists if the prog tries to add data to the ring buffer but the buffer is full. This can occur if userspace isn't emptying the ring buffer fast enough.

danobi:
LGTM, I'll merge if @anakryiko doesn't any any comments in a few days

anakryiko:
Sorry, I haven't payed too much attention. I don't mind, let's land it here.

We should discuss setting up Rust-based infra for libbpf-tools separately. The more complicated the tooling, the biggest advantage of Rust there is over C. But this is a separate discussion from this, of course.

PR #134, opened 2021-10-14
Bump libbpf-sys to 0.5.0-1 plus a few more dependencies

davide125:
In preparation for a new below release


davide125:
Thanks @danobi! Could you add myself, @brianc118 and @dschatzberg please? Both for libbpf-{rs,cargo} and for the below crates.

PR #137, opened 2021-10-29
Add set_print, to send libbpf output to a callback

benfogle:
I want to be able to send output from libbpf-rs to the `log` crate and filter it by level, along with all the other chatter my application produces. Right now the best I can do is send it to stderr. This PR adds `set_print`, which allows a user-defined callback to receive libbpf output.


mimullin-bbry:
I was just thinking that something like this was needed. :)

libbpf-rs/src/print.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+/// Set a callback to receive log messages from libbpf, instead of printing them to stderr
+///
+/// This overrides (and is overridden by) [`ObjectBuilder::debug`]
+pub fn set_print(func: PrintCallback) {
+    let cb = unsafe { std::mem::transmute::<PrintCallback, *mut ()>(func) };
+    PRINT_CB.store(cb, Ordering::Relaxed);
+    unsafe { libbpf_sys::libbpf_set_print(Some(outer_print_cb)) };
```
anakryiko, line 59: libbpf_set_print returns the previous instance of print callback. It's quite important and useful when someone wants to temporarily replace print callback and restore the previous one. Can the same pattern be support in Rust?

benfogle: I can rework this to include that. I was a little unsure about how to handle return values. The return from `libbpf_set_print` would be a nullable, unsafe function pointer, so at best it could be a token to restore things later. Since in my implementation, the real callback is a static function, the `libbpf_set_print`'s return value alone wouldn't be sufficient to restore the print callback in a way the user expects.

Would something like a resource guard be acceptable here, or do you think there is a real need to expose the return value?

```
{
    let _lock = libbpf_rs::set_print_temporary(callback);
    ...
}
// Original callback restored here
```

anakryiko: Resource guard seems like an overkill. We can make set_print() not return anything in libbpf-rs. Alternatively you can return PrintCallback and just keep track of current callback on purely Rust side. It might be a bit fragile if someone is calling into libbpf_set_print() through FFI, but that should be extremely rare (an we can say unsupported) case.

benfogle: All right, that should be updated.

libbpf-rs/src/print.rs
```diff
@@ -24,8 +25,29 @@ impl From<libbpf_sys::libbpf_print_level> for PrintLevel {
...
+    ($x:expr) => {
+        unsafe { std::mem::transmute::<PrintCallback, *mut ()>($x) }
+    };
+}
+
+fn from_ptr(ptr: *mut ()) -> PrintCallback {
+    unsafe { std::mem::transmute::<*mut (), PrintCallback>(ptr) }
+}
+
 // There is no AtomicFnPtr. This is a workaround.
```
insearchoflosttime, line 42: Do you think using the [atomic crate](https://github.com/Amanieu/atomic-rs) for the AtomicFnPtr works as a viable alternative as well? It seems like other rust users are using the atomic crate for atomic function ptrs per the conversation in https://github.com/rust-lang/rfcs/issues/2481.

libbpf-rs/src/print.rs
```diff
@@ -34,27 +56,33 @@ extern "C" fn outer_print_cb(
     fmtstr: *const c_char,
     va_list: *mut libbpf_sys::__va_list_tag,
 ) -> i32 {
-    // can't be null: we cant't get here until print_cb has been initialized with a function
-    // pointer.
-    let cb = PRINT_CB.load(Ordering::Relaxed);
-    let cb = unsafe { std::mem::transmute::<*const (), PrintCallback>(cb) };
+    let cb = from_ptr(PRINT_CB.load(Ordering::Relaxed));
     match unsafe { vsprintf::vsprintf(fmtstr, va_list) } {
```
insearchoflosttime, line 53: Some callbacks may filter by log level (eg benchmark tests may discard all debug logs). In those cases, it'd be great to just return early here instead of calling vsprintf and calling into the cb. Maybe in the `set_print` API, we could also take in the level to filter by as another parameter, and then in this function here, check whether the message should be discarded or forwarded to the cb based off that filter. What are your thoughts on this?

insearchoflosttime (commented):
Thanks for working on this!!

libbpf-rs/src/print.rs
benfogle: That's a good point. Rather than complicate the API, I changed the paramter from `&str` to `&dyn Display`. This way it will be lazily evaluated only when the user needs it.

benfogle:
I rebased off master to resolve some of the test errors I was seeing.

libbpf-rs/src/print.rs
benfogle: I'll let the maintainers weigh in on this. It seemed unnecessary to add a dependency for a single use in an obscure little function, but I'm happy to do so.

Note that the use of atomics here is just a way to have global mutable state without locking or having questionable soundness. We can't actually say it's thread safe, because the underlying implementation of `libbpf_set_print` is not thread safe.

libbpf-rs/src/print.rs
```diff
@@ -23,16 +24,31 @@ impl From<libbpf_sys::libbpf_print_level> for PrintLevel {
     }
 }
 
-pub type PrintCallback = fn(PrintLevel, &str);
+pub type PrintCallback = fn(PrintLevel, &dyn fmt::Display);
```
insearchoflosttime, line 27: I think the simplest solution is expanding the `set_print` API to take in the log level to filter by a parameter, and keeping this print callback as a `fn(PrintLevel, &str)`. One additional benefit of this is that users also then no longer have to do an if check themselves in their callback function to filter logs by the log level they want.

Other than this, your commits LGTM!

benfogle: I'm not going to push back very hard on this, because it's all minor details, but I still don't know about adding a filter parameter:

1. Should the parameter be a minimum level to show or a maximum level to hide?
2. Inclusive or exclusive?
3. Depending on the previous answers, we'll need to introduce a PrintLevel::All or PrintLevel::None that does not exist in libbpf.

I really don't think that requiring the user to write an if-statement is much of a burden, especially because I'm of the opinion that the most common use will be to pass this on to `log` or `syslog` or something, which means users will need a match to convert `PrintLevel` to `log::Level`, etc. anyway..

benfogle: Reflecting on this, maybe the real solution is to make take a `dyn Fn(...) + 'static` so that we can pass closures as well as function pointers. That would be a much nicer interface, and there's really no reason I couldn't have done it that way.

insearchoflosttime: I don't think we need to add PrintLevel::All or PrintLevel::None. The way I envision it working is hierarchically -  if the filter is set to LIBBPF_WARN, then only LIBBPF_WARN messages should go through (LIBBPF_INFO and LIBBPF_DEBUG messages should be discarded). If the filter is set to LIBBPF_INFO, then LIBBPF_WARN and LIBBPF_INFO go through while LIBBPF_DEBUG messages are discarded. And if the filter is set to LIBBPF_DEBUG messages, then all messages go through.

benfogle: How should this work with restoring a previous callback?

insearchoflosttime: We could add a getter function that will return to the user the current callback and print level, and correspondingly  `set_print` would return back nothing. If the user has a particular callback + print level they want to restore in the future, then the user is responsible for getting and keeping that around, before the user replaces `set_print` with a new callback + print level. What are your thoughts on this?

benfogle: I can implement that. We'll still need a PrintLevel::None if we want to avoid the vsprintf penalty when turning off log messages completely. Or the callback can be an Option.

anakryiko:
@benfogle, for the PrintLevel::None you mean if the users doesn't want to get any logs at all, right? I think Option is probably best, it's same as passing NULL to libbpf C api.

benfogle:
Alright, another try. Changes made:

* `set_print` takes an `Option<(level, callback)>` for a parameter.
* `get_print` returns current callback as `Option<(level, callback)>`.
* `PrintCallback` takes a `String`. (We don't do anything with the string afterwards so we might as well let the user take ownership of it.)
* Examples in the docstrings.

libbpf-rs/Cargo.toml
```diff
@@ -26,9 +26,11 @@ nix = "0.22"
...
 strum_macros = "0.21"
 vsprintf = "2.0"
+lazy_static = "1.4"
 
 [dev-dependencies]
 libc = "0.2"
 plain = "0.2.3"
 scopeguard = "1.1"
 serial_test = "0.5"
+log = "0.4"
```
insearchoflosttime, line 36: Is this needed? I'm not seeing where log is being used in the code.

libbpf-rs/src/print.rs
```diff
@@ -1,10 +1,10 @@
 use crate::*;
-use std::fmt;
+use lazy_static::lazy_static;
 use std::io::{self, Write};
 use std::os::raw::c_char;
-use std::sync::atomic::{AtomicPtr, Ordering};
+use std::sync::Mutex;
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
```
insearchoflosttime, line 7: Curious, why is Clone needed?

insearchoflosttime (commented):
Thank you for these changes, these look great. Have a few small questions asked inline.

libbpf-rs/Cargo.toml
benfogle: This is used by the doctests for the example code with `set_print` which shows passing output to the log crate.

libbpf-rs/src/print.rs
benfogle: get_print copies the (level, callback) tuple for its return value, and Copy requires Clone.

libbpf-rs/src/print.rs
```diff
@@ -0,0 +1,124 @@
...
+    if let Some((min_level, func)) = { *PRINT_CB.lock().unwrap() } {
+        if level <= min_level {
+            let msg = match unsafe { vsprintf::vsprintf(fmtstr, va_list) } {
+                Ok(s) => s,
+                Err(e) => format!("Failed to parse libbpf output: {}", e),
+            };
+            func(level, msg);
+        }
+    }
+    1 // return value is ignored by libbpf
```
anakryiko, line 64: it is ignored, but conventionally 0 means success, so might as well return 0

libbpf-rs/src/print.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// ```
+/// use libbpf_rs::{get_print, set_print};
+///
+/// let prev = get_print();
+/// set_print(None);
+/// // do things quietly
+/// set_print(prev);
+/// ```
+pub fn get_print() -> Option<(PrintLevel, PrintCallback)> {
```
anakryiko, line 122: if we have PrintLevel and PrintCallback in tuples together everywhere, we might as well return it from `set_print`. Nothing wrong with keeping `get_print` as well.

anakryiko (commented):
Same as for the other PR, please squash, provide good commit description and add your Signed-off-by:. Thanks!

benfogle:
Comments should be addressed. Let me know if the commit message looks good and if the signed-off-by line has all the correct info for this PR.

insearchoflosttime:
This LGTM. Thanks for making these changes. 

One small nit: your signed-off by line should be 

Signed-off-by: Ben Fogle <your-email-address>

not Andrii's info :) (and for more information on "signed-off-by", section 11 of https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html might be helpful). This is something we will make more clear in the contributors' guidelines.

insearchoflosttime (approved):

benfogle:
hahahaha. It's been a long week, I guess. Fixed in both PRs.

PR #140, opened 2021-11-01
update to libbpf-sys 0.5.0-2

mimullin-bbry:


libbpf-rs/src/object.rs
```diff
@@ -68,6 +68,8 @@ impl ObjectBuilder {
             attach_prog_fd: 0,
             kconfig: ptr::null(),
             btf_custom_path: ptr::null(),
+            __bindgen_padding_0: <[u8; 6]>::default(),
+            __bindgen_padding_1: <[u8; 4]>::default(),
```
anakryiko, line 72: this is interesting... 10 byte padding?.. how does that happen, any idea?

mimullin-bbry: libbpf-sys is now using rust-bindgen flag --explicit-padding due to the bug I raised for bpf_tc_opts not giving correct zeroization on default (https://github.com/alexforster/libbpf-sys/issues/25)
----
Explanation for the added padding in this specific struct

From https://github.com/libbpf/libbpf/blob/master/src/libbpf.h#L69

size_t sz(8) + char* object_name(8) + bool relaxed_maps(1) + bool relaxed_core_relocs(1) = 18 (144bits)

The next 64 bit alignment is at 192bits (24 bytes) requiring 6 bytes of padding, thus __bindgen_padding_0 of 6bytes

above (24) + char* pin_root_path (8) + u32 attach_prog_fd (4) = 36 (288bits)

The next 64 bit alignment is at 320bits (40bytes), thus __bindgen_padding_1 of 4bytes
----

Note: I'm now wondering if BPF is run on 32bit machines (doesn't arm have 32bit ebpf?)

anakryiko: Oh, so this is not the order of fields in the struct, it's just initializing two separate "fields" that correspond to two different paddings. Got it, thanks for explanation!

anakryiko (approved):
LGTM!

PR #141, opened 2021-11-05
Emit type declarations for structs in global array variables.

benfogle:
The following code causes `SkeltonBuilder` to emit invalid Rust because it doesn't define `struct foo`:

**src/sample.bpf.c**
```
struct foo {
    int x;
};

const volatile struct foo array[1] = { {0} };
```

**build.rs**
```
use libbpf_cargo::SkeletonBuilder;

fn main() {
    SkeletonBuilder::new("src/sample.bpf.c")
        .debug(true)
        .generate("src/sample_skel.rs")
        .unwrap();
}
```

Build output:
```
$ cargo build
   Compiling libbpf-cargo v0.9.2 (/home/ben/scratch/libbpf-rs/libbpf-cargo)
   Compiling example-broken v0.1.0 (/home/ben/scratch/libbpf-rs/example-broken)
error[E0412]: cannot find type `foo` in this scope
  --> src/sample_skel.rs:65:19
   |
65 |         pub bar: [foo; 1],
   |                   ^^^ not found in this scope

For more information about this error, try `rustc --explain E0412`.
error: could not compile `example-broken` due to previous error
```

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -355,6 +355,24 @@ impl<'a> Btf<'a> {
     ///
     /// `ty` must be a struct, union, enum, or datasec type.
     pub fn type_definition(&self, type_id: u32) -> Result<String> {
+        let next_type = |mut id| -> Result<Option<u32>> {
+            loop {
+                match self.type_by_id(id)? {
+                    BtfType::Struct(_)
+                    | BtfType::Union(_)
+                    | BtfType::Enum(_)
+                    | BtfType::Datasec(_) => return Ok(Some(id)),
+                    BtfType::Ptr(t) => id = t.pointee_type,
+                    BtfType::Array(t) => id = t.val_type_id,
+                    BtfType::Volatile(t) => id = t.type_id,
+                    BtfType::Const(t) => id = t.type_id,
+                    BtfType::Restrict(t) => id = t.type_id,
+                    BtfType::Typedef(t) => id = t.type_id,
+                    _ => return Ok(None),
+                }
+            }
+        };
+
         let is_terminal = |id| -> Result<bool> {
             match self.type_by_id(id)?.kind() {
                 BtfKind::Struct | BtfKind::Union | BtfKind::Enum | BtfKind::Datasec => Ok(false),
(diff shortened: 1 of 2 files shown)


libbpf-cargo/src/btf/btf.rs
```diff
@@ -355,6 +355,24 @@ impl<'a> Btf<'a> {
...
+                    BtfType::Volatile(t) => id = t.type_id,
+                    BtfType::Const(t) => id = t.type_id,
+                    BtfType::Restrict(t) => id = t.type_id,
+                    BtfType::Typedef(t) => id = t.type_id,
+                    _ => return Ok(None),
+                }
+            }
+        };
+
         let is_terminal = |id| -> Result<bool> {
```
anakryiko, line 376: is is_terminal still being used? If yes, clearly Array is not terminal, right? So need fixing as well?

libbpf-cargo/src/test.rs
```diff
@@ -753,6 +753,92 @@ fn test_skeleton_builder_clang_opts() {
...
+        prog,
+        r#"
+        #include "vmlinux.h"
+        #include <bpf/bpf_helpers.h>
+
+        struct mystruct {{
+            int x;
+        }};
+
+        const volatile struct mystruct my_array[0] = {{ {{0}} }};
```
anakryiko, line 780: can you please also check that if struct has an array member everything works? Especially if it's an array of structs. Thanks!

benfogle: Added some extra checks.

libbpf-cargo/src/btf/btf.rs
anakryiko: That's fine, let's leave it as is, thanks.

anakryiko (approved):
Looks good to me.

Can you please squash your two commits into one and provide a bit fuller commit message. Can you also put "Signed-off-by: First Last <email@whatever.bla>", following libbpf's and kernel's approach. It would be good to maintain uniformity of style going forward.

Thanks!

benfogle:
Updated. Let me know if that commit message looks good to you.

insearchoflosttime:
LGTM. Thanks for adding this, @benfogle.

Issue #143, opened 2021-11-08
All Examples should use the same vmlinux.h file


mimullin-bbry:
No need for every example to have it's own vmlinux.h file.  

Put one vmlinux_505.h inside of a ./examples/header/ directory and have all examples' vmlinux.h files soft link to that. (or something similar)

DevasiaThomas:
@mimullin-bbry I had a few questions about this :). does the vmlinux.h file vary across OS distros for the same kernel version? Is it better to compile with a vmlinux.h from a higher kernel version or lower one when considering CO-RE applications?

DevasiaThomas:
Hmm, I commented on that PR before I read this, because I was worried about CO-RE.  I'll continue the conversation there. #144

anakryiko:
> > Can the vmlinux.h file vary across OS distros for the same kernel version?
> 
> Yes. This seen when using RHEL or RHEL derivative kernels. RHEL backports a lot of bpf code (which is why you see some #ifdef RHEL stuff in tracee) to it's old kernels. Along with these backports, you'll see at minimum the additional bpf helpers.
> 
> > Is it better to compile with a vmlinux.h from a higher kernel version or lower one when considering CO-RE applications?

Generally newer kernels have more features so it makes sense to use newest vmlinux.h. You'll still have to take into considerations all the incompatible changes in old kernels. 

> 
> I don't have the full technical picture to give you a definitive answer, but if you see the the PR associated with this Issue, you'll notice a major break for the task_struct::state field. With the state field changing in 5.14, BPF code that uses the state field will not compile on kernel 5.14+, and bpf code that uses the new __state field will not pass the verifier on kernels < 5.14. (see https://github.com/libbpf/libbpf-rs/pull/144/files).
> 
> The end result of both of these answers put together is that even with CORE, you need to focus your project on a target distribution-version-kernel triple (ie ubuntu-20.04-kernel5.8aws) that has the bare minimum features you need, and as you progress to other distributions+versions+kernels you will need to make hacks (sometimes requiring recompilation, see the state vs __state).

You most certainly don't want to base this on distro-version-kernel triple if at all avoidable. BPF CO-RE provides various mechanism to both detect which specific set of features/fields kernel provides, as well as an ability to handle that within a single compiled BPF CO-RE application. See https://nakryiko.com/posts/bpf-core-reference-guide/#handling-incompatible-field-and-type-changes.

> 
[...]

PR #144, opened 2021-11-09
Each Example does not need it's own vmlinux_505.h file

mimullin-bbry:
ISSUE: https://github.com/libbpf/libbpf-rs/issues/143

Move vmlinux_505.h from each example into it's own directory at ./examples/vmlinux_header/vmlinux_505.h

Softlink each example's vmlinux.h to
examples/vmlinux_header/vmlinux_505.h

This will reduce the amount of disk space needed

Create a new file examples/vmlinux_header/vmlinux_514.h
This file was generated via bpftool on an Arch Linux machine running
Kernel 5.14.16
bpftool btf dump file /sys/kernel/btf/vmlinux format c > examples/vmlinux_header/vmlinux_514.h
Softlink this file to examples/runqslower/src/bpf/vmlinux_new.h

This new header file is required because in change
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/include/linux/sched.h?h=v5.14.17&id=2f064a59a11ff9bc22e52e9678bc601404c7cb34
the task_struct::state field was changed to __state

Fixup runqslower.bpf.c to use proper state vs __state depending on
kernel version


DevasiaThomas:
@mimullin-bbry Is there a runtime approach to deal with ``state`` vs ``__state`` in runqslower.bpf.c?  Else one would have to compile on 2 kernels? not really CO-RE?

DevasiaThomas:
Can we do what @anakryiko did in his [blogpost](https://nakryiko.com/posts/bpf-portability-and-co-re/#dealing-with-kernel-version-and-configuration-differences), the struct flavors part,  for this one?

mimullin-bbry:
@DevasiaThomas I don't think so (eager to be schooled on this issue).  What I'm reading in @anakryiko 's post is to solve runtime kernel decision making, but this issue is a compile time issue where task_struct::__state does not exist pre 5.14 and task_struct::state does not exist post 5.14.

Thus to compile runqslower using vmlinux_505.h the verifier will barf upon loading into a 5.14 kernel saying something along the lines of "symbol task_struct::state does not exist."

Since example code is intended to be run on the machine that built it, users trying to understand how to "do bpf" will have a pretty poor time with runqslower on Fedora 35, arch or any distro using a 5.14 kernel and beyond.  (note: arch has another issue where their uapi is not up to date, but thats not a problem I can fix).

DevasiaThomas:
> EDIT: Actually, after a re-read of the blogpost, I think you are correct that something better can be done. Let me give this some cycles.

Yeah I'm hoping you can use a 5.14 vmlinux.h but define the older struct(for state) within the example's code as __pre_5_1_4 and do what @anakryiko did and compile it on 5.14 but also be able to run on pre 5.14.  If that happens, everybody wins 🎉

mimullin-bbry:
So the method does require duplicate definition of task_struct, which IMHO is not acceptable for a learning tool.  I don't think it's in the projects best interest to copy such a large structure with such a large amount of compile time options in it.

https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L723

mimullin-bbry:
I've updated to use the techniques in https://nakryiko.com/posts/bpf-portability-and-co-re/#dealing-with-kernel-version-and-configuration-differences

I have not however, tested on a pre 5.13 kernel. (will update this comment when that is complete).

Pushing pre-testing for potential feedback w.r.t. copying the 505 version of struct task_struct, and whether that is a direction the project wants to take.

DevasiaThomas:
Thats good news @mimullin-bbry . How did you deal with the compile time ifdefs for the task struct?

anakryiko:
See https://github.com/iovisor/bcc/pull/3659/files for how it was done for runqslower tool, for example.

You might also find https://nakryiko.com/posts/bpf-core-reference-guide/#handling-incompatible-field-and-type-changes useful.

Either way, I'd rather have only one vmlinux.h version to reduce the size of the repo. So we can update to the latest vmlinux.h, but we should get rid of the old one and update examples accordingly.

examples/runqslower/src/bpf/runqslower.bpf.c
```diff
@@ -62,6 +66,16 @@ int handle__sched_wakeup_new(u64 *ctx)
 	return trace_enqueue(p->tgid, p->pid);
 }
 
+static inline long get_task_state(struct task_struct *ts)
+{
+	if (LINUX_KERNEL_VERSION < KERNEL_VERSION(5, 14, 0)) {
```
anakryiko, line 71: see https://github.com/iovisor/bcc/pull/3659/files, there are better ways than using LINUX_KERNEL_VERSION check. Due to backporting, kernel version isn't always representative. CO-RE provides bpf_core_field_exists() check, we should use that here.

examples/vmlinux_header/vmlinux_505_structs.h
```diff
@@ -0,0 +1,181 @@
...
+struct task_struct_505 {
+	struct thread_info thread_info;
+	volatile long int state;
+	void *stack;
+	refcount_t usage;
+	unsigned int flags;
+	unsigned int ptrace;
+	struct llist_node wake_entry;
+	int on_cpu;
+	unsigned int cpu;
```
anakryiko, line 13: you don't need all the fields of task_struct, only few necessary fields can be defined. It's also better to leave such definitions close to BPF source code that's using this type. See https://nakryiko.com/posts/bpf-core-reference-guide/#handling-incompatible-field-and-type-changes for some more examples and details.

anakryiko (changes requested):
let's put vmlinux_514.h under examples/ and create a symlink vmlinux.h -> vmlinux_514.h right there.

Also, I was expecting that search path to vmlinux.h would need to be changed somewhere either in build script or somewhere else. Don't see it here. What am I missing?

examples/vmlinux_header/vmlinux_505_structs.h
mimullin-bbry: >  Also, I was expecting that search path to vmlinux.h would need to be changed somewhere either in build script or somewhere else. Don't see it here. What am I missing?

Not sure what you mean by this.  I've moved vmlinux_514.h into the examples directory, and softlinked a vmlinux.h to it.  I then softlink each examples vmlinux.h to the example/vmlinux.h file.  The build scripts do not need to be aware of the vmlinux.h file.

```
pwd && ls -l 
/code/single-vmlinux/examples/runqslower/src/bpf
total 8
-rw-r--r-- 1 mm mm 2677 Nov 10 20:46 runqslower.bpf.c
-rw-r--r-- 1 mm mm  232 Nov  8 19:00 runqslower.h
lrwxrwxrwx 1 mm mm   18 Nov 10 19:15 vmlinux.h -> ../../../vmlinux.h
```

mimullin-bbry:
Closing as I had problems squashing and merging.  Please see https://github.com/libbpf/libbpf-rs/pull/149

examples/vmlinux_header/vmlinux_505_structs.h
anakryiko: Never mind, I forgot we do vmlinux.h symlinking in each example to avoid needing to specify extra -I<path> to the compiler. Please keep vmlinux_514.h name and leave all the vmlinux.h as symlinks

DevasiaThomas: > you don't need all the fields of task_struct, only few necessary fields can be defined.

@anakryiko I didn't know you could do this neither do I know the concept behind this. The closes thing I know is in OOPs where you can assign child object to variable of type parent and just refer to the fields the parent knows about. Is this only possible in bpf CO-RE code or can you do this in regular C as well? I would love to read about this concept if its possible in regular C.

anakryiko: It's BPF CO-RE thing. Only fields that are accessed are necessary (to let compiler generate CO-RE relocation that will be used by libbpf to find corresponding type and field in the full kernel BTF definition).

PR #152, opened 2021-11-23
libbpf-rs: Fix PerfBuilder callbacks lifetime definition

rgushchin:
Currently callbacks passed to PerfBuilder are assumed to have the
'static lifetime, which makes them hard to use.

A simple example based on the runqslower: let's say we want to store
events in a hashmap instead of printing them. We can't pass a mutable
reference to a hashmap into a closure and use the closure as a
sample_cb callback without giving an ownership with a "move" keyword.
But then we can't use the map afterwards.

Actually, handlers lifetime should be bound to the lifetime of
PerfBuffer object. Once it's gone, we don't need callbacks anymore.

After this change the following code can work:

    let mut hm = HashMap::new();

    {
	let perf = PerfBufferBuilder::new(skel.maps_mut().events())
            .sample_cb(|cpu: i32, data: &[u8]| { handle_event(&mut hm, cpu, data); })
            .lost_cb(handle_lost_events)
            .build()?;

	loop {
            perf.poll(Duration::from_millis(100))?;
	    break;
	}
    }

    for (k, v) in &hm {
	println!("{}:{}", k, v);
    }

Signed-off-by: Roman Gushchin <guro@fb.com>

### DIFF
--- a/libbpf-rs/src/perf_buffer.rs
+++ b/libbpf-rs/src/perf_buffer.rs
@@ -12,26 +12,26 @@ fn is_power_of_two(i: usize) -> bool {
 // Workaround for `trait_alias`
 // (https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html)
 // not being available yet. This is just a custom trait plus a blanket implementation.
-pub trait SampleCb: FnMut(i32, &[u8]) + 'static {}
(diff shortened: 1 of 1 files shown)


insearchoflosttime (approved):
This looks good to me! Thank you for adding this, @rgushchin!

PR #158, opened 2021-12-09
Test Updates for libbpf-rs

mimullin-bbry:
There is an error in the test version of runqslower.bpf.c which was addressed in a previous commit to the examples/ code, but not updated in the testing code.

To address this issue
patch 0: build a small infrastructure to easily recompile the bpf objects contained in libbpf-rs/tests/bin.  
patch 1: Then update runqslower in the libbpf-rs/tests/bin/src directory and recompile the bpf object
patch 2: libbpf now has .bss as a map, and the test.rs code must be updated to observe that test.
patch 3: two tests are racing on a pined map. Rename the map to be explicit to the test being run

NOTE: Apologies for replying to comments using the masmullin2000 address.  I was incorrectly logged in to my personal github account while replying.  masmullin2000 and mimullin-bbry are the same person.


libbpf-rs/tests/test.rs
```diff
@@ -98,8 +98,8 @@ fn test_object_maps_iter() {
     for map in obj.maps_iter() {
         eprintln!("{}", map.name());
     }
-    // This will include .rodata, so our expected count is 3, not 2
-    assert!(obj.maps_iter().count() == 3);
+    // This will include .rodata and .bss, so our expected count is 4, not 2
+    assert!(obj.maps_iter().count() == 4);
```
insearchoflosttime, line 102: Thanks for fixing this! I'm going to look into whether we can set up some sort of CI for the libbpf-rs tests

insearchoflosttime (commented):
For some reason it's not letting me comment on commit #2, so I will just paste the comments here :) 


I like the idea of keeping the tests runqslower and the examples runqslower separate - in the future, there could be differences between the two (eg adding things to the test's runqslower that aren't relevant to the example's). 

I think we should also keep the runqslower.bpf.o binary so that users can directly run the test without knowing they have to run a script first. 

Btw, I think you also forgot to remove the last line of the commit message in commit #2 :)

libbpf-rs/tests/bpf_object_regen.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+SRC=`find bin/src/ -name "*.bpf.c"`
+
+function compile() {
+    base=$(basename -- $file)
+    objname="${base%.*}"
+    objname+=".o"
+    clang -g -O2 -target bpf -c -D__TARGET_ARCH_x86 $1 -o bin/$objname
```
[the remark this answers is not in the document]
masmullin2000: > Will this work on arm64 machines as well?

the target_arch thing is unnecessary, I shall remove.

insearchoflosttime:
> > I think we should also keep the runqslower.bpf.o binary so that users can directly run the test without knowing they have to run a script first.
> 
> I agree. The binary is kept, but updated so that it will run on more recent kernels, along with the kernels it currently works on. Running the script is not necessary per my changes.

Ah, gotcha. I thought your change was removing the binary altogether but looking more closely at that file, it looks like you're just updating it :)

libbpf-rs/tests/bpf_object_regen.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
insearchoflosttime, line 1: When I try running this script locally on the latest master, I'm seeing 
```
[~/local/libbpf-rs/libbpf-rs/tests] ./bpf_object_regen.sh
bin/src/percpu_map.bpf.c:3:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
bin/src/ringbuf.bpf.c:4:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
bin/src/runqslower.bpf.c:4:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
bin/src/taskiter.bpf.c:3:10: fatal error: 'bpf/bpf_helpers.h' file not found
#include <bpf/bpf_helpers.h>
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

Does it compile cleanly for you?

mimullin-bbry: Yes, but I have a full libbpf as provided by my distribution installed.  Meaning that I have /usr/include/bpf/bpf_helpers.h on my machine.

libbpf-sys provides the bpf helpers for a normal compilation process of libbpf-rs and libbpf-cargo, but this is a raw compile to generate the objects without the libbpf-cargo assistance EG outside of the project.

I'm not sure if there is a non-complex way of grabbing the version of the headers as provided by the github-repo version of libbpf that libbpf-sys uses.  I can look into this if it's demanded, but it would create a very complicated script, and the intention was to be simple to read/understand, as it should be used very rarely.

I would prefer to update the test's readme to tell users to grab the appropriate packages from their distribution.

insearchoflosttime: Awesome, do you mid updating the test's readme to include this info? thanks!

mimullin-bbry: Done, all distributions described (unless explicitly marked untested) have been tested with the procedures given. (pfeww I need a better method of testing multiple distros)

insearchoflosttime (approved):

insearchoflosttime:
Thank you for your contribution, @mimullin-bbry!

PR #162, opened 2021-12-23
libbpf-rs: Add functionality for the libbpf TC interface

mimullin-bbry:
Documentation of the libbpf TC interface can be found here
https://lwn.net/ml/bpf/20210512103451.989420-3-memxor@gmail.com/


anakryiko:
I'll be out on vacation for the rest of 2021, so I'll do a proper review after New Year. But it would be great to split out the example into a separate commit, if you don't mind. Thank you!

examples/tc_port_whitelist/src/bpf/tc.bpf.c
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    if (tcph) {
+        dst = tcph->dest;
+        src = tcph->source;
+    } else if (udph) {
+        dst = udph->dest;
+        src = udph->source;
+    }
+
```
insearchoflosttime, line 237: I think we can greatly simplify this file if we just do the parsing in this function instead of defining `is_ipv4`, `is_ipv6`, `is_udp` and `is_tcp` in separate functions. I think we can just do this:

```
  __u8 proto = 0;
  void *trans_data;
  if (eth + 1 > data_end) 
      return TC_ACT_UNSPEC;
  if (eth->h_proto == bpf_htons(ETH_P_IP)) { // ipv4
      struct iphdr *iph = (struct iphdr *)(eth + sizeof(*eth));
      if (iph + 1 > data_end)
          return TC_ACT_UNSPEC;
      proto = iph->protocol;
      trans_data = iph + (iph->ihl * 4);
  } else if (eth->h_proto == bpf_htons(ETH_P_IPV6)) { // ipv6
      struct ipv6hdr *ip6h = (struct ipv6hdr *)(data + sizeof(*eth));
      if (ip6h + 1 > data_end)
          return TC_ACT_UNSPEC;
     proto = ip6h->nexthdr;
      trans_data = ip6h + 1;
  }
   if (proto == IPPROTO_TCP)  {
       tcph = (struct tcphdr *)trans_data;
       dst = tcph->dst;
       src = tcph->source;
   } else if (proto == IPPROTO_UDP) {
       udph = (struct udphdr *)trans_data;
       dst = udph->dest;
       src = udph->source;
   } else {
     goto allow_unknown;
  }
   
```

examples/tc_port_whitelist/src/bpf/tc.bpf.c
```diff
@@ -0,0 +1,256 @@
...
+    if (hdr_sz == sizeof(struct iphdr)) {
+        struct iphdr *v4 = (struct iphdr*)iph;
+        proto = v4->protocol;
+    } else if (hdr_sz == sizeof(struct ipv6hdr)) {
+        struct ipv6hdr *v6 = (struct ipv6hdr*)iph;
+        proto = v6->nexthdr;
+    }
+
+    if (proto == IPPROTO_UDP) {
+        udph = (struct udphdr*)((void*)iph + hdr_sz);
```
insearchoflosttime, line 116: I don't think this is correct - ipv4 headers are variable size lengths so in the ipv4 case, the tcp or udp header would be at `iph + (iph->ihl * 4)`

examples/tc_port_whitelist/src/bpf/tc.bpf.c
```diff
@@ -0,0 +1,256 @@
...
+    if (hdr_sz == sizeof(struct iphdr)) {
+        struct iphdr *v4 = (struct iphdr*)iph;
+        proto = v4->protocol;
+    } else if (hdr_sz == sizeof(struct ipv6hdr)) {
+        struct ipv6hdr *v6 = (struct ipv6hdr*)iph;
+        proto = v6->nexthdr;
+    }
+
+    if (proto == IPPROTO_TCP) {
+        tcph = (struct tcphdr*)((void*)iph + hdr_sz);
```
insearchoflosttime, line 153: Same as the above comment regarding ipv4 header sizes

libbpf-rs/tests/bin/src/tc-unit.bpf.c
```diff
@@ -0,0 +1,21 @@
...
+#define TC_ACT_UNSPEC	(-1)
+#define TC_ACT_OK		0
+#define TC_ACT_RECLASSIFY	1
+#define TC_ACT_SHOT		2
+#define TC_ACT_PIPE		3
+#define TC_ACT_STOLEN		4
+#define TC_ACT_QUEUED		5
+#define TC_ACT_REPEAT		6
+#define TC_ACT_REDIRECT		7
+#define TC_ACT_TRAP		8
```
insearchoflosttime, line 14: Since only TC_ACT_UNSPEC is being used, let's just define only that.

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,288 @@
+use nix::errno;
+use nix::errno::Errno::{EEXIST, EINVAL, ENOENT};
+
+use crate::*;
+
+pub type TcAttachPoint = libbpf_sys::bpf_tc_attach_point;
+pub const INGRESS: TcAttachPoint = libbpf_sys::BPF_TC_INGRESS;
+pub const EGRESS: TcAttachPoint = libbpf_sys::BPF_TC_EGRESS;
+pub const CUSTOM: TcAttachPoint = libbpf_sys::BPF_TC_CUSTOM;
```
insearchoflosttime, line 9: I think we should rename it with the preface TC_ (or maybe not even redeclare it here and just use libbpf_sys's declaration everywhere?). I think ingress/egress/custom is too common and doesn't pertain to just tc (eg cgroup uses it too afaik).

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,288 @@
...
+
+pub type TcFlags = libbpf_sys::bpf_tc_flags;
+pub const BPF_TC_F_REPLACE: TcFlags = libbpf_sys::BPF_TC_F_REPLACE;
+
+pub const TC_H_INGRESS: u32 = 0xFFFFFFF1;
+pub const TC_H_CLSACT: u32 = TC_H_INGRESS;
+
+pub const TC_H_MIN_INGRESS: u32 = 0xFFF2;
+pub const TC_H_MIN_EGRESS: u32 = 0xFFF3;
```
insearchoflosttime, line 18: Can you add a comment here saying these are taken  from `include/uapi/linux/pkt_sched.h`?

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,288 @@
...
+        };
+
+        tc_hook.hook.sz = std::mem::size_of::<libbpf_sys::bpf_tc_hook>() as u64;
+        tc_hook.opts.sz = std::mem::size_of::<libbpf_sys::bpf_tc_opts>() as u64;
+        tc_hook.opts.prog_fd = fd;
+
+        tc_hook
+    }
+
+    /// create a TcHook creates a New bpf_tc_hook
```
insearchoflosttime, line 56: nit: please fix this comment

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Set the parent of a hook
+    /// Will return EINVAL if set upon an EGRESS/INGRESS/EGRESS|INGRESS hook
+    ///
+    /// Must be set on a CUSTOM hook
+    /// Current acceptable values are TC_H_CLSACT for maj, and
+    /// TC_H_MIN_EGRESS or TC_H_MIN_INGRESS for min
+    pub fn parent(&mut self, maj: u32, min: u32) -> &mut Self {
+        /* values from libbpf.h BPF_TC_PARENT() */
+        let parent = (maj & 0xFFFF0000_u32) | (min & 0x0000FFFF_u32);
```
insearchoflosttime, line 100: Please define 0xFFFF0000_u32 and 0x0000FFFF_u32 as TC_H_MAJ_MASK and TC_H_MIN_MASK  and use TC_H_MAJ_MASK/TC_H_MIN_MASK

insearchoflosttime (changes requested):
Thanks for working on this.

masmullin2000:
> Thanks for working on this.

Updated PR.  Sorry it took so long to respond.

examples/query/src/main.rs
```diff
@@ -31,7 +31,7 @@ fn prog() {
 
 fn map() {
     for map in query::MapInfoIter::default() {
-        println!("name={:<16} type={}", map.name, map.ty.to_string(),);
+        println!("name={:<16} type={}", map.name, map.ty);
```
masmullin2000, line 34: This is a cargo clipyy fix for a new default linter (found using rust 1.58.1)

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+        self.opts.priority = priority;
+        self
+    }
+
+    /// Query a hook to inspect the program identifier (prog_id)
+    pub fn query(&mut self) -> Result<u32> {
+        let mut opts = self.opts;
+        opts.prog_id = 0;
+        opts.prog_fd = 0;
+        opts.flags = 0;
```
insearchoflosttime, line 140: I'm new to bpf tc - would you  mind explaining why prog_fd and prog_id have to be 0 when calling bpf_tc_query?

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// Users can create a second hook by changing the handle, the priority
+    /// or the attach_point and calling the attach() method again.
+    /// Beware doing this.  It might be better to Copy the TcHook and change
+    /// the values on the copied hook for easier detach()
+    ///
+    /// NOTE: Once a TcHook is attached, it, and the maps it uses,
+    ///       will outlive the userspace application that spawned them
+    ///       Make sure to detach if this is not desired
+    pub fn attach(&mut self) -> Result<Self> {
+        self.opts.prog_id = 0;
```
insearchoflosttime, line 164: Same here - would you mind explaining why  you can't attach when the prog_id is non-zero?

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    /// Detach a TcHook
+    pub fn detach(&mut self) -> Result<()> {
+        let mut opts = self.opts;
+        opts.prog_id = 0;
+        opts.prog_fd = 0;
+        opts.flags = 0;
+
+        let err = unsafe { libbpf_sys::bpf_tc_detach(&self.hook as *const _, &opts as *const _) };
+        if err != 0 && err != -(errno::Errno::ENOENT as i32) {
```
insearchoflosttime, line 182: I don't see the path in netlink.c where ENOENT can be returned - can you point me to where this is?

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// If called on a hook with an attach_point of EGRESS|INGRESS, will destroy the clsact tc
+    /// qdisc and detach all hooks
+    ///
+    /// Will error with EOPNOTSUPP if attach_point is BPF_TC_CUSTOM
+    ///
+    /// It is good practice to query before destroying as the tc qdisc may be used
+    /// by multiple programs
+    pub fn destroy(&mut self) -> Result<()> {
+        let err = unsafe { libbpf_sys::bpf_tc_hook_destroy(&mut self.hook as *mut _) };
+        if (err == -(EINVAL as i32) || err == -(ENOENT as i32))
```
insearchoflosttime, line 204: Same here - I'm not seeing where the path to ENOENT is in netlink.c for bpf_tc_hook_destroy. Where do you see this?

insearchoflosttime, line 204: I see EINVAL can also get returned with attach point == TC_EGRESS | TC_INGRESS, in the case where the parent of the hook is missing (in specific, I'm looking at the attach_point_to_config function in netlink.c). It seems like this shouldn't return Ok()

examples/tc_port_whitelist/src/bpf/tc.bpf.c
```diff
@@ -0,0 +1,297 @@
...
+ * @param data_end, end of the entire packet
+ *
+ * @return a struct iphdr if this is an ipv4 packet
+ *         NULL otherwise
+ */
+/*static struct iphdr* is_ipv4(struct ethhdr *eth, void *data_end) */
+/*{*/
+    /*struct iphdr *iph = NULL;*/
+    /*if (!eth || !data_end) {*/
+        /*return NULL;*/
```
insearchoflosttime, line 37: did you mean to remove the commented out code from this commit?

insearchoflosttime (commented):
Thanks for the updates. Left a few comments + questions inline

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+pub const TC_H_CLSACT: u32 = TC_H_INGRESS;
+pub const TC_H_MIN_INGRESS: u32 = 0xFFF2;
+pub const TC_H_MIN_EGRESS: u32 = 0xFFF3;
+pub const TC_H_MAJ_MASK: u32 = 0xFFFF0000;
+pub const TC_H_MIN_MASK: u32 = 0x0000FFFF;
+
+/// The BPF TC subsystem has different control paths from other BPF programs
+/// As such a BPF program using a TC Hook (SEC("classifier") | SEC("tc")) must be operated
+/// more independently from other [libbpf-rs::Program]s where the program itself can
+/// be attach_*() a program using TC must grab a hook point, or create one
```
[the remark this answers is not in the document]
mimullin-bbry: the attach_*() refers to the other "normal" methods of using libbpf-rs such as attach_xdp attach_kprobes, etc.  

My grammar here is poor, I will address.

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+    /// create a new TcHook
+    ///
+    /// if a TcHook already exists with the same parameters as the hook calling
+    /// create(), this function will still succeed.
+    ///
+    /// Will always fail on a TC_CUSTOM hook
+    pub fn create(&mut self) -> Result<Self> {
+        let err = unsafe { libbpf_sys::bpf_tc_hook_create(&mut self.hook as *mut _) };
+        // the hook may already exist, this is not an error
+        if err != 0 && err != -(EEXIST as i32) {
```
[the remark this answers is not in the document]
mimullin-bbry: > I'm not seeing where bpf_tc_hook_create can return -EEXIST

This error is coming out of the kernel rather than libbpf.  I shall track it down.

I **think** the error is coming out from here (https://elixir.bootlin.com/linux/latest/source/net/sched/cls_api.c#L2111).  But this is a wild guess.  During development I noticed the error during a double create, and tried to track through libbpf to see where the EEXIST comes from, and stopped as I could not track it via libbpf code spelunking.

libbpf-rs/src/tc.rs
mimullin-bbry: https://lwn.net/ml/bpf/20210512103451.989420-3-memxor@gmail.com/

There is a hard check for the values being unset here:

https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/netlink.c#L724

libbpf-rs/src/tc.rs
mimullin-bbry: This is also coming out of the kernel.  There are many spots in cls_api.c that it could be and I need to track it down to give a good answer.  Please allow some time.

libbpf-rs/src/tc.rs
mimullin-bbry: My semantic intention was to return an Ok() if the hooks were "gone".  A hook with a missing parent in my opinion is semantically the same as destroyed as I don't think a EGRESS|INGRESS hook can exist without some sort of parent.

EG if destroy() is called twice, my opinion is that the second destroy should not return an error as it's semantically a no-op.

If you would prefer, I can enable the error for the EINVAL case, it just means the consumers of the API will need to pay a bit more attention to which error value they receive as an EINVAL shouldn't dramatically alter their program flow.

libbpf-rs/src/tc.rs
mimullin-bbry: https://lwn.net/ml/bpf/20210512103451.989420-3-memxor@gmail.com/

There is a hard check for the values being unset here:

https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/netlink.c#L724

libbpf-rs/src/tc.rs
mimullin-bbry: It's a demand from the documentation
https://lwn.net/ml/bpf/20210512103451.989420-3-memxor@gmail.com/

There is a hard check for unset here:
https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/netlink.c#L579

examples/tc_port_whitelist/src/bpf/tc.bpf.c
mimullin-bbry: Yes.  Sorry.  will address.

libbpf-rs/src/tc.rs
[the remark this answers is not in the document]
insearchoflosttime: > the attach_*() refers to the other "normal" methods of using libbpf-rs such as attach_xdp attach_kprobes, etc.

Ah I see! That makes a lot of sense, thanks for clarifying!

libbpf-rs/src/tc.rs
insearchoflosttime: Why does the documentation demand that? That's what I'm trying to figure out  - to me it seems okay that you would be able to query even if you have the prog_fd in the hook set.

libbpf-rs/src/tc.rs
insearchoflosttime: Same here - the documentation says "The following opts must be unset: prog_id - The ID of the BPF prog", but why is there this restriction? If you know the ID of the  bpf prog, why would that prohibit you from attaching a filter to the hook? 

(btw, these aren't questions against your patch - they're more just questions about the implementation in libbpf and I wanted to ask you in case you have more context about bpf tc).

libbpf-rs/src/tc.rs
insearchoflosttime: Gotcha, thanks for your explanation. I personally like the idea of propagating whatever error libbpf returns - I think this is more future proof where if in the future libbpf returns -EINVAL for some other edge case, that will be automatically propagated by libbpf-rs. I'm also fine with leaving as is though, so the call is up to you :)

libbpf-rs/src/tc.rs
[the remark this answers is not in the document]
insearchoflosttime: I think it's coming from here https://elixir.bootlin.com/linux/latest/source/net/sched/sch_api.c#L1625 - `bpf_tc_hook_create` triggers the `tc_modify_qdisc` function in `sch_api.c` since we send  a `RTM_NEWQDISC` message when we create the tc hook (in tools/lib/bpf/netlink.c)

libbpf-rs/src/tc.rs
insearchoflosttime: I think it's coming from `tc_del_tfilter` in net/sched/cls_api.c. -ENOENT is returned if the filter wasn't found (https://elixir.bootlin.com/linux/latest/source/net/sched/cls_api.c#L2294) but looking at that function, -ENOENT can also be returned if the filter's protocol, handle, or kind is set, in which case we shouldn't return back OK() to the user. I'm in favor of propagating -ENOENT to the user whenever libbpf_sys::bpf_tc_detach returns back -ENOENT.

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+        let err =
+            unsafe { libbpf_sys::bpf_tc_attach(&self.hook as *const _, &mut self.opts as *mut _) };
+        if err != 0 {
+            Err(Error::System(errno::errno()))
+        } else {
+            Ok(*self)
+        }
+    }
+
+    /// Detach a TcHook
```
insearchoflosttime, line 174: Can you clarify  this as "Detach a filter from a TcHook"?

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,291 @@
...
+
+        let err = unsafe { libbpf_sys::bpf_tc_query(&self.hook as *const _, &mut opts as *mut _) };
+        if err != 0 {
+            Err(Error::System(errno::errno()))
+        } else {
+            Ok(opts.prog_id)
+        }
+    }
+
+    /// Attach the TcHook so that the program starts processing
```
insearchoflosttime, line 150: Would you mind clarifying this to  "Attach a filter to a TcHook"?

masmullin2000: Resolved

libbpf-rs/src/tc.rs
masmullin2000: Resolved

examples/tc_port_whitelist/src/bpf/tc.bpf.c
masmullin2000: Resolved

libbpf-rs/src/tc.rs
masmullin2000: As of most recent push, all errors are passed to the user.  With the exception of create().  IMHO erroring on a "double create" is very difficult for users to understand (see test_tc_double_create in test_tc.rs)

libbpf-rs/src/tc.rs
masmullin2000: resolved, see comment below

libbpf-rs/src/tc.rs
masmullin2000: I've been looking for a while now, and I dont see why prog_fd or flags particularly matter.  I'm probably missing something.

If found this discussion on the LKML (http://lkml.iu.edu/hypermail/linux/kernel/2104.0/02018.html), and I wonder if it has anything to do with the paragraph where the Author describes "Suppose a user doesn't properly fill the opts during query (e.g. in case of filters)"

libbpf-rs/src/tc.rs
mimullin-bbry: https://elixir.bootlin.com/linux/latest/source/net/sched/sch_api.c#L1485

Resolved by returning the errors back to the user.

libbpf-rs/src/tc.rs
```diff
@@ -0,0 +1,284 @@
...
+        self.hook.attach_point = ap;
+        self
+    }
+
+    /// Set the parent of a hook
+    /// Will return EINVAL if set upon an TC_EGRESS/TC_INGRESS/TC_EGRESS|TC_INGRESS hook
+    ///
+    /// Must be set on a TC_CUSTOM hook
+    /// Current acceptable values are TC_H_CLSACT for maj, and
+    /// TC_H_MIN_EGRESS or TC_H_MIN_INGRESS for min
```
insearchoflosttime, line 98: I don't see where we enforce this in `pub fn parent`. Is this documentation meant to be placed here?

examples/tc_port_whitelist/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+    if opts.attach {
+        for (i, port) in opts.ports.iter().enumerate() {
+            let key = (i as u32).to_ne_bytes();
+            let val = port.to_ne_bytes();
+            if let Err(e) = skel.maps_mut().ports().update(&key, &val, MapFlags::ANY) {
+                bail!("Example limited to 10 ports: {}", e);
+            }
+        }
+        ingress.create()?;
+
```
insearchoflosttime, line 116: Do we need to call `egress.create()` before we call `egress.attach()`?

insearchoflosttime:
This revision looks great, thanks for making all the changes! I left a few comments inline on the latest commits, but they are fairly minor

examples/tc_port_whitelist/src/main.rs
mimullin-bbry: Yes.  If a previous destroy was called on INGRESS|EGRESS, or if this is the first time the bpf tc is run from bootup, the parent clsact Qdisc will not exist.

The libbpf function bpf_tc_hook_create will create this qdisc, along with the bpf_hook that can then be attached to.

libbpf-rs/src/tc.rs
mimullin-bbry: The documentation is unclear.  If a parent is set upon a non-TC_CUSTOM hook, the attach() call will return EINVAL, as setting the parent is only valid for TC_CUSTOM.

I've updated the documentation to be more clear.

libbpf-rs/tests/test_tc.rs
```diff
@@ -0,0 +1,307 @@
...
+    (obj, tc_builder, fd)
+}
+
+#[test]
+#[serial]
+fn test_tc_basic_cycle() {
+    bump_rlimit_mlock();
+
+    let (_obj, tc_builder, fd) = test_helper_get_tc_builder("handle_tc");
+    //assert!(!destroyer.destroy().is_err());
```
insearchoflosttime, line 53: Did you mean to remove this commented out line?

libbpf-rs/tests/test_tc.rs
```diff
@@ -0,0 +1,307 @@
...
+fn clear_clsact(fd: i32) -> Result<()> {
+    // Ensure clean clsact tc qdisc
+    let mut destroyer = TcHook::new(fd);
+    destroyer
+        .ifindex(LO_IFINDEX)
+        .attach_point(TC_EGRESS | TC_INGRESS);
+
+    let res = destroyer.destroy();
+    if let Err(Error::System(err)) = res {
+        if err != -(ENOENT as i32) && err != -(EINVAL as i32) {
```
insearchoflosttime, line 24: Should -EINVAL  be ignored or should that be treated  as a real error?

examples/tc_port_whitelist/src/main.rs
insearchoflosttime: Gotcha, thanks for the explanation!

> The libbpf function bpf_tc_hook_create will create this qdisc, along with the bpf_hook that can then be attached to.

Does this mean we don't need the ingress.create() on line 115 then?

insearchoflosttime:
I left a few questions on some things I'm still a bit confused about, but this looks great, thank you for your work on this!

libbpf-rs/tests/test_tc.rs
insearchoflosttime: Gotcha. If I'm understanding this correctly, what you're saying is that -EINVAL is not an error since the ingress queue handle is 0 in the success case (where the ingress queue is created and handle is kzalloc-ed). Is my understanding correct?

insearchoflosttime:
Would you mind running rustfmt as well? I'm seeing a trivial rustfmt error in the build. Thanks!

insearchoflosttime:
@mimullin-bbry I'd love to get this merged in. Would you mind rebasing this onto master to resolve any merge conflicts?

insearchoflosttime:
@mimullin-bbry would you mind squashing the commits into 2 commits (libbpf-rs changes and the example) and then rebasing onto master? I will merge these changes in right after. Thanks.

Issue #164, opened 2021-12-29
Hard error if a `package.metadata.*` section exists that isn't `.libbpf`


jyn514:
```
$ cargo libbpf make
Compiling BPF objects
Error: Failed to compile BPF objects

Caused by:
    Failed to process package=bpf-graceful-restart, error=unknown variant `deb`, expected `libbpf`
```

The short-term workaround is to comment out the `[package.metadata.deb]` section, but it would be nice for it to just work.

PR #165, opened 2021-12-29
Allow having package.metadata sections that aren't for libbpf

jyn514:
This allows having sections of the following form in Cargo.toml:
```
[package.metadata.deb]
    # ...
```

Previously, they would give a hard error:

```
$ cargo libbpf make
Compiling BPF objects
Error: Failed to compile BPF objects

Caused by:
    Failed to process package=bpf-graceful-restart, error=unknown variant `deb`, expected `libbpf`
```

Serde by default requires all enum variants to be known ahead of time;
this works by changing the deserialized type to a struct instead.

Signed-off-by: Joshua Nelson <jnelson@cloudflare.com>

Fixes https://github.com/libbpf/libbpf-rs/issues/164.

### DIFF
--- a/libbpf-cargo/src/metadata.rs
+++ b/libbpf-cargo/src/metadata.rs
@@ -15,8 +15,9 @@ struct LibbpfPackageMetadata {
 
 #[derive(Deserialize)]
 #[serde(rename_all = "lowercase")]
-enum PackageMetadata {
-    Libbpf(LibbpfPackageMetadata),
+struct PackageMetadata {
+    #[serde(default)]
+    libbpf: LibbpfPackageMetadata,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


insearchoflosttime:
Thanks for adding this. One question - it looks like there can be the case where if there's some non-libbpf package.metadata section that has the same keys as `LibbpfPackageMetadata`, then we'll automatically interpret those values as libbpf package.metadata values instead of setting to default values. Is there some way to avoid this case?

jyn514:
@insearchoflosttime naming the field `libbpf` should make it only look at `.libbpf` sections - I added a test making sure that's the case :)

insearchoflosttime:
Great!! This LGTM.

insearchoflosttime (approved):

PR #167, opened 2022-01-02
libbpf-cargo: Cannot generate unnamed unions

mimullin-bbry:
An unnamed union in C cannot be generated to rust code as rust demands
names on it's variables.

Solution: Give the union a name of __unnamed_X where X is the
declaration of the union.

NOTE: The usage of __unnamed_ and __anon_ requires the consumer-programmer of the skeleton have intimate knowledge of the generated skeleton.   Please see the usage of __unnamed _anon_1 and __unnamed_anon_4 in the test code.  IMHO this is exceptionally confusing to the users of libbpf-cargo.
We need to figure out the best way to make this easier for consumers; but this code will at least stop compile breaking unnamed union code.

### DIFF
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -2043,3 +2043,105 @@ pub struct Foo {
 
     assert_definition(&btf, struct_foo, expected_output);
 }
+
+#[test]
+fn test_btf_dump_definition_unnamed_union() {
+    let prog_text = r#"
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+
+// re-typed 'struct bpf_sock_tuple tup' from vmlinux as of kernel 5.15
+// with a little bit added for additional complexity testing
+struct bpf_sock_tuple_5_15 {
+	union {
+		struct {
+			__be32 saddr;
+			__be32 daddr;
+			__be16 sport;
+			__be16 dport;
+		} ipv4;
+		struct {
+			__be32 saddr[4];
+			__be32 daddr[4];
+			__be16 sport;
+			__be16 dport;
+		} ipv6;
+	};
+
+    union {
+        int a;
+        char *b;
+    };
+};
+struct bpf_sock_tuple_5_15 tup;
+"#;
+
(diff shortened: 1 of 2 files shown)


insearchoflosttime (approved):

libbpf-cargo/src/test.rs
```diff
@@ -2043,3 +2043,105 @@ pub struct Foo {
...
+    };
+};
+struct bpf_sock_tuple_5_15 tup;
+"#;
+
+    let expected_output = r#"
+#[derive(Debug, Default, Copy, Clone)]
+#[repr(C)]
+pub struct bpf_sock_tuple_5_15 {
+    pub __unnamed___anon_1: __anon_1,
```
anakryiko, line 2083: couldn't the field name be the same as type name? Rust doesn't allow this?

anakryiko (commented):
Curious, was this a situation where anonymous unions appear from some pre-defined C type or user actually defined their own types with unions? In the latter case I'd imagine users would be able to restructure their types to avoid this issue. But if it's the former, then I'm just curious about the situation.

mimullin-bbry:
> was this a situation where anonymous unions appear from some pre-defined C type

I was fiddling around with an application which will utilize the TC stuff when/if we get it up-to-snuff for inclusion in the project, and I wanted to transmit up the bpf_socket_tuple structure.  It bombed out missing a name.  My first thought was "grrr I hate it when C developers don't name their unions" and my second thought was "libbpf-cargo shouldn't bomb this... fix will be ugly, but it shouldn't bomb it."

libbpf-cargo/src/test.rs
mimullin-bbry: Didn't think to try.

I think it would be best to not have the __unnamed_.  I shall fix later on today after work and have a review for tomorrow.

mimullin-bbry: Addressed in https://github.com/libbpf/libbpf-rs/pull/174.

Sorry this is later than I claimed.

Issue #163, opened 2021-12-29
Non-deterministic output starting in libbpf-cargo 0.8+


jyn514:
I tried the following:
```
$ cargo libbpf --version
cargo-libbpf-libbpf 0.7.2
$ cargo libbpf make
$ git add src
$ cargo libbpf make
$ git diff --stat # no output

$ cargo install libbpf-cargo
$ cargo libbpf --version
cargo-libbpf-libbpf 0.8.0
$ cargo libbpf make
$ git add src
$ cargo libbpf make
$ git diff
diff --git a/src/my_file.skel.rs b/src/my_file.skel.rs
index 9bd1e5d..43e3718 100644
--- a/src/my_file.skel.rs
+++ b/src/my_file.skel.rs
@@ -487,7 +487,7 @@ const DATA: &[u8] = &[
     0, 0, 0, 0, 0, 0, 208, 1, 0, 0, 0, 0, 0, 0, 64, 1, 0, 0, 4, 0, 155, 0, 0, 0, 8, 1, 1, 251, 14,
     13, 0, 1, 1, 1, 1, 0, 0, 0, 1, 0, 0, 1, 115, 114, 99, 47, 98, 112, 102, 0, 47, 117, 115, 114,
     47, 105, 110, 99, 108, 117, 100, 101, 47, 97, 115, 109, 45, 103, 101, 110, 101, 114, 105, 99,
-    0, 47, 116, 109, 112, 47, 46, 116, 109, 112, 70, 80, 118, 116, 74, 84, 47, 98, 112, 102, 0, 47,
+    0, 47, 116, 109, 112, 47, 46, 116, 109, 112, 57, 98, 80, 117, 48, 105, 47, 98, 112, 102, 0, 47,
     117, 115, 114, 47, 105, 110, 99, 108, 117, 100, 101, 47, 108, 105, 110, 117, 120, 0, 0, 117,
     100, 112, 95, 114, 111, 117, 116, 101, 46, 98, 112, 102, 46, 99, 0, 1, 0, 0, 105, 110, 116, 45,
     108, 108, 54, 52, 46, 104, 0, 2, 0, 0, 98, 112, 102, 95, 104, 101, 108, 112, 101, 114, 95, 100,
```

This makes it annoying to commit the generated file to git, since it will always be marked as out of date.

jyn514:
I've bisected this to 46ca228836ae9b4b2c8bd61a48ea58f0a303690f.

[...]

PR #168, opened 2022-01-03
libbpf-cargo: Make generated object files deterministic by not using a temporary directory for include headers

jyn514:
Clang includes the name of the directory where includes are stored in the `.debug_line` section of
the headers when debug info is enabled. Since the directory was randomly created on each new run,
that meant the object file varied randomly each time it was created.

This changes the directory name to be consistently in a subdirectory of `target/` instead.

Fixes https://github.com/libbpf/libbpf-rs/issues/163. Closes https://github.com/libbpf/libbpf-rs/pull/169.

Signed-off-by: Joshua Nelson <jnelson@cloudflare.com>

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -6,7 +6,7 @@ use std::process::Command;
 use anyhow::{anyhow, bail, Context, Result};
 use regex::Regex;
 use semver::Version;
-use tempfile::TempDir;
+use tempfile::tempdir;
 
 use crate::metadata;
 use crate::metadata::UnprocessedObj;
@@ -46,12 +46,12 @@ fn extract_version(output: &str) -> Result<&str> {
 ///
 /// Directory and enclosed contents will be removed when return object is dropped.
 #[cfg(not(feature = "novendor"))]
-fn extract_libbpf_headers_to_disk() -> Result<Option<TempDir>> {
+fn extract_libbpf_headers_to_disk(target_dir: &Path) -> Result<Option<PathBuf>> {
     use std::fs::OpenOptions;
     use std::io::Write;
 
-    let tempdir = TempDir::new()?;
-    let dir = tempdir.path().join("bpf");
+    let parent_dir = target_dir.join("bpf").join("src");
+    let dir = parent_dir.join("bpf");
     fs::create_dir_all(&dir)?;
     for (filename, contents) in libbpf_sys::API_HEADERS.iter() {
         let path = dir.as_path().join(filename);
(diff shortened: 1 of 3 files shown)


jyn514:
Hmm, this makes it deterministic, but only on the same machine, if you change the name of the git directory or a different user clones and compiles the project, it will still change the generated output. I wonder if it makes more sense to omit debuginfo altogether ...

mimullin-bbry (approved):
+1

jyn514:
>  I wonder if it makes more sense to omit debuginfo altogether ...

I tried doing this and got a segfault (!!), I'm not sure what's going on there.

```diff
diff --git a/libbpf-cargo/src/build.rs b/libbpf-cargo/src/build.rs
index 2c55a8a..ecfaccf 100644
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -125,8 +125,8 @@ fn compile_one(debug: bool, source: &Path, out: &Path, clang: &Path, options: &s
     if !options.is_empty() {
         cmd.args(options.split_whitespace());
     }
-    cmd.arg("-g")
-        .arg("-O2")
+    // cmd.arg("-g")
+        cmd.arg("-O2")
         .arg("-target")
         .arg("bpf")
         .arg("-c")
```

```
Compiling BPF objects
Generating skeletons
libbpf: BTF is required, but is missing or corrupted.
Segmentation fault
```

jyn514:
Ok, https://github.com/libbpf/libbpf-rs/pull/169 fixes this in more cases.

insearchoflosttime:
Thanks for your work on this, @jyn514. I personally like this fix better than [the one using `-ffile-prefix-map`](https://github.com/libbpf/libbpf-rs/pull/169); if we have a use case where we need to use the debug info in the future, this fix is more accommodating to that whereas using -ffile-prefix-map doesn't link to where the actual path of the headers is (and please correct me if this understanding is incorrect). What are your thoughts?

jyn514:
Hmm, I'm not sure how you would be able to get debug info out of the generated .o file - it's only used in the `.skel.rs` file, right? I'm having a hard time imagining how you would debug this in the Rust binary ... note that the debug info in the .o is completely unrelated to whether the rust code has debug symbols or not.

My only reluctance about using this version is that the original motivation in https://github.com/libbpf/libbpf-rs/issues/163 was to check the generated file into git, and this means that regenerating it on different machines will still give you different versions. But on the other hand that was true before 1.8.0 too and no one complained. So I'm ok with merging this one.

jyn514:
@insearchoflosttime I notice you haven't had time to add support for static linking yet - that's totally understandable, but would you be willing to make a release with this change in the meantime? That would let me update to a version of libbpf-sys with https://github.com/libbpf/libbpf-sys/pull/22.

insearchoflosttime:
@jyn514 The libbpf-rs release version was just updated and now includes the latest changes

jyn514:
@insearchoflosttime thank you so much!

PR #170, opened 2022-01-04
Allow local BPF programs to share one map FD without pinning it globally. 

thesw4rm:
Extracted part of reuse_pinned_maps to a new function "reuse_fd".


thesw4rm:
Solves #171

libbpf-rs/src/map.rs
```diff
@@ -61,15 +61,11 @@ impl OpenMap {
...
-        }
-
+    /**
+    * Reuse a map created within the same userspace process.
+    * To use maps across userspace processes, pin the map first and use "reuse_pinned_map".
+    */
+    pub fn reuse_fd(&self, fd: i32) -> Result<()> {
         let ret = unsafe { libbpf_sys::bpf_map__reuse_fd(self.ptr, fd) };
 
         // Always close `fd` regardless of if `bpf_map__reuse_fd` succeeded or failed
```
insearchoflosttime, line 71: I don't think we should be closing the file descriptor here. In the case of `reuse_pinned_map`, it made more sense to close the fd since we were opening the fd first with `bpf_obj_get`.

DevasiaThomas: Makes sense, since the ``fd`` in such a scenario should be closed by first owner or when it gets dropped. Nice catch @insearchoflosttime .
@thesw4rm thoughts/reasons otherwise?

thesw4rm: > I don't think we should be closing the file descriptor (on line 78) for the standalone `reuse_fd`. In the case of `reuse_pinned_map`, it made more sense to close the fd since we opened the fd first with `bpf_obj_get`.

Good point. I'll change the line to only close it within `reuse_pinned_map`. Why are we able to close the fd for the pinned map and still use that fd within the BPF program?

insearchoflosttime: > Why are we able to close the fd for the pinned map and still use that fd within the BPF program?

For the pinned map case, the fd isn't exposed outside the `reuse_pinned_map` function whereas in the non-pinned map case, the fd is able to be used outside the function.

thesw4rm: > > Why are we able to close the fd for the pinned map and still use that fd within the BPF program?
> 
> For the pinned map case, the fd isn't exposed outside the `reuse_pinned_map` function whereas in the non-pinned map case, the fd is able to be used outside the function.

Then how exactly does the userspace/BPF program use the pinned map if the fd is already closed/not available outside the function? Does userspace open a new stream to the map after reuse_pinned_map is called?

anakryiko: libbpf's bpf_map__reuse_fd() is doing dup3() internally and creates a new FD out of the one provided. So it's ok to close the original FD that was opened from BPF FS. When reusing map by FD directly, we don't own/control original FD, so we shouldn't close it.

thesw4rm: That makes sense thank you!

libbpf-rs/src/map.rs
```diff
@@ -61,6 +61,20 @@ impl OpenMap {
         unsafe { libbpf_sys::bpf_map__set_inner_map_fd(self.ptr, inner.fd()) };
     }
 
+    /**
+     * Reuse a map created within the same userspace process.
```
anakryiko, line 65: it doesn't have to be created within the userspace process, why these unnecessary and misleading details? Just something like "reuse provided BPF map FD instead of creating a new map" or something along those line would be more accurate.

You can get BPF map FD in many ways even if map wasn't created by your own process (e.g., look up by map ID, open from BPF FS, or even transfer FD over unix domain socket).

libbpf-rs/src/map.rs
```diff
@@ -61,6 +61,20 @@ impl OpenMap {
...
+     * To use maps across userspace processes, pin the map first and use "reuse_pinned_map".
```
anakryiko, line 66: and again, it's one of the ways, and it has nothing to do with `reuse_fd` API. I'm not sure doc comment on reuse_fd is the right place to explain all possible ways to share maps between processes, tbh.

libbpf-rs/src/map.rs
thesw4rm: I was trying to define the difference between `reuse_fd` and `reuse_pinned_map` in the comments to justify why I split some code from `reuse_pinned_map` into a new function.  What do you think would be the best way to do that, or is that not needed at all?

anakryiko: I think the name is pretty clear, one reuses by map FD, another reuses map pinned in BPF FS. If someone doesn't know about pinning, they'll have to learn concepts before being able to use this API, I don't think there is a short cut.

thesw4rm: Sounds good. I'll remove that doc comment then.

anakryiko (approved):
please squash all the commits into one and add your 

Signed-off-by: First Last <you@email.com>

thesw4rm:
> please squash all the commits into one and add your
> 
> Signed-off-by: First Last [you@email.com](mailto:you@email.com)

Squashing the commits I can do. Is the signed off part referring to GPG verification of a commit, or specifically put that in the commit message?

thesw4rm:
> > please squash all the commits into one and add your
> > Signed-off-by: First Last [you@email.com](mailto:you@email.com)
> 
> Squashing the commits I can do. Is the signed off part referring to GPG verification of a commit, or specifically put that in the commit message?

I did a force push to my repo by mistake lol. I'll make a completely new PR with all the changes in one commit.

PR #174, opened 2022-01-07
libbpf-cargo: __unnamed_ is unnecessary as a field prefix

mimullin-bbry:
Rust can have a field name be the same as the type.  __unnamed_ is
not necessary for unnamed unions, and harms readability


insearchoflosttime (approved):
LGTM

PR #178, opened 2022-01-13
Minor fixe and addition

arilou:
* Add get_section_name function for Program
* Fix pages method for PerfBuffer

### DIFF
--- a/libbpf-rs/src/perf_buffer.rs
+++ b/libbpf-rs/src/perf_buffer.rs
@@ -71,9 +71,13 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
     }
 
     /// The number of pages to size the ring buffer.
-    pub fn pages(&mut self, pages: usize) -> &mut Self {
-        self.pages = pages;
-        self
+    pub fn pages(self, pages: usize) -> PerfBufferBuilder<'a, 'b> {
+        PerfBufferBuilder {
+            map: self.map,
+            pages,
+            sample_cb: self.sample_cb,
+            lost_cb: self.lost_cb,
+        }
     }
 
     pub fn build(self) -> Result<PerfBuffer<'b>> {
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -12,11 +12,12 @@ use crate::*;
 /// This object exposes operations that need to happen before the program is loaded.
 pub struct OpenProgram {
     ptr: *mut libbpf_sys::bpf_program,
+    section: String,
 }
 
 impl OpenProgram {
-    pub(crate) fn new(ptr: *mut libbpf_sys::bpf_program) -> Self {
-        OpenProgram { ptr }
+    pub(crate) fn new(ptr: *mut libbpf_sys::bpf_program, section: String) -> Self {
+        Self { ptr, section }
     }
 
     pub fn set_prog_type(&mut self, prog_type: ProgramType) {
@@ -35,6 +36,11 @@ impl OpenProgram {
         unsafe {
             libbpf_sys::bpf_program__set_ifindex(self.ptr, idx);
         }
+    }
+
+    /// Name of the section this `Program` belongs to.
+    pub fn section(&self) -> &str {
+        &self.section
     }
 }
 
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -782,6 +782,8 @@ fn gen_skel_contents(_debug: bool, raw_obj_name: &str, obj_file_path: &Path) -> 
         r#"
         }}
 
+        unsafe impl<'a> Send for {name}Skel<'a> {{}}
+
         impl<'a> {name}Skel<'a> {{
         "#,
         name = &obj_name,
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -168,8 +168,13 @@ impl OpenObject {
             let name = unsafe { libbpf_sys::bpf_program__name(next_ptr) };
             let name = util::c_ptr_to_string(name)?;
 
+            // Get the program section
+            // bpf_program__section_name never returns NULL, so no need to check the pointer.
+            let section = unsafe { libbpf_sys::bpf_program__section_name(next_ptr) };
+            let section = util::c_ptr_to_string(section)?;
+
             // Add the program to the hashmap
-            obj.progs.insert(name, OpenProgram::new(next_ptr));
+            obj.progs.insert(name, OpenProgram::new(next_ptr, section));
             prog = next_ptr;
         }
 


libbpf-rs/src/perf_buffer.rs
```diff
@@ -71,9 +71,13 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
...
-    pub fn pages(&mut self, pages: usize) -> &mut Self {
-        self.pages = pages;
-        self
+    pub fn pages(self, pages: usize) -> PerfBufferBuilder<'a, 'b> {
+        PerfBufferBuilder {
+            map: self.map,
+            pages,
+            sample_cb: self.sample_cb,
+            lost_cb: self.lost_cb,
+        }
```
insearchoflosttime, line 80: I don't think this is quite right - to clarify, what is the intention with your proposed API changes to pages? The original version of pages looks fine to me, someone can set the pages of the perf buffer through calling something like:

```
let perf = 
   PerfBufferBuilder::new(skel.maps_mut().events()).pages(5)
   .sample_cb(handle_event).lost_cb(handle_lost_events).build()?;
```

libbpf-rs/src/program.rs
```diff
@@ -36,6 +36,12 @@ impl OpenProgram {
             libbpf_sys::bpf_program__set_ifindex(self.ptr, idx);
         }
     }
+
+    /// Get program section name
+    pub fn get_section_name(&self) -> Result<String> {
+        let section = unsafe { libbpf_sys::bpf_program__section_name(self.ptr) };
+        util::c_ptr_to_string(section)
+    }
 }
```
insearchoflosttime, line 45: This LGTM. Thanks for adding this

libbpf-rs/src/perf_buffer.rs
arilou: Without my fix your line will produces this error for me:
   Compiling libbpf-rs v0.15.0 (/.../libbpf-rs/libbpf-rs)
   Compiling rust_bpf_agent v0.1.0 (/.../rust_bpf_agent)
error[E0507]: cannot move out of a mutable reference
   --> src/main.rs:116:16
    |
116 |       let perf = PerfBufferBuilder::new(skel.maps_mut().events())
    |  ________________^
117 | |         .sample_cb(handle_event)
118 | |         .lost_cb(handle_lost_events)
119 | |         .pages(1024)
120 | |         .build()?;
    | |________________^ move occurs because value has type `PerfBufferBuilder<'_, '_>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0507`.
error: could not compile `rust_bpf_agent` due to previous error

arilou: While we are talking about this, is there a way I can pass PerfBuffer loop to a different thread? (Currently it does not implement Send)
Think of a case where the eBPF programs sends A LOT of messages and you want to keep fetching events and put it in some channel to do further more cpu intensive processing, but you dont want the ebpf map to get full...

anakryiko: Looking at lost_cb definition it differs from what is done in pages, so the change looks in line with other setters. I'm surprised this was never caught before, don't we have some example or test that uses PerfBuffer? Can you please add one?

anakryiko: > While we are talking about this, is there a way I can pass PerfBuffer loop to a different thread? (Currently it does not implement Send) Think of a case where the eBPF programs sends A LOT of messages and you want to keep fetching events and put it in some channel to do further more cpu intensive processing, but you dont want the ebpf map to get full...

Not clear what exactly you are sending in this case, but keep in mind that as it is designed in libbpf right now, PerfBuffer is single-consumer and poll has to be called from single thread. Also keep in mind that each per-CPU buffer has to be single-consumer. The only wiggle room is to do poll on different per-cpu buffers on different threads.

But please start a separate issue to discuss this use case.

libbpf-rs/src/program.rs
anakryiko: let's not use "get_" prefix for getters. Libbpf follows this rule as well.

arilou: Fixed in next commit

libbpf-rs/src/perf_buffer.rs
arilou: I dont think it's such a big deal, one could always wrap *Skel in it's own struct and implement Send for it
But in general I dont understand why the *Skel cannot be Send, unless I'm missing something

libbpf-rs/src/program.rs
```diff
@@ -36,6 +36,12 @@ impl OpenProgram {
             libbpf_sys::bpf_program__set_ifindex(self.ptr, idx);
         }
     }
+
+    /// Get program section name
+    pub fn section_name(&self) -> Result<String> {
```
anakryiko, line 41: hm... just looked at Program and it seems like it has similar accessor, but it's called just `section` and it returns `&str` (because it creates internal copy). We should probably keep them consistent, no?

libbpf-rs/src/perf_buffer.rs
anakryiko: Probably no one ever needed it. I'm not too familiar with Sync and Send, but if there are no issues with the way PerfBuffer is implemented, then yeah, we should implement those traits.

libbpf-rs/src/program.rs
arilou: Do you mean the name should be the same? if so then sure, I'll make the change, as for the copy, if you will refer to how it work with Program, it's actually a member in the struct that got there when it was constructed
```
            // Get the program section
            // bpf_program__section_name never returns NULL, so no need to check the pointer.
            let section = unsafe { libbpf_sys::bpf_program__section_name(next_ptr) };
            let section = util::c_ptr_to_string(section)?;

            // Get the program fd
            let fd = unsafe { libbpf_sys::bpf_program__fd(next_ptr) };
            if fd < 0 {
            }

            // Add the program to the hashmap
            obj.progs
                .insert(name.clone(), Program::new(next_ptr, name, section));
            prog = next_ptr;
```

arilou:
The reason for pages and adding the Send (not the Sync, for Sync you need some Mutex or something to help you) is if you will make a program that sends A LOT of events for instance every system call in the system. so in order to catch up and not fill the per-core PerfBuf ring buffer you need another thread to "drain" and not spend time working on the events.

libbpf-rs/src/program.rs
anakryiko: I meant both the name and return type. OpenProgram could do the same thing as Program, right? Are there any downsides to that?

anakryiko:
> The reason for pages and adding the Send (not the Sync, for Sync you need some Mutex or something to help you) is if you will make a program that sends A LOT of events for instance every system call in the system. so in order to catch up and not fill the per-core PerfBuf ring buffer you need another thread to "drain" and not spend time working on the events.

Yeah, I didn't mean to imply that we need to implement Sync. It's just that Send and Sync are usually described together it seems. From https://stackoverflow.com/a/60109068 it seems like Send is a totally reasonable thing for PerfBuffer (and RingBuffer, while we are at it) to support. So feel free to send a PR with the required changes.

libbpf-rs/src/program.rs
arilou: Done

arilou:
Regarding Send, I think it should happen in the libbpf-cargo on the *Skel struct created after load
so the generator will also add
unsafe impl<'a> Send for FooSkel<'a> {}
Do you agree?

anakryiko:
> Regarding Send, I think it should happen in the libbpf-cargo on the *Skel struct created after load so the generator will also add unsafe impl<'a> Send for FooSkel<'a> {} Do you agree?

Yep, LGTM.

Please squash your commits so that we have only three of them (skel Send, section_name, pages).

arilou:
Done, separated to 3 patches

anakryiko (approved):
LGTM, thanks.

PR #180, opened 2022-01-24
libbpf-rs: Fix clippy error in examples/query

insearchoflosttime:

### DIFF
--- a/examples/query/src/main.rs
+++ b/examples/query/src/main.rs
@@ -31,7 +31,7 @@ fn prog() {
 
 fn map() {
     for map in query::MapInfoIter::default() {
-        println!("name={:<16} type={}", map.name, map.ty.to_string(),);
+        println!("name={:<16} type={}", map.name, map.ty);
     }
 }
 


examples/query/src/main.rs
```diff
@@ -31,7 +31,7 @@ fn prog() {
 
 fn map() {
     for map in query::MapInfoIter::default() {
-        println!("name={:<16} type={}", map.name, map.ty.to_string(),);
+        println!("name={:<16} type={}", map.name, map.ty,);
```
anakryiko, line 34: is that trailing coma necessary? looks weird

anakryiko (approved):

examples/query/src/main.rs
insearchoflosttime: Rust supports trailing commas but I agree it looks unnatural. I will remove this.

insearchoflosttime:
This check is still failing due to clippy complaining about "error: this implementation is unsound, as some fields in `RunqslowerSkel<'a>` are `!Send`" but this warning is misguided and fixed in Rust 1.58 as per https://github.com/rust-lang/rust-clippy/issues/8045. We'll need to update the build to use the latest version of Rust.

PR #184, opened 2022-01-26
perf_buffer: Add epoll_fd, consume, consume_buffer, ... wrappers

insearchoflosttime:
This patch adds epoll_fd, consume, consume_buffer, buffer_cnt, and
buffer_fd support for perf buffers.

This is part of a larger effort to reach feature-parity with libbpf.

### DIFF
--- a/libbpf-rs/src/perf_buffer.rs
+++ b/libbpf-rs/src/perf_buffer.rs
@@ -157,9 +157,35 @@ pub struct PerfBuffer<'b> {
 }
 
 impl<'b> PerfBuffer<'b> {
+    pub fn epoll_fd(&self) -> i32 {
+        unsafe { libbpf_sys::perf_buffer__epoll_fd(self.ptr) }
+    }
+
     pub fn poll(&self, timeout: Duration) -> Result<()> {
         let ret = unsafe { libbpf_sys::perf_buffer__poll(self.ptr, timeout.as_millis() as i32) };
         util::parse_ret(ret)
+    }
+
+    pub fn consume(&self) -> Result<()> {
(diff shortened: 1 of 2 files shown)


masmullin2000:
LGTM

masmullin2000:
Funnily enough, I was just thinking about the epoll FD, and wondering if it could be used to expose libbpf objects into an async scenario.

I was thinking that significant work would need to be done to create a Future object for the PerfBuffer (ie probably a new struct called AsyncPerfBuffer), and then perf_buffer__poll() would need to be replaced as it comes from the libbpf-C project and is a uses epoll_wait() which is blocking (and thus not suitable for async).

Any thoughts about adding async functionality to perf buffers and ring buffers?

mimullin-bbry:
LGTM

mimullin-bbry:
Funnily enough, I was just thinking about the epoll FD, and wondering if it could be used to expose libbpf objects into an async scenario.

I was thinking that significant work would need to be done to create a Future object for the PerfBuffer (ie probably a new struct called AsyncPerfBuffer), and then perf_buffer__poll() would need to be replaced as it comes from the libbpf-C project and is a uses epoll_wait() which is blocking (and thus not suitable for async).

Any thoughts about adding async functionality to perf buffers and ring buffers?

mimullin-bbry (approved):

insearchoflosttime:
I think for adding async functionality to perf buffers and ring buffers, we'd likely need to implement our own libbpf-rs async version of these instead of using libbpf's since libbpf's uses blocking calls

insearchoflosttime:
> @insearchoflosttime insearchoflosttime force-pushed the perf-buffer-functions branch from a7761c0 to f61252a now

Rebase onto origin/master

PR #192, opened 2022-02-26
libbpf-rs: Add set_autoload/set_attach_target to OpenProgram

### LINKED ISSUE #175: Add support for bpf_program__set_autoload
The libbpf API `bpf_program__set_autoload` can be used to disable loading and verifying specific BPF programs. This API should be added to libbpf-rs.

chenhengqi:
Add bpf_program__set_autoload / bpf_program__set_attach_target equivalent API to libbpf-rs.

Closes #175 and #176.

Signed-off-by: Hengqi Chen <chenhengqi@outlook.com>

### DIFF
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -360,12 +360,6 @@ impl Object {
             let section = unsafe { libbpf_sys::bpf_program__section_name(next_ptr) };
             let section = util::c_ptr_to_string(section)?;
 
-            // Get the program fd
-            let fd = unsafe { libbpf_sys::bpf_program__fd(next_ptr) };
-            if fd < 0 {
-                return Err(Error::System(-fd));
-            }
-
             // Add the program to the hashmap
             obj.progs
                 .insert(name.clone(), Program::new(next_ptr, name, section));
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -1,5 +1,6 @@
 use std::convert::TryFrom;
 use std::path::Path;
+use std::ptr;
 
 use num_enum::TryFromPrimitive;
 use strum_macros::Display;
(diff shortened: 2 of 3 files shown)


libbpf-rs/src/object.rs
```diff
@@ -361,9 +361,11 @@ impl Object {
...
 
             // Get the program fd
-            let fd = unsafe { libbpf_sys::bpf_program__fd(next_ptr) };
-            if fd < 0 {
-                return Err(Error::System(-fd));
+            if unsafe { libbpf_sys::bpf_program__autoload(next_ptr) } {
+                let fd = unsafe { libbpf_sys::bpf_program__fd(next_ptr) };
+                if fd < 0 {
+                    return Err(Error::System(-fd));
+                }
```
insearchoflosttime, line 368: Looking at this some more, I think we can just remove the `libbpf_sys::bpf_program__fd` call altogether - we don't do anything with the program fd. WDYT?

libbpf-rs/src/program.rs
```diff
@@ -46,6 +47,29 @@ impl OpenProgram {
...
     }
+
+    pub fn set_attach_target(
+        &mut self,
+        attach_prog_fd: i32,
+        attach_func_name: Option<String>,
+    ) -> Result<()> {
+        let ret = if let Some(name) = attach_func_name {
+            // NB: we must hold onto a CString otherwise our pointer dangles
+            let name_c = util::str_to_cstring(&name)?;
```
insearchoflosttime, line 58: I don't think we need to create a CString for this. We don't need to persist the string inside `bpf_program__set_attach_target`

insearchoflosttime (changes requested):

libbpf-rs/src/object.rs
chenhengqi: Yes, that is what I had in mind. Let's drop it then.

libbpf-rs/src/program.rs
chenhengqi: We have to hold this `CString` so that `as_ptr` remain valid.

https://github.com/libbpf/libbpf-rs/blob/42b84c0e4a22d8e398649fe8e787080e8340a2db/libbpf-rs/src/object.rs#L67-L74

insearchoflosttime: > We have to hold this CString so that as_ptr remain valid.

Can you elaborate on this ? Inside `bpf_program__set_attach_target`, the name is only used for strcmping and snprintfing so it looks like this pointer does not need to be persisted anywhere. Why do we need to hold onto the CString and what would cause the pointer dangle?

chenhengqi: The pointer remains valid as long as you hold the CString.

Please refer to https://doc.rust-lang.org/std/ffi/struct.CString.html#examples-1

chenhengqi:
```rust
let ptr = CString::new("Hello").expect("CString::new failed").as_ptr();
// the temporary CString destruct at here, ptr points to freed memory
```

insearchoflosttime: Interesting, that seems to directly contradict their documentation about the Cstring being deallocated immediately after the `CString::new("Hello").expect("CString::new failed").as_ptr()` expression is evaluated.

I tested it locally and I verified that the pointer does remain valid, so you're right :) I'll merge the changes in.

PR #202, opened 2022-03-09
add getter and setter for bpf program flags

### LINKED ISSUE #198: Get/set bpf program flags
Implement the ability to get and set program flags. This should be a wrapper around the libbpf library's

`__u32 bpf_program__flags(const struct bpf_program *prog);`

and

`int bpf_program__set_flags(struct bpf_program *prog, __u32 flags);`

(https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L8538-L8550). 

The libbpf-rs code for interacting with programs is located at libbpf-rs/src/program.rs

kckeiks:
Fixes #198

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -70,6 +70,11 @@ impl OpenProgram {
         };
         util::parse_ret(ret)
     }
+
+    pub fn set_flags(&self, flags: u32) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_program__set_flags(self.ptr, flags) };
+        util::parse_ret(ret)
+    }
 }
 
 /// Type of a [`Program`]. Maps to `enum bpf_prog_type` in kernel uapi.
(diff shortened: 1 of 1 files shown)


libbpf-rs/src/program.rs
```diff
@@ -70,6 +70,11 @@ impl OpenProgram {
         };
         util::parse_ret(ret)
     }
+
+    pub fn set_prog_flags(&self, flags: u32) -> Result<()> {
```
chenhengqi, line 74: Should be `set_flags` ?

kckeiks: There is a method `set_prog_type` so I was following that pattern. I can change it though.

anakryiko: good point, shouldn't prog_type and set_prog_type be just `type` and `set_type` to match libbpf's API? Though I suspect "type" is reserved keyword in Rust?

chenhengqi: Yep, `type` is a keyword.

insearchoflosttime (approved):

insearchoflosttime:
Thanks for adding this!

PR #203, opened 2022-03-10
add bpf_insn wrapper to read program instructions

### LINKED ISSUE #200: Get program's bpf instructions + instruction count
Implement the ability to get the program's bpf instructions + instruction count. This should be a wrapper around the libbpf library's

```
/**
 * @brief **bpf_program__insns()** gives read-only access to BPF program's
 * underlying BPF instructions.
 * @param prog BPF program for which to return instructions
 * @return a pointer to an array of BPF instructions that belong to the
 * specified BPF program
 *
 * Returned pointer is always valid and not NULL. Number of `struct bpf_insn`
 * pointed to can be fetched using **bpf_program__insn_cnt()** API.
 *
 * Keep in mind, libbpf can modify and append/delete BPF program's
 * instructions as it processes BPF object file and prepares everything for
 * uploading into the kernel. So depending on the point in BPF object
 * lifetime, **bpf_program__insns()** can return different sets of
 * instructions. As an example, during BPF object load phase BPF program
 * instructions will be CO-RE-relocated, BPF subprograms instructions will be
 * appended, ldimm64 instructions will have FDs embedded, etc. So instructions
 * returned before **bpf_object__load()** and after it might be quite
 * different.
 */
LIBBPF_API const struct bpf_insn *bpf_program__insns(const struct bpf_program *prog);
/**
 * @brief **bpf_program__insn_cnt()** returns number of `struct bpf_insn`'s
 * that form specified BPF program.
 * @param prog BPF program for which to return number of BPF instructions
 *
 * See **bpf_program__insns()** documentation for notes on how libbpf can
 * change instructions and their count during different phases of
 * **bpf_object** lifetime.
 */
LIBBPF_API size_t bpf_program__insn_cnt(const struct bpf_program *prog);
```

(https://github.com/libbpf/libbpf/blob/3591deb9bc6b0848721f831790b78ae0a593b4c5/src/libbpf.h#L305-L335)

kckeiks:
add wrapper around `bpf_program__insn_cnt`

Fixes #200

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -75,6 +75,16 @@ impl OpenProgram {
         let ret = unsafe { libbpf_sys::bpf_program__set_flags(self.ptr, flags) };
         util::parse_ret(ret)
     }
+
+    /// Returns the number of instructions that form the program.
+    ///
+    /// Note: Keep in mind, libbpf can modify the program's instructions
+    /// and consequently its instruction count, as it processes the BPF object file.
+    /// So [`OpenProgram::insn_cnt`] and [`Program::insn_cnt`] may return different values.
+    ///
+    pub fn insn_cnt(&self) -> usize {
+        unsafe { libbpf_sys::bpf_program__insn_cnt(self.ptr) as usize }
+    }
 }
 
 /// Type of a [`Program`]. Maps to `enum bpf_prog_type` in kernel uapi.
@@ -400,5 +410,13 @@ impl Program {
         } else {
             Ok(Link::new(ptr))
         }
+    }
+
+    /// Returns the number of instructions that form the program.
+    ///
+    /// Please see note in [`OpenProgram::insn_cnt`].
+    ///
+    pub fn insn_cnt(&self) -> usize {
+        unsafe { libbpf_sys::bpf_program__insn_cnt(self.ptr) as usize }
     }
 }


libbpf-rs/src/insn.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/// Represents the instructions of a program.
+///
+/// Note: Keep in mind, libbpf can modify the program's instructions
+/// as it processes the BPF object file. So depending on the point in
+/// the BPF object's lifetime, `insns()` in [`OpenProgram`] and [`Program`] can
+/// return different sets of instructions.
+///
+pub struct Insns {
```
kckeiks, line 19: Not sure I like that there is a `Insns` and a `Insn` ...

libbpf-rs/src/program.rs
```diff
@@ -391,4 +399,11 @@ impl Program {
             Ok(Link::new(ptr))
         }
     }
+
+    /// Returns program's underlying BPF instructions.
+    ///
+    /// Please see note in [`Insns`].
+    pub fn insns(&self) -> Insns {
+        Insns::new(self.ptr)
```
insearchoflosttime, line 407: Interesting! Is there some reason we can't just return a Vec<Insn> to the user?

kckeiks: That's what I am not sure about. It could return a Vec and then we could remove `Insns` and ignore the pointer to `bpf_insn` array but according to this [patch](https://lore.kernel.org/bpf/20211025224531.1088894-3-andrii@kernel.org/), the pointer array returned by `bpf_program__insns` can be used for [cloning](https://github.com/anakryiko/retsnoop/pull/1) so, to support this use-case, we'd need `Insns` to hold our pointer to the `bpf_insn` array. Alternatively, I guess we could add the ability to convert the Vec into a C array and use that for cloning?

kckeiks: I pushed commit [b569bfe](https://github.com/libbpf/libbpf-rs/pull/203/commits/b569bfe847bfa6b9edede071428cd9876c46e155) which returns a Vec as you suggested. Let me know how that one looks. I'll squash the commits once we move forward.

anakryiko:
Given we currently have no API to disassemble raw bpf_insn into human-readable BPF assembly. And given that we can't just pass Vec[Insns] into bpf_prog_create() API (like we can with C), it seems a bit premature to add this as an API. There isn't much you'll be able to do with those instructions in Rust anyways. It's just numbers, which humans can't do much with.

So maybe let's hold off on adding these APIs? insn_cnt() might be a good idea to capture the size of the program code, but getting instructions themselves isn't very practical (at least right now).

kckeiks:
> Given we currently have no API to disassemble raw bpf_insn into human-readable BPF assembly. And given that we can't just pass Vec[Insns] into bpf_prog_create() API (like we can with C), it seems a bit premature to add this as an API. There isn't much you'll be able to do with those instructions in Rust anyways. It's just numbers, which humans can't do much with.
> 
> So maybe let's hold off on adding these APIs? insn_cnt() might be a good idea to capture the size of the program code, but getting instructions themselves isn't very practical (at least right now).

Sounds good. I will remove `insn()` and keep `insn_cnt()`.

insearchoflosttime (approved):

insearchoflosttime:
Thanks for working on this @kckeiks!

PR #211, opened 2022-03-19
Add test for BTF_KIND_FLOAT

chenhengqi:
Add a test case for BTF_KIND_FLOAT.

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -731,7 +731,7 @@ impl<'a> Btf<'a> {
     fn load_type(&mut self, data: &'a [u8]) -> Result<BtfType<'a>> {
         let t = data.pread::<btf_type>(0)?;
         let extra = &data[size_of::<btf_type>()..];
-        let kind = (t.info >> 24) & 0xf;
+        let kind = (t.info >> 24) & 0x1f;
 
         match BtfKind::try_from(kind)? {
             BtfKind::Void => {
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -2145,3 +2145,25 @@ pub struct __anon_3 {
 
     assert_definition(&btf, struct_bpf_sock_tuple, expected_output);
 }
+
+#[test]
+fn test_btf_dump_float() {
+    let prog_text = r#"
+float f = 2.16;
+double d = 12.15;
+"#;
+
+    let btf = build_btf_prog(prog_text);
+
+    let f = find_type_in_btf!(btf, Var, "f");
+    let d = find_type_in_btf!(btf, Var, "d");
+
+    assert_eq!(
+        "f32",
+        btf.type_declaration(f).expect("Failed to generate f decl")
+    );
+    assert_eq!(
+        "f64",
+        btf.type_declaration(d).expect("Failed to generate d decl")
+    );
+}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,8 +25,9 @@ jobs:
     - name: Build
       run: cargo build --verbose --workspace --exclude runqslower
     - name: Run tests
-      # Skip BTF tests which require sudo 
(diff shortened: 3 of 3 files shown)


libbpf-cargo/src/test.rs
```diff
@@ -2145,3 +2145,30 @@ pub struct __anon_3 {
 
     assert_definition(&btf, struct_bpf_sock_tuple, expected_output);
 }
+
+#[test]
+fn test_btf_dump_float() {
+    let prog_text = r#"
+#include "vmlinux.h"
```
anakryiko, line 2152: do you need vmlinux.h at all?

chenhengqi: No, just remove it.

libbpf-cargo/src/btf/btf.rs
```diff
@@ -731,7 +731,7 @@ impl<'a> Btf<'a> {
     fn load_type(&mut self, data: &'a [u8]) -> Result<BtfType<'a>> {
         let t = data.pread::<btf_type>(0)?;
         let extra = &data[size_of::<btf_type>()..];
-        let kind = (t.info >> 24) & 0xf;
```
anakryiko, line 734: please check if it's the only place where we extract kind from btf_type. It's probably best to move this into a separate function, just like libbpf does.

chenhengqi: OK, will follow up on a new PR.

Currently, the code already has a `get_kind()` method, but it is used to extract kind_flag actually.

insearchoflosttime:
I'm seeing this test failure in CI:

```
error: test failed, to rerun pass '-p libbpf-cargo --lib'
---- test::test_btf_dump_float stdout ----
Metadata for package=proj
	null
Found bpf progs to compile:
	UnprocessedObj { package: "proj", path: "/tmp/.tmpyV83Wg/proj/src/bpf/prog.bpf.c", out: "/tmp/.tmpyV83Wg/proj/target/bpf", name: "prog" }
Building /tmp/.tmpyV83Wg/proj/src/bpf/prog.bpf.c
thread 'test::test_btf_dump_float' panicked at 'assertion failed: `(left == right)`
  left: `"f32"`,
 right: `"std::ffi::c_void"`', libbpf-cargo/src/test.rs:2161:5
```

@chenhengqi  would you mind taking a look?

chenhengqi:
> I'm seeing this test failure in CI:
> 
> ```
> error: test failed, to rerun pass '-p libbpf-cargo --lib'
> ---- test::test_btf_dump_float stdout ----
> Metadata for package=proj
> 	null
> Found bpf progs to compile:
> 	UnprocessedObj { package: "proj", path: "/tmp/.tmpyV83Wg/proj/src/bpf/prog.bpf.c", out: "/tmp/.tmpyV83Wg/proj/target/bpf", name: "prog" }
> Building /tmp/.tmpyV83Wg/proj/src/bpf/prog.bpf.c
> thread 'test::test_btf_dump_float' panicked at 'assertion failed: `(left == right)`
>   left: `"f32"`,
>  right: `"std::ffi::c_void"`', libbpf-cargo/src/test.rs:2161:5
> ```
> 
> @chenhengqi would you mind taking a look?

The test passed in my dev machine.
And the CI logs shows that it does NOT triggered on the latest commit. Just force push to trigger a rerun.

insearchoflosttime:
I'll update the clang version CI uses to clang-13 (right now it's on clang-10). Some older clang versions emit void for float types.

insearchoflosttime:
The git workflow uses ubuntu-latest (ubuntu 20.04) which does not support clang-13 yet. We will have to put the test on hold until then. We can merge in the first commit (fixing the btf kind extraction) but I think it'll have to be resubmitted as a standalone PR without the test in order to be merged in

chenhengqi:
How about just ignore the test for now ?

insearchoflosttime:
Great, I just merged this in. Thanks for contributing to libbpf-rs @chenhengqi!

Issue #191, opened 2022-02-21
libbpf-rs: Add support for creating map standalone


zyy17:
## User Story

We need to create pinned map when some conditions meet, but libbpf-rs only support create map from ELF file.

## Example

libbpf support `bpf_map_create()` and I think we can create the new abstraction to support creating map standalone. Refer to [cilium/ebpf](https://github.com/cilium/ebpf) implementation, maybe we can provide the following API(same as [map.go](https://github.com/cilium/ebpf/blob/master/map.go) but just the demonstration):

```rust
pub struct Map {
    name: String,
    fd: i32,
    typ: libbpf_sys::bpf_map_type,
    key_size: u32,
    value_size: u32,
    max_entries: u32,
    flags: u32,
    pinned_path: String,
}

pub struct MapSpec {
    name: String,
    key_size: u32,
    value_size: u32,
    max_entries: u32,
    map_type: libbpf_sys::bpf_map_type,
    flags: u32,
    pinning: PinType,
}

impl Map {
    pub fn new_map_with_options(spec: MapSpec, opts: MapOptions) -> Result<Self> {  
        /* use create_map() of MapSpec */ 
    }
}

impl MapSpec {
    fn create_map(&self) -> Result<Map> {  /* use libbpf_sys::bpf_map_create() */ }
}
```

After map creating, user space could use `fd` to access map.

danobi:
Can you elaborate what your use case is? And why creating maps from elf file is insufficient?

zyy17:
In my scenario, we use the pinned map like the global configuration that shares between multiple BPF progs. 
For example:
1. Our daemon service will create pinned map after launching;
2. daemon service will accept the RPC request to decide which BPF prog should be loaded and what configuration should be set in the configuration pinned map; 
3. daemon can load multiple BPF progs and control their behavior by the global configuration map;

If we use ELF file(as we currently do), the daemon should always maintain some kind of ELF object file(or can hardcode in the code). I think it's not an elegant way to handle such situations, maybe the standalone API will be the better solution.

insearchoflosttime:
Hi @zyy17, this is definitely on the libbpf-rs roadmap as part of our larger goal to reach feature parity with libbpf 1.0. We will prioritize this. Thanks for your patience.

zyy17:
@insearchoflosttime Looking forward to 1.0 😊  If you agree the design of  map API of cilium/ebpf, maybe I can provide the PR(I already implement the internal version).

[...]

PR #212, opened 2022-04-03
libbpf-rs: add support for creating bpf map standalone

zyy17:
libbpf-rs: add support for creating bpf map standalone(for resolving issue #191 )

Signed-off-by: zyy17 <zyylsxm@gmail.com>

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -2,10 +2,11 @@ use core::ffi::c_void;
 use std::convert::TryFrom;
 use std::path::Path;
 use std::ptr;
+use std::ptr::null;
 
 use bitflags::bitflags;
 use nix::{errno, unistd};
-use num_enum::TryFromPrimitive;
+use num_enum::{IntoPrimitive, TryFromPrimitive};
 use strum_macros::Display;
 
 use crate::*;
@@ -85,6 +86,8 @@ pub struct Map {
     ty: libbpf_sys::bpf_map_type,
     key_size: u32,
     value_size: u32,
+
+    // The ptr will be null if we use Map::create to create the map from the userspace side directly.
     ptr: *mut libbpf_sys::bpf_map,
 }
 
@@ -153,18 +156,29 @@ impl Map {
         let path_c = util::path_to_cstring(path)?;
         let path_ptr = path_c.as_ptr();
 
-        let ret = unsafe { libbpf_sys::bpf_map__pin(self.ptr, path_ptr) };
+        let ret = if self.ptr.is_null() {
+            unsafe { libbpf_sys::bpf_obj_pin(self.fd, path_ptr) }
+        } else {
+            unsafe { libbpf_sys::bpf_map__pin(self.ptr, path_ptr) }
+        };
+
         util::parse_ret(ret)
     }
 
     /// [Unpin](https://facebookmicrosites.github.io/bpf/blog/2018/08/31/object-lifetime.html#bpffs)
     /// from bpffs
     pub fn unpin<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
-        let path_c = util::path_to_cstring(path)?;
-        let path_ptr = path_c.as_ptr();
-
-        let ret = unsafe { libbpf_sys::bpf_map__unpin(self.ptr, path_ptr) };
-        util::parse_ret(ret)
+        if self.ptr.is_null() {
+            match std::fs::remove_file(path) {
+                Ok(_) => Ok(()),
+                Err(e) => Err(Error::Internal(format!("remove pin map failed: {}", e))),
+            }
+        } else {
+            let path_c = util::path_to_cstring(path)?;
+            let path_ptr = path_c.as_ptr();
+            let ret = unsafe { libbpf_sys::bpf_map__unpin(self.ptr, path_ptr) };
+            util::parse_ret(ret)
+        }
     }
 
     /// Returns map value as `Vec` of `u8`.
@@ -409,6 +423,59 @@ impl Map {
     pub fn keys(&self) -> MapKeyIter {
         MapKeyIter::new(self, self.key_size())
     }
+
+    /// Create the bpf map standalone.
+    pub fn create<T: AsRef<str>>(
+        map_type: MapType,
+        name: Option<T>,
+        key_size: u32,
+        value_size: u32,
+        max_entries: u32,
+        opts: &libbpf_sys::bpf_map_create_opts,
+    ) -> Result<Map> {
+        let (map_name_str, map_name) = match name {
+            Some(name) => (
+                util::str_to_cstring(name.as_ref())?,
+                name.as_ref().to_string(),
+            ),
+
+            // The old version kernel don't support specifying map name, we can use 'Option::<&str>::None' for the name argument.
+            None => (util::str_to_cstring("")?, "".to_string()),
+        };
+
+        let map_type = map_type.into();
+        let map_name_ptr = {
+            if map_name_str.as_bytes().is_empty() {
+                null()
+            } else {
+                map_name_str.as_ptr()
+            }
+        };
+
+        let fd = unsafe {
+            libbpf_sys::bpf_map_create(
+                map_type,
+                map_name_ptr,
+                key_size,
+                value_size,
+                max_entries,
+                opts,
+            )
+        };
+
+        if fd < 0 {
+            return Err(Error::System(fd));
+        }
+
+        Ok(Map {
+            fd,
+            name: map_name,
+            ty: map_type,
+            key_size,
+            value_size,
+            ptr: std::ptr::null_mut(),
+        })
+    }
 }
 
 #[rustfmt::skip]
@@ -426,7 +493,7 @@ bitflags! {
 // If you add a new per-cpu map, also update `is_percpu`.
 #[non_exhaustive]
 #[repr(u32)]
-#[derive(Clone, TryFromPrimitive, PartialEq, Display)]
+#[derive(Clone, TryFromPrimitive, IntoPrimitive, PartialEq, Display)]
 pub enum MapType {
     Unspec = 0,
     Hash,
--- a/libbpf-rs/tests/test.rs
+++ b/libbpf-rs/tests/test.rs
@@ -9,7 +9,7 @@ use nix::errno;
 use plain::Plain;
 use scopeguard::defer;
 
-use libbpf_rs::{num_possible_cpus, Iter, MapFlags, Object, ObjectBuilder};
+use libbpf_rs::{num_possible_cpus, Iter, Map, MapFlags, MapType, Object, ObjectBuilder};
 
 fn get_test_object_path(filename: &str) -> PathBuf {
     let mut path = PathBuf::new();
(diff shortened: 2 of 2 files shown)


libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
...
-	const EXIST    = 1 << 1;
-	const LOCK     = 1 << 2;
+        const ANY      = 0;
+        const NO_EXIST = 1;
+        const EXIST    = 1 << 1;
+        const LOCK     = 1 << 2;
+    }
+
+    /// Flags to configurae [`Map::create_map`] operations.
+    pub struct MapCreateFlags: u32 {
```
insearchoflosttime, line 490: libbpf-sys already has bindings for these flags (https://github.com/libbpf/libbpf-sys/blob/master/src/bindings.rs#L511-L523). I think we can just reuse those instead of redefining them here

libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
...
     /// Flags to configure [`Map`] operations.
     pub struct MapFlags: u64 {
-	const ANY      = 0;
-	const NO_EXIST = 1;
-	const EXIST    = 1 << 1;
-	const LOCK     = 1 << 2;
+        const ANY      = 0;
+        const NO_EXIST = 1;
+        const EXIST    = 1 << 1;
+        const LOCK     = 1 << 2;
```
insearchoflosttime, line 486: I'm not sure what the change here is - this looks the same to me. Maybe some whitespace difference? Would you mind reverting any whitespace changes here so it doesn't show up as modified?

libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
     pub fn keys(&self) -> MapKeyIter {
         MapKeyIter::new(self, self.key_size())
     }
+
+    /// Create the bpf map standalone.
+    pub fn create_map(
+        name: &str,
```
insearchoflosttime, line 432: I think defining the type here as AsRef<str> allows for more generic usages. In other libbpf-rs functions, we pass in AsRef<str>s.

libbpf-rs/tests/test.rs
```diff
@@ -647,3 +649,40 @@ fn test_object_task_iter() {
     // Check for init
     assert!(items.iter().any(|&item| item.pid == 1));
 }
+
+#[test]
+fn test_create_map_and_pin() {
+    bump_rlimit_mlock();
+
+    let mut map = Map::create_map(
```
insearchoflosttime, line 657: I'm seeing this test fail in CI. 
```

running 1 test
test test_create_map_and_pin ... FAILED

failures:

---- test_create_map_and_pin stdout ----
thread 'test_create_map_and_pin' panicked at 'assertion failed: `(left == right)`
  left: `-1`,
 right: `0`: Setting RLIMIT_MEMLOCK failed with errno: 1', libbpf-rs/tests/test.rs:48:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_create_map_and_pin

```

Would you mind taking a look?

insearchoflosttime (changes requested):

libbpf-rs/src/map.rs
zyy17: I think you are right(the reason why I create `MapCreateFlags` is influend by `MapFlags`). In your opinion, I think the `create_map()` should be like(use `u32` type on `flags`):

```rust
    pub fn create_map(
        name: &str,
        map_type: MapType,
        key_size: u32,
        value_size: u32,
        max_entries: u32,
        flags: u32,
    ) -> Result<Map> { ... }
```

And user can call the `create_map()` like this:

```rust
...
    let mut map = Map::create_map(
        "mymap_test_create_map_and_pin",
        MapType::Hash,
        4,
        8,
        8,
        libbpf_sys::BPF_F_NO_PREALLOC
    )
...
```

zyy17: I saw the next comments and `bpf_map_create_opts` will be better.

libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
...
+        map_type: MapType,
+        key_size: u32,
+        value_size: u32,
+        max_entries: u32,
+        flags: MapCreateFlags,
```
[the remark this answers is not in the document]
zyy17: sounds reasonable to me.

libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
     pub fn keys(&self) -> MapKeyIter {
         MapKeyIter::new(self, self.key_size())
     }
+
+    /// Create the bpf map standalone.
+    pub fn create_map(
+        name: &str,
+        map_type: MapType,
```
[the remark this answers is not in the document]
zyy17: sounds reasonable to me.

libbpf-rs/tests/test.rs
zyy17: OK, I will fix the CI error.

libbpf-rs/src/map.rs
zyy17: sounds reasonable to me.

libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
...
+            inner_map_fd: 0,
+            map_extra: 0,
+            numa_node: 0,
+            map_ifindex: 0,
+        };
+
+        let map_name_c = util::str_to_cstring(name)?;
+        let fd = unsafe {
+            libbpf_sys::bpf_map_create(
+                map_type.clone() as u32,
```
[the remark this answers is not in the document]
zyy17: I will return `Result<Map>`:

```rust
        Ok(Map {
            fd,
            name: name.to_string(),
            ty: map_type as u32,
            key_size,
            value_size,
            ptr: std::ptr::null_mut(),
        })
```
 
Because the `MapType` is not copyable and the `map_type` need to use in `bpf_map_create()`,  then I use the `clone()` to reuse the variable.

libbpf-rs/src/map.rs
```diff
@@ -409,16 +426,81 @@ impl Map {
     pub fn keys(&self) -> MapKeyIter {
         MapKeyIter::new(self, self.key_size())
     }
+
+    /// Create the bpf map standalone.
+    pub fn create_map(
```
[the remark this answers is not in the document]
zyy17: sounds reasonable to me.

libbpf-rs/tests/test.rs
zyy17: It seems that the original test function `test_create_map_and_pin()` needs sudo operation and should skip in `Run tests`(`./github/rust/yml`).
I rename the test function to `test_object_map_create_and_pin()` which start with `test_object_` prefix to adapt the CI configurations.

libbpf-rs/src/map.rs
```diff
@@ -153,18 +153,41 @@ impl Map {
...
         let path_ptr = path_c.as_ptr();
 
-        let ret = unsafe { libbpf_sys::bpf_map__pin(self.ptr, path_ptr) };
+        let ret = if self.ptr.is_null() {
+            if self.fd <= 0 {
+                return Err(Error::InvalidInput(format!(
+                    "fd must be greater than 0 and got {}",
+                    self.fd
+                )));
+            }
```
chenhengqi, line 162: Why do you need this `self.fd <= 0` check ?

libbpf-rs/src/map.rs
```diff
@@ -409,6 +432,41 @@ impl Map {
...
     }
+
+    /// Create the bpf map standalone.
+    pub fn create<T: AsRef<str>>(
+        map_type: MapType,
+        name: T,
+        key_size: u32,
+        value_size: u32,
+        max_entries: u32,
+        opts: libbpf_sys::bpf_map_create_opts,
```
chenhengqi, line 443: `opts` should be pointer.

libbpf-rs/src/map.rs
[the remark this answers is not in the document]
chenhengqi: I think let MapType derive `num_enum::IntoPrimitive` is a better choice.

libbpf-rs/src/map.rs
zyy17: I think you're right, it's not necessary to check `fd`. I will remove the check.

libbpf-rs/src/map.rs
zyy17: sounds reasonable to me.

libbpf-rs/src/map.rs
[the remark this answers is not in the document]
zyy17: I'm not familiar with `num_enum` crate, I will try your advice after reading the docs.

libbpf-rs/src/map.rs
```diff
@@ -409,6 +420,42 @@ impl Map {
     pub fn keys(&self) -> MapKeyIter {
         MapKeyIter::new(self, self.key_size())
     }
+
+    /// Create the bpf map standalone.
+    pub fn create<T: AsRef<str>>(
+        map_type: MapType,
+        name: T,
```
anakryiko, line 427: in libbpf map name is optional as some (very) old kernels don't support specifying map name. So I wonder if this should be an Option[T]?

zyy17: You're right, `Option<T>` will be better.

insearchoflosttime:
Thanks for making the edits and apologies for the late reply. Can you combine the commits instead of having it as 10 separate ones? That generally makes the commit log a lot cleaner when it gets merged in. Thanks.

zyy17:
> Thanks for making the edits and apologies for the late reply. Can you combine the commits instead of having it as 10 separate ones? That generally makes the commit log a lot cleaner when it gets merged in. Thanks.

Thank you for your review. I already squash the commits and make it as a clean commit 😊

insearchoflosttime (commented):
Thanks for working on this! Left a few comments inline, the rest LGTM!

libbpf-rs/src/map.rs
```diff
@@ -2,10 +2,11 @@ use core::ffi::c_void;
 use std::convert::TryFrom;
 use std::path::Path;
 use std::ptr;
+use std::ptr::null;
 
 use bitflags::bitflags;
-use nix::{errno, unistd};
-use num_enum::TryFromPrimitive;
+use nix::{errno, unistd, NixPath};
```
insearchoflosttime, line 8: Where does NixPath get used?

libbpf-rs/src/map.rs
```diff
@@ -409,6 +421,50 @@ impl Map {
...
+        value_size: u32,
+        max_entries: u32,
+        opts: &libbpf_sys::bpf_map_create_opts,
+    ) -> Result<Map> {
+        let (map_name_c_ptr, map_name) = match name {
+            Some(name) => (
+                util::str_to_cstring(name.as_ref())?.as_ptr(),
+                name.as_ref().to_string(),
+            ),
+            // The old version kernel don't support specifying map name, we can use 'Option::<&str>::None' for the name argument.
```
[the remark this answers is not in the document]
insearchoflosttime: Same comment here on this revision - could you break this comment up into two lines?

libbpf-rs/src/map.rs
zyy17: My mistake... I will fix that.

insearchoflosttime:
Thanks for your work on this.

PR #214, opened 2022-04-12
Problems with internal maps (.rodata) and old kernels

trtt:
Found some problems with >= libbpf-sys 0.6.0 and old kernels (in my case it was CentOS/RHEL 7). The problems were with internal maps (.rodata) generated (but not really used) by a relatively new compiler. 

What I found:
1. double free in the error path that actually resulted in segfault
2. after getting rid of the segfault there remains an error with loading such programs

### DIFF
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -266,12 +266,10 @@ impl OpenObject {
             return Err(Error::System(-ret));
         }
 
-        let obj = Object::new(self.ptr)?;
-
+        let ptr = self.ptr;
         // Prevent object from being closed once `self` is dropped
         self.ptr = ptr::null_mut();
-
-        Ok(obj)
+        Object::new(ptr)
     }
 }
 
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -306,10 +306,10 @@ impl Object {
         };
 
         // Populate obj.maps
-        let mut map: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
+        let mut next_ptr: *mut libbpf_sys::bpf_map = std::ptr::null_mut();
         loop {
             // Get the pointer to the next BPF map
-            let next_ptr = unsafe { libbpf_sys::bpf_object__next_map(obj.ptr, map) };
+            next_ptr = unsafe { libbpf_sys::bpf_object__next_map(obj.ptr, next_ptr) };
             if next_ptr.is_null() {
                 break;
             }
@@ -327,10 +327,12 @@ impl Object {
 
             // Get the map fd
             let fd = unsafe { libbpf_sys::bpf_map__fd(next_ptr) };
-            // Old kernels can have uninitialized internal maps:
+
+            // Old kernels can have uninitialized internal maps with fd -1:
             // https://github.com/libbpf/libbpf/blob/5c31bcf220f66e70f39fd141f5b0c55c6ab65e8e/src/libbpf.c#L5009-L5022
-            if fd < 0 && !unsafe { libbpf_sys::bpf_map__is_internal(next_ptr) } {
-                return Err(Error::System(-fd));
+            // Skip them
+            if fd < 0 {
+                continue;
             }
 
             // Add the map to the hashmap
@@ -338,7 +340,6 @@ impl Object {
                 name.clone(),
                 Map::new(fd, name, def.type_, def.key_size, def.value_size, next_ptr),
             );
-            map = next_ptr;
         }
 
         // Populate obj.progs
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -327,7 +327,9 @@ impl Object {
 
             // Get the map fd
             let fd = unsafe { libbpf_sys::bpf_map__fd(next_ptr) };
-            if fd < 0 {
+            // Old kernels can have uninitialized internal maps:
+            // https://github.com/libbpf/libbpf/blob/5c31bcf220f66e70f39fd141f5b0c55c6ab65e8e/src/libbpf.c#L5009-L5022
+            if fd < 0 && !unsafe { libbpf_sys::bpf_map__is_internal(next_ptr) } {
                 return Err(Error::System(-fd));
             }
 


libbpf-rs/src/object.rs
```diff
@@ -266,12 +266,10 @@ impl OpenObject {
...
         }
 
-        let obj = Object::new(self.ptr)?;
-
+        let ptr = self.ptr;
         // Prevent object from being closed once `self` is dropped
         self.ptr = ptr::null_mut();
-
-        Ok(obj)
+        Object::new(ptr)
```
chenhengqi, line 272: I don't see double-free on the original code and your changes would introduce a leak.

libbpf-rs/src/object.rs
```diff
@@ -327,7 +327,9 @@ impl Object {
 
             // Get the map fd
             let fd = unsafe { libbpf_sys::bpf_map__fd(next_ptr) };
-            if fd < 0 {
+            // Old kernels can have uninitialized internal maps:
+            // https://github.com/libbpf/libbpf/blob/5c31bcf220f66e70f39fd141f5b0c55c6ab65e8e/src/libbpf.c#L5009-L5022
+            if fd < 0 && !unsafe { libbpf_sys::bpf_map__is_internal(next_ptr) } {
```
chenhengqi, line 332: `bpf_map__is_internal` check is handled by the generated skeleton. Why do we need it here ?

I am afraid the changes I made in #192 is wrong.

trtt: I left reference to `libbpf` in the comment, it basically explains the reason.
For old kernels some data maps (internal) are not really created and their `fd` is `-1` after loading the object. So without relaxing the checks in `Object::new` it just won't load.

> I am afraid the changes I made in https://github.com/libbpf/libbpf-rs/pull/192 is wrong.

Sorry, I don't see the connection.

libbpf-rs/src/object.rs
chenhengqi: OK, I see the problem. We can prevent create `Object` prematurely in Object::new() instead.

libbpf-rs/src/object.rs
chenhengqi: In the generated skeleton, we have:

```rust
    pub fn load(mut self) -> libbpf_rs::Result<RunqslowerSkel<'a>> {
        let ret = unsafe { libbpf_sys::bpf_object__load_skeleton(self.skel_config.get()) };
        if ret != 0 {
            return Err(libbpf_rs::Error::System(-ret));
        }

        let obj = unsafe { libbpf_rs::Object::from_ptr(self.obj.take_ptr())? };

        Ok(RunqslowerSkel {
            obj,
            skel_config: self.skel_config,
            links: RunqslowerLinks::default(),
        })
    }
```

If we reach `Object::new`, all BPF maps and progs should be loaded (or skipped) successfully.

I think we can skip `fd < 0` check entirely.

libbpf-rs/src/object.rs
trtt: So you want `Object::new` to take ownership of the input pointer in one codepath, but not the other one. It's kinda ambiguous, don't you think? That's exactly the reason there is a bug here. Could you care to explain why is it better than the solution I provided?

libbpf-rs/src/object.rs
trtt: I'm a bit confused. You linked code from another crate. I was not even aware of it, nor have I used it.

> I think we can skip fd < 0 check entirely.

If you think it's ok, sure. I assumed there should be a reason for them to be there in the first place.

insearchoflosttime, line 332: Do we also need the "!kernel_supports(obj, FEAT_GLOBAL_DATA))" check here as well?

libbpf-rs/src/object.rs
insearchoflosttime: I think your current approach here @trtt is the most straightforward one / simplest fix to me.

insearchoflosttime:
Thanks for finding this and submitting a fix! I left a comment in your 2nd commit but other than that, this makes sense to me!

libbpf-rs/src/object.rs
chenhengqi: The code I posted is geneated by libbpf-cargo (examples/runqslower), the primary user of `Object`. In this use-case, if we reach `Object::new`, we pass the libbpf skeleton load phase.

As I mentioned above, if we can't skip this check here, the change I maded in #192 is probably wrong.

trtt: That would be ideal. I'm afraid these bits are not exposed by libbpf (`kernel_supports` itself and the array it queries are `static`).

trtt: In the PR you've linked I see that you've removed similar checks for program fds, are you referring to that? I don't think this is necessarily wrong. To me, these checks look more like safety checks, but honestly I don't know.

libbpf-rs/src/object.rs
```diff
@@ -329,7 +327,9 @@ impl Object {
```
anakryiko, line 327: as an aside, bpf_map__def() is deprecated, we should try to remove it before libbpf 1.0 release

libbpf-rs/src/object.rs
```diff
@@ -329,7 +327,9 @@ impl Object {
 
             // Get the map fd
             let fd = unsafe { libbpf_sys::bpf_map__fd(next_ptr) };
-            if fd < 0 {
+            // Old kernels can have uninitialized internal maps:
+            // https://github.com/libbpf/libbpf/blob/5c31bcf220f66e70f39fd141f5b0c55c6ab65e8e/src/libbpf.c#L5009-L5022
+            if fd < 0 && !unsafe { libbpf_sys::bpf_map__is_internal(next_ptr) } {
                 return Err(Error::System(-fd));
```
anakryiko, line 333: as @chenhengqi  mentioned, if object load succeeded, then map can't have fd < 0, except if that map wasn't created *due to supported reasons*. Right now the only case when map is not created is `.rodata.*` that's not used on old kernels that don't support such maps. But I'm planning on adding a feature that will allow to opt-out map loading by user. So we definitely shouldn't hardcode any assumption about is_internal_map, FEAT_GLOBAL_DATA, etc.

If map's fd is < 0, then we either should skip it or keep it with FD -1 (or whatever other means).

trtt:
Hey, I'm still interested in fixing this. Could you give me a clearer indication on what needs to change? Do you all agree on removing the `fd < 0` check completely?  Or maps with -1 fd should be filtered out instead of causing error? Or something else entirely?

insearchoflosttime:
Thanks for following up @trtt! imo the consensus is that if fd < 0 then we should skip it.

trtt:
> Thanks for following up @trtt! imo the consensus is that if fd < 0 then we should skip it.

done

AnthonyMikh:
Hi. From what I can see, the problem still persists and libbpf-rs still bails out on encountering a map with `-1` fd. Is it correct? If so, what can I do in order to get this fix merged? Does it need some kind of kernel version check to be implemented?

danielocfb:
Hi @AnthonyMikh, if it's still an issue for you then feel free to open a pull request with the fix. Please also make sure to include a test reproducing the original problem. Thanks!

AnthonyMikh:
@danielocfb thanks for fast answer.

> if it's still an issue for you then feel free to open a pull request with the fix.

I could probably do it, but running `cargo test` (as well as `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="sudo -E" cargo test`) on fresh repo clone fails. Do tests require `libbpf-cargo` to be installed in order to work? I was unable to find such technical details in contribution guide.

> Please also make sure to include a test reproducing the original problem.

It would be hard to do since it requires using a specific version of kernel, and so far all the tests seemingly use just the host version of OS. How would I add such a test? Would [testcontainers](docs.rs/testcontainers) help?

And anyway, I suspect this issue is not the best place to ask such questions. What would be an appropriate place for development-related discussions?

danielocfb:
> @danielocfb thanks for fast answer.
> 
> > if it's still an issue for you then feel free to open a pull request with the fix.
> 
> I could probably do it, but running `cargo test` (as well as `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="sudo -E" cargo test`) on fresh repo clone fails. Do tests require `libbpf-cargo` to be installed in order to work? I was unable to find such technical details in contribution guide.

`libbpf-cargo` doesn't have to be installed. If it's just the tests requiring root failing, you can run `cargo test` from each of the sub-directories (`libbpf-rs` / `libbpf-cargo`). If not, we need details.

> > Please also make sure to include a test reproducing the original problem.
> 
> It would be hard to do since it requires using a specific version of kernel, and so far all the tests seemingly use just the host version of OS. How would I add such a test? Would [testcontainers](docs.rs/testcontainers) help?

Yeah, tests currently run on host kernel. If it's specific to a certain kernel version we'd still want a test, even if it passes at this point. At least we can run it as a one-off on an older kernel and are sure to catch regressions.

Running tests on different kernels is definitely an option, though I wouldn't want a test suite that requires docker to be installed unconditionally, as `testcontainers` seems to require it. I think it should be a GitHub Actions thing (but will require some research to understand our options there). I don't think it will be a requirement for the fix to be merged, though.

> And anyway, I suspect this issue is not the best place to ask such questions. What would be an appropriate place for development-related discussions?

Personally I am fine either way. You can open an issue or start a [discussion](https://github.com/libbpf/libbpf-rs/discussions).

PR #221, opened 2022-05-01
libbpf-rs: update libbpf-sys dependency

willfindlay:
libbpf-sys 0.7.1 includes a newer libbpf version that has many useful features including the new
BPF_KPROBE_SYSCALL macro. Let's update so that users that are using a vendored libbpf can
take advantage of these features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,7 +54,7 @@ name = "bpf_query"
 version = "0.1.0"
 dependencies = [
  "libbpf-rs",
- "nix",
+ "nix 0.23.1",
  "structopt",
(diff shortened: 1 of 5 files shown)


willfindlay:
I'll fix the build failures later this evening.

BurntBrunch:
@willfindlay this got merged in a way that did not move `master`. The resulting commit (e6fcd07) is not reachable from a tag or a branch and can only be accessed via `git fetch origin e6fcd07 && git checkout e6fcd07`.

Not sure if I'm missing something but this feels unintentional.

PR #222, opened 2022-05-05
Add struct_ops attach support

dschatzberg:
Libbpf has support for BPF STRUCT_OPS
https://lwn.net/Articles/809092/ which allows for bpf to implement
specific kernel's function pointers.

libbpf has bpf_map__attach_struct_ops() to attach these. This commit
adds the equivalent functionality to libbpf-rs without which
STRUCT_OPS cannot be used.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -476,6 +476,26 @@ impl Map {
             ptr: std::ptr::null_mut(),
         })
     }
+
+    /// Attach a struct ops map
+    pub fn attach_struct_ops(&mut self) -> Result<Link> {
+        if self.map_type() != MapType::StructOps {
+            return Err(Error::InvalidInput(format!(
+                "Invalid map type ({}) for attach_struct_ops()",
+                self.map_type(),
+            )));
+        }
+
+        let ret = unsafe {
+            let p = libbpf_sys::bpf_map__attach_struct_ops(self.ptr);
+            let rc = libbpf_sys::libbpf_get_error(p as *const c_void);
+            if rc != 0 {
+                return Err(Error::System(rc as i32));
+            }
+            p
+        };
+        Ok(Link::new(ret))
+    }
 }
 
 #[rustfmt::skip]


anakryiko (approved):
LGTM from overall libbpf API standpoint. @insearchoflosttime can you please check that all the libbpf-rs-specific issues are taken care of? Thanks!

anakryiko:
@dschatzberg there is some build issue, please take a look: https://github.com/libbpf/libbpf-rs/runs/6313461399?check_suite_focus=true

libbpf-rs/src/map.rs
```diff
@@ -476,6 +476,22 @@ impl Map {
...
+                "Invalid map type ({}) for attach_struct_ops()",
+                self.map_type(),
+            )));
+        }
+
+        let ret = unsafe { libbpf_sys::bpf_map__attach_struct_ops(self.ptr) };
+        if ret.is_null() {
+            return Err(Error::System(nix::errno::errno()));
+        }
+        Ok(Link::new(ret))
```
chenhengqi, line 493: I think you should use `libbpf_get_error()` instead.

anakryiko: should libbpf-rs use libbpf 1.0 strict mode always?

chenhengqi: Not sure where and how to place `libbpf_set_strict_mode()` in the codebase in a Rust way.

libbpf-rs/src/map.rs
```diff
@@ -476,6 +476,26 @@ impl Map {
...
+    /// Attach a struct ops map
+    pub fn attach_struct_ops(&mut self) -> Result<Link> {
+        if self.map_type() != MapType::StructOps {
+            return Err(Error::InvalidInput(format!(
+                "Invalid map type ({}) for attach_struct_ops()",
+                self.map_type(),
+            )));
+        }
+
+        let ret = unsafe {
```
insearchoflosttime, line 489: nit: would you mind renaming "ret" to "ptr" to make it more clear that it returns a link ptr?

libbpf-rs/src/map.rs
anakryiko: We can probably add it before bpf_object__open() call. But that's separate from this change.

libbpf-rs/src/map.rs
```diff
@@ -476,6 +476,26 @@ impl Map {
...
+            let p = libbpf_sys::bpf_map__attach_struct_ops(self.ptr);
```
insearchoflosttime, line 490: Unfortunately, I think this will crash in the case where the map is created by the userspace application (in the create call in line 428 above) because the ptr is null in those cases (we only get the fd). What are your thoughts on adding a check for whether self.ptr is null and erroring out in those cases?

insearchoflosttime (changes requested):
Thanks for working on this!

PR #223, opened 2022-05-06
libbpf-cargo: enable split building and generation in SkeletonBuilder

willfindlay:
It might be desirable for an end-user to run SkeletonBuilder directly on an object file
rather than compiling with clang. This could be useful for a few reasons, such as
establishing greater control over the build process or working with vendored object files.

As such, this commit refactors SkeletonBuilder slightly, breaking generate() up into
build() and generate() respectively. A new function is added called build_and_generate()
which encompasses the old behaviour. To make this work, we also remove the source argument
from SkeletonBuilder::new() and instead shift over to SkeletonBuilder::source() and
SkeletonBuilder::obj() to specify a source and object file respectively.

In cases where SkeletonBuilder::obj is missing, we fall back to creating a temporary
directory as before.

Note that this is technically a breaking change for users of the SkeletonBuilder API
(generally in a build.rs file for example). I've updated the examples as necessary to reflect this.

Signed-off-by: William Findlay <will@isovalent.com>

### DIFF
--- a/libbpf-cargo/src/lib.rs
+++ b/libbpf-cargo/src/lib.rs
@@ -55,10 +55,12 @@
 //! build`. This is a convenience command so you don't forget any steps. Alternatively, you could
 //! write a Makefile for your project.
 
-use std::path::{Path, PathBuf};
-use std::result;
+use std::{
+    path::{Path, PathBuf},
+    result,
+};
 
-use tempfile::tempdir;
+use tempfile::{tempdir, TempDir};
 use thiserror::Error;
 
 // libbpf-cargo binary is the primary consumer of the following modules. As such,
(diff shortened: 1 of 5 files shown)


libbpf-cargo/src/lib.rs
```diff
@@ -97,35 +99,63 @@ pub type Result<T> = result::Result<T, Error>;
...
+
+impl Default for SkeletonBuilder {
+    fn default() -> Self {
+        Self::new()
+    }
 }
 
 impl SkeletonBuilder {
     /// Create a new builder instance, where `source` is the path to the BPF object source
     /// (typically suffixed by `.bpf.c`)
```
insearchoflosttime, line 128: I think you can remove this comment as well

libbpf-cargo/src/lib.rs
```diff
@@ -97,35 +99,63 @@ pub type Result<T> = result::Result<T, Error>;
...
-    pub fn new<P: AsRef<Path>>(source: P) -> Self {
+    pub fn new() -> Self {
         SkeletonBuilder {
             debug: false,
-            source: source.as_ref().to_path_buf(),
+            //source: source.as_ref().to_path_buf(),
```
insearchoflosttime, line 132: I think you meant to remove this commented out line too?

insearchoflosttime (commented):
lgtm

willfindlay:
Fixed up the comments, thanks

PR #226, opened 2022-05-14
libbpf-cargo: use clap derive

loshz:
As stated in the [docs](https://github.com/TeXitoi/structopt/blob/master/README.md#maintenance): structopt is now in maintenance mode.

This PR migrates to use the new [clap derive](https://github.com/clap-rs/clap/tree/master/clap_derive) functionality that has an almost identical interface.
I have updated and compiled all of the examples to ensure they are still working as expected.

Signed-off-by: Dan Bond <danbond@protonmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,30 +12,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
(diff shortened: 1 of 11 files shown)


insearchoflosttime:
Awesome! Thanks for adding this!

PR #228, opened 2022-05-16
ring buffer: less strict lifetime for callback

tobithiel:
Similar to #152 this makes the lifetime requirements for `RingBuffer` callbacks less strict. Currently they require `'static`, but they only need to live as long as the the `RingBuffer` object (they won't be called if the `RingBuffer` is gone).


insearchoflosttime:
lgtm, thanks for submitting this @tobithiel!

PR #229, opened 2022-06-02
ringbuf: Allow poll() to block for an indefinite time period

insearchoflosttime:
ringbuf: Allow poll() to block for an indefinite time period

    ring_buffer__poll in libbpf internally calls epoll_wait. Specifying a
    timeout of -1 to epoll_wait causes it to block indefinitely if no event
    transpires. We should support this functionality in libbpf-rs as well.

    In this change, if the user passes in Duration::MAX as the timeout, this
    will cause the ringbuf poll to block indefinitely until an event occurs.

### DIFF
--- a/libbpf-rs/src/ringbuf.rs
+++ b/libbpf-rs/src/ringbuf.rs
@@ -132,11 +132,17 @@ pub struct RingBuffer<'a> {
 impl<'a> RingBuffer<'a> {
     /// Poll from all open ring buffers, calling the registered callback for
     /// each one. Polls continually until we either run out of events to consume
-    /// or `timeout` is reached.
+    /// or `timeout` is reached. If `timeout` is Duration::MAX, this will block
+    /// indefinitely until an event occurs.
     pub fn poll(&self, timeout: Duration) -> Result<()> {
         assert!(!self.ptr.is_null());
 
-        let ret = unsafe { libbpf_sys::ring_buffer__poll(self.ptr, timeout.as_millis() as i32) };
+        let mut timeout_ms = -1;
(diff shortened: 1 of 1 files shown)


anakryiko (approved):
kind of unfortunate that all existing users will have to update their use of ringbuf, but oh well

insearchoflosttime:
An alternative would be to introduce a new function like epoll_indefinite

danobi:
Couldn't you check if the value is `Duration::MAX` and pass -1 if it is? Not sure we need an API break for this

anakryiko:
> Couldn't you check if the value is `Duration::MAX` and pass -1 if it is? Not sure we need an API break for this

@insearchoflosttime did you get a chance to try @danobi's suggestion with Duration::MAX ?

insearchoflosttime:
@anakryiko  yes! I will rebase and push it out today.

libbpf-rs/src/ringbuf.rs
```diff
@@ -136,7 +136,12 @@ impl<'a> RingBuffer<'a> {
     pub fn poll(&self, timeout: Duration) -> Result<()> {
         assert!(!self.ptr.is_null());
 
-        let ret = unsafe { libbpf_sys::ring_buffer__poll(self.ptr, timeout.as_millis() as i32) };
+        let mut timeout_ms = -1;
+        if timeout != Duration::MAX {
+            timeout_ms = timeout.as_millis() as i32;
```
anakryiko, line 141: looks good, but let's update the doc comment for poll() to specify that Duration::MAX is used to mean "infinite timeout"?

PR #232, opened 2022-06-22
libbpf-cargo: Add open_opts method for SkelBuilder.

chengshuyi:
SkelBuilder only has the open method, and we can't easily pass in parameters like btf_custom_path.

Therefore, this commit adds the open_opts method for SkelBuilder. Developers can obtain an instance of libbpf_sys::bpf_object_open_opts through obj_builder.opts(std::ptr::null()), and then set the btf_custom_path parameter.

Signed-off-by: Shuyi Cheng <chengshuyi@linux.alibaba.com>

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -716,6 +716,22 @@ fn gen_skel_contents(_debug: bool, raw_obj_name: &str, obj_file_path: &Path) -> 
                     skel_config
                 }})
             }}
+
+            pub fn open_opts(self, open_opts: libbpf_sys::bpf_object_open_opts) -> libbpf_rs::Result<Open{name}Skel<'a>> {{
+                let mut skel_config = build_skel_config()?;
+
+                let ret = unsafe {{ libbpf_sys::bpf_object__open_skeleton(skel_config.get(), &open_opts) }};
+                if ret != 0 {{
+                    return Err(libbpf_rs::Error::System(-ret));
+                }}
+
+                let obj = unsafe {{ libbpf_rs::OpenObject::from_ptr(skel_config.object_ptr())? }};
+
+                Ok(Open{name}Skel {{
+                    obj,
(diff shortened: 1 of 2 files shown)


anakryiko:
is the idea that user will use ObjectBuilder to construct bpf_object_open_opts and then get it through (#doc(hidden)!) opts()? If yes, should we be a bit more explicit about that? Maybe it's better to have a native Rust counterpart to these opts?

chengshuyi:
Thanks for your review. 

> If yes, should we be a bit more explicit about that? 

yes. 

> Maybe it's better to have a native Rust counterpart to these opts?

Do you mean to provide the rust version of libbpf_sys::bpf_object_open_opts? That is, to create a struct ObjectOpenOpts structure?

insearchoflosttime:
> > > Maybe it's better to have a native Rust counterpart to these opts?
> > 
> > 
> > Do you mean to provide the rust version of libbpf_sys::bpf_object_open_opts? That is, to create a struct ObjectOpenOpts structure?
> 
> yep, as an idea... @insearchoflosttime any opinion?

I don't think bpf_object_open_opts gets used that often so just using  libbpf_sys::bpf_object_open_opts seems sufficient to me, but I also don't feel strongly about this. Either way sounds good to me.

chengshuyi:
> > > > Maybe it's better to have a native Rust counterpart to these opts?
> > > 
> > > 
> > > Do you mean to provide the rust version of libbpf_sys::bpf_object_open_opts? That is, to create a struct ObjectOpenOpts structure?
> > 
> > 
> > yep, as an idea... @insearchoflosttime any opinion?
> 
> I don't think bpf_object_open_opts gets used that often so just using libbpf_sys::bpf_object_open_opts seems sufficient to me, but I also don't feel strongly about this. Either way sounds good to me.

Okay, then i still use just using libbpf_sys::bpf_object_open_opts, I will update pr this week.

Issue #231, opened 2022-06-15
the use of btf_custom_path


dafu-wu:
I'm new to Rust，How do I pass the BTF file to btf_custom_path？
thank you very much for taking the time to answer my little friend.

insearchoflosttime:
Hi @dafu-wu. Can you clarify your use case? We don't currently support passing in a BTF file.

dafu-wu:
> 

First of all, thank you. Some of the kernel CONFIG_DEBUG_INFO_BTF=n,but we can get BTF file from btfhub or pahole,so,I want to pass a local BTF file to 'bpf_object__open_file'.

insearchoflosttime:
Gotcha, I just took a look at https://lpc.events/event/11/contributions/948/attachments/906/1776/LPC21_Towards_truly_portable_eBPF.pdf which has some more information about btfhub and using a custom external BTF file.

If you want to pass in the BTF file to btf_custom_path on your local version of libbpf-rs, I think you'll need to expand ObjectBuilder::opts to take in a the btf custom path as an arg, and then pass that to the libpf_sys::bpf_object_open_opts call (https://github.com/libbpf/libbpf-rs/blob/master/libbpf-rs/src/object.rs#L50). Does this answer your question?

chengshuyi:
'''
pub fn open(mut self) -> libbpf_rs::Result<OpenIcmpSkel<'a>> {
        let mut skel_config = build_skel_config()?;
        let open_opts = self.obj_builder.opts(std::ptr::null());

        let ret = unsafe { libbpf_sys::bpf_object__open_skeleton(skel_config.get(), &open_opts) };
        if ret != 0 {
            return Err(libbpf_rs::Error::System(-ret));
        }

        let obj = unsafe { libbpf_rs::OpenObject::from_ptr(skel_config.object_ptr())? };

        Ok(OpenIcmpSkel { obj, skel_config })
    }
'''

Maybe we can refer to the code generated by libbpf-cargo and add an open_opts function. 

'''
 pub fn open_opts(mut self, open_opts: libbpf_sys::bpf_object_open_opts) -> libbpf_rs::Result<OpenIcmpSkel<'a>> {
        let mut skel_config = build_skel_config()?;
        // let open_opts = self.obj_builder.opts(std::ptr::null());

        let ret = unsafe { libbpf_sys::bpf_object__open_skeleton(skel_config.get(), &open_opts) };
        if ret != 0 {
[...]

PR #233, opened 2022-06-27
libbpf-cargo: add fn open_with_btf_custom_path for SkelBuilder

dafu-wu:
add fn open_with_btf_custom_path for SkelBuilder

Temporarily resolved the problem that custom BTF files cannot be sent parameters.
#231

### DIFF
--- a/libbpf-cargo/src/gen.rs
+++ b/libbpf-cargo/src/gen.rs
@@ -716,6 +716,24 @@ fn gen_skel_contents(_debug: bool, raw_obj_name: &str, obj_file_path: &Path) -> 
                     skel_config
                 }})
             }}
+
+            pub fn open_with_btf_custom_path(mut self, btf_custom_path: &str) -> libbpf_rs::Result<Open{name}Skel<'a>> {{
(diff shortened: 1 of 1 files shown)


insearchoflosttime (commented):
Do we need `open_with_btf_custom_path` or can the user just pass to `open_opts` an opts with the btf path set?

dafu-wu:
> Do we need `open_with_btf_custom_path` or can the user just pass to `open_opts` an opts with the btf path set?

It doesn't look like there are any other parameters to pass,constructing an open_opts seems to be a costly undertaking.
Of course, if we provide a way to build open_opts easily, I think it is a better solution.

chengshuyi:
> Do we need `open_with_btf_custom_path` or can the user just pass to `open_opts` an opts with the btf path set?

I personally think open_with_btf_custom_path and open_opts do not conflict, they give users more choices.

anakryiko:
a big fat no to open_with_btf_custom_path, open and open_opts should be more than enough, there is no need to add extra method for each possible open_opts field

PR #239, opened 2022-07-10
libbpf-cargo: allow user to set _TARGET_ARCH flag though clang_args.

saimizi:
When doing cross-compile such as compiling bpf programs for arm64 on a
x86_64 PC like following:

	cargo build --target aarch64-unknown-linux-gnu --release

_TARGET_ARCH should be the target one (arm64) instead of the one of the
build host (x86).

In this case, cross-compile user may set _TARGET_ARCH him/herself though
clang_args. This commit skip setting _TARGET_ARCH when found it in
clang_args.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -111,26 +111,36 @@ fn check_clang(debug: bool, clang: &Path, skip_version_checks: bool) -> Result<(
 ///
 /// for each prog.
 fn compile_one(debug: bool, source: &Path, out: &Path, clang: &Path, options: &str) -> Result<()> {
-    let arch = match std::env::consts::ARCH {
-        "x86_64" => "x86",
-        "aarch64" => "arm64",
-        _ => std::env::consts::ARCH,
-    };
-
     if debug {
         println!("Building {}", source.display());
     }
 
     let mut cmd = Command::new(clang.as_os_str());
+    let mut arch_set = false;
+
+    // When doing cross-compile, user may set -D__TARGET_ARCH_<arch> in clang_args.
     if !options.is_empty() {
         cmd.args(options.split_whitespace());
+
+        if options.contains("-D__TARGET_ARCH_") {
+            arch_set = true;
+        }
     }
+
+    if !arch_set {
+        let arch = match std::env::consts::ARCH {
+            "x86_64" => "x86",
+            "aarch64" => "arm64",
+            _ => std::env::consts::ARCH,
+        };
+        cmd.arg(format!("-D__TARGET_ARCH_{}", arch));
+    }
+
     cmd.arg("-g")
         .arg("-O2")
         .arg("-target")
         .arg("bpf")
         .arg("-c")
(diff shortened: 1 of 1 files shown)


libbpf-cargo/src/build.rs
```diff
@@ -111,26 +111,38 @@ fn check_clang(debug: bool, clang: &Path, skip_version_checks: bool) -> Result<(
...
-    };
-
     if debug {
         println!("Building {}", source.display());
     }
 
     let mut cmd = Command::new(clang.as_os_str());
+    let mut arch_set = false;
+
+    /*
```
insearchoflosttime, line 121: nit: in Rust, i think the convention is to use "//" for comments. do you mind changing this to using "//"?

insearchoflosttime:
LGTM!

saimizi:
"_TARGET_ARCH_*" is really an argument to clang,  so I think using "clang_args" to set it is much more straightforward.

libbpf-cargo/src/build.rs
saimizi: Sure, I updated the commit to use change "/*" to "//", thank you.

libbpf-cargo/src/build.rs
```diff
@@ -111,26 +111,36 @@ fn check_clang(debug: bool, clang: &Path, skip_version_checks: bool) -> Result<(
...
     }
+
+    if !arch_set {
+        let arch = match std::env::consts::ARCH {
+            "x86_64" => "x86",
+            "aarch64" => "arm64",
+            _ => std::env::consts::ARCH,
+        };
+        cmd.arg(format!("-D__TARGET_ARCH_{}", arch));
+    }
```
insearchoflosttime, line 137: I think this logic could be simplified with something like:


    let mut cmd = Command::new(clang.as_os_str());

    if !options.is_empty() {
        cmd.args(options.split_whitespace());
    }

    if !options.contains("-D__TARGET_ARCH_") {
        let arch = match std::env::consts::ARCH {
            "x86_64" => "x86",
            "aarch64" => "arm64",
            _ => std::env::consts::ARCH,
        };
        cmd.arg(format!("-D__TARGET_ARCH_{}", arch));
    }


and that way we don't need the arch_set. What are your thoughts?

saimizi: I think your fix is better!
Is it possible for you to update the patch and add a "Signed-off-by" when merging the code?

insearchoflosttime: I will try to figure out how to do this in github :) What "Signed-off-by" name should I put for you? 

And if I can't figure this out, then I will just land this as-is and submit a follow-up. 

Thanks for contributing to libbpf-rs.

saimizi: > I will try to figure out how to do this in github :)  What "Signed-off-by" name should I put for you?

Thank you.
Please use follows:
     
     Signed-off-by: Seimizu Joukan <joukan.seimizu@gmail.com>

insearchoflosttime:
Merged in by command line.

PR #241, opened 2022-07-13
Allow generated file to be included via `include!()`

heinrich5991:
Do this by putting everything in a submodule so that we can have global
attributes without `#![]` style attributes that aren't permitted to be
used through `include!()`.


examples/capable/src/capable.rs
```diff
@@ -0,0 +1 @@
+include!(concat!(env!("OUT_DIR"), "/skel.rs"));
```
insearchoflosttime, line 1: instead of having to create a new file for this, does having this line inside the src/main.rs file work?

examples/capable/build.rs
```diff
@@ -1,23 +1,14 @@
...
-    // that cannot be `include!()`ed via macro. And we cannot use the `#[path = "..."]`
-    // trick either because you cannot yet `concat!(env!("OUT_DIR"), "/skel.rs")` inside
-    // the path attribute either (see https://github.com/rust-lang/rust/pull/83366).
-    //
-    // However, there is hope! When the above feature stabilizes we can clean this
-    // all up.
-    create_dir_all("./src/bpf/.output").unwrap();
-    let skel = Path::new("./src/bpf/.output/capable.skel.rs");
+    let mut out = PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR must be set in build script"));
+    out.push("skel.rs");
```
insearchoflosttime, line 8: I think these "skel.rs" should stay the same naming as "bpf-file-name.skel.rs" to stick with convention

examples/capable/build.rs
```diff
@@ -1,23 +1,14 @@
...
-    // Reasons are because the generated skeleton contains compiler attributes
-    // that cannot be `include!()`ed via macro. And we cannot use the `#[path = "..."]`
-    // trick either because you cannot yet `concat!(env!("OUT_DIR"), "/skel.rs")` inside
-    // the path attribute either (see https://github.com/rust-lang/rust/pull/83366).
-    //
-    // However, there is hope! When the above feature stabilizes we can clean this
-    // all up.
-    create_dir_all("./src/bpf/.output").unwrap();
-    let skel = Path::new("./src/bpf/.output/capable.skel.rs");
+    let mut out = PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR must be set in build script"));
```
insearchoflosttime, line 7: If the user doesn't have the OUT_DIR environment variable set, can we create one as a default instead of failing out?

examples/runqslower/build.rs
```diff
@@ -1,23 +1,14 @@
...
 
 fn main() {
-    // It's unfortunate we cannot use `OUT_DIR` to store the generated skeleton.
-    // Reasons are because the generated skeleton contains compiler attributes
-    // that cannot be `include!()`ed via macro. And we cannot use the `#[path = "..."]`
-    // trick either because you cannot yet `concat!(env!("OUT_DIR"), "/skel.rs")` inside
-    // the path attribute either (see https://github.com/rust-lang/rust/pull/83366).
-    //
-    // However, there is hope! When the above feature stabilizes we can clean this
-    // all up.
```
insearchoflosttime, line 15: Can you clarify why this is now okay to do?

examples/capable/src/capable.rs
heinrich5991: Done.

examples/capable/build.rs
heinrich5991: Done.

examples/capable/build.rs
heinrich5991: The build script is guaranteed to have an `OUT_DIR` variable: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
> `OUT_DIR` — the folder in which all output and intermediate artifacts should be placed. This folder is inside the build directory for the package being built, and it is unique for the package in question.

examples/runqslower/build.rs
heinrich5991: The commit message tries to summarize it. The comment says
> ```rs
>     // Reasons are because the generated skeleton contains compiler attributes
>     // that cannot be `include!()`ed via macro. […]
> ```
If you tried that before my commit, you'd get errors about the `#![]`-style attributes. I removed those by moving all of the definitions into a submodule where the attributes can be applied using `#[]`-style attributes.

insearchoflosttime: Gotcha, thanks for clarifying

insearchoflosttime:
@heinrich5991 , can you squash your 2 commits into 1?

heinrich5991:
Done.

insearchoflosttime:
This looks great! Thanks for adding this!!

PR #242, opened 2022-07-14
Expose libbpf-sys' static feature

javierhonduco:
To statically link elfutils and libz. Tested on [rbperf](https://github.com/javierhonduco/rbperf) compiled with `RUSTFLAGS="-L /usr/lib -L /usr/lib64" cargo b`

**With the feature disabled (the default)**

```
[javierhonduco@fedora rbperf]$ ldd target/debug/rbperf
        linux-vdso.so.1 (0x00007fffa8de3000)
        libelf.so.1 => /lib/libelf.so.1 (0x00007ff9a5727000)
        libz.so.1 => /lib64/libz.so.1 (0x00007ff9a570d000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007ff9a56ed000)
        libm.so.6 => /lib64/libm.so.6 (0x00007ff9a560f000)
        libc.so.6 => /lib64/libc.so.6 (0x00007ff9a540e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff9a5fa2000)
```

**With the static feature**
```
[javierhonduco@fedora rbperf]$ ldd target/debug/rbperf
        linux-vdso.so.1 (0x00007ffc5ed6a000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007ff391800000)
        libm.so.6 => /lib64/libm.so.6 (0x00007ff391722000)
        libc.so.6 => /lib64/libc.so.6 (0x00007ff391521000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff392098000)
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,11 +19,13 @@ jobs:
         components: rustfmt, clippy
         override: true
     - name: Install deps
-      run: sudo apt-get install -y clang-12 libelf-dev
+      run: sudo apt-get install -y clang-12 libelf-dev zlib1g-dev
     - name: Symlink clang
       run: sudo rm -f /bin/clang && sudo ln -s /usr/bin/clang-12 /bin/clang
     - name: Build
       run: cargo build --verbose --workspace --exclude runqslower
+    - name: Build capable example with static libelf and libz
+      run: RUSTFLAGS='-L /usr/lib/x86_64-linux-gnu' cargo b --package capable --features=static
     - name: Run tests
       # Skip BTF tests which require sudo
       # Skip BTF dump float test for now, we can enable it when we have access to clang 13+
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,9 +256,9 @@ dependencies = [
 
 [[package]]
 name = "libbpf-sys"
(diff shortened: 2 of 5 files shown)


.github/workflows/rust.yml
```diff
@@ -19,11 +19,13 @@ jobs:
         components: rustfmt, clippy
         override: true
     - name: Install deps
-      run: sudo apt-get install -y clang-12 libelf-dev
+      run: sudo apt-get install -y clang-12 libelf-dev zlib1g-dev
```
javierhonduco, line 22: Perhaps this worked before because clang or libelf require it, too. Let's make it explicit

insearchoflosttime: Does the capable example with static libelf and libz fail to compile without zlib1g-dev explicitly added?

javierhonduco: It should still work as it's already installed on the machine. It includes [both the dynamic library and the static archive](https://packages.ubuntu.com/focal/amd64/zlib1g-dev/filelist). Added it to make the dependencies clearer for people reading the CI setup

insearchoflosttime:
Thanks for adding this @javierhonduco!

PR #248, opened 2022-07-26
libbpf-rs: add map freeze

insearchoflosttime:


danielocfb (commented):
The change seems sensible to me. Thanks! I am just wondering whether we should add a doc comment providing more details on what the method does, as it seems to be publicly accessible by users. What do you think?

insearchoflosttime:
I think that's a good idea, I will update this accordingly

danielocfb (approved):
Thanks!

PR #249, opened 2022-07-26
libbpf-rs: Add more setter functionalities on open maps

insearchoflosttime:
As part of the initiative to reach feature-parity with libbpf-1.0, this
commit adds:

* set_map_flags
* set_numa_node
* set_map_extra
* set_autocreate
* set_pin_path

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -62,8 +62,36 @@ impl OpenMap {
         util::parse_ret(ret)
     }
 
+    pub fn set_map_flags(&mut self, flags: u32) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_map__set_map_flags(self.ptr, flags) };
+        util::parse_ret(ret)
+    }
+
+    pub fn set_numa_node(&mut self, numa_node: u32) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_map__set_numa_node(self.ptr, numa_node) };
+        util::parse_ret(ret)
+    }
+
     pub fn set_inner_map_fd(&mut self, inner: &Map) {
         unsafe { libbpf_sys::bpf_map__set_inner_map_fd(self.ptr, inner.fd()) };
+    }
+
+    pub fn set_map_extra(&mut self, map_extra: u64) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_map__set_map_extra(self.ptr, map_extra) };
+        util::parse_ret(ret)
+    }
+
+    pub fn set_autocreate(&mut self, autocreate: bool) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_map__set_autocreate(self.ptr, autocreate) };
+        util::parse_ret(ret)
+    }
+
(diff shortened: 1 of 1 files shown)


libbpf-rs/src/map.rs
```diff
@@ -26,7 +26,7 @@ impl OpenMap {
         OpenMap { ptr }
     }
 
-    pub fn set_map_ifindex(&mut self, idx: u32) {
+    pub fn set_ifindex(&mut self, idx: u32) {
```
danielocfb, line 29: Can you comment on why this rename is here, please? Is it intentional? I didn't see it mentioned in the pull request/commit description.

libbpf-rs/src/map.rs
```diff
@@ -62,10 +62,38 @@ impl OpenMap {
...
+    pub fn set_autocreate(&mut self, autocreate: bool) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_map__set_autocreate(self.ptr, autocreate) };
+        util::parse_ret(ret)
+    }
+
+    pub fn set_pin_path<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
+        let path_c = util::path_to_cstring(path)?;
+        let path_ptr = path_c.as_ptr();
+
+        let ret = unsafe { libbpf_sys::bpf_map__set_pin_path(self.ptr, path_ptr) };
```
danielocfb, line 93: I am new to the code base, so apologies if this is a stupid question, but what ensures that `path` lives sufficiently long here? Asked differently, what happens if it goes out of scope while `self` continues to live on? To me it seems as if we'd end up with a dangling pointer.

danielocfb (commented):
Looks mostly good to me, two questions, though.

libbpf-rs/src/map.rs
insearchoflosttime: The intention of the rename was to make it consistent with the libbpf API `bpf_map__set_ifindex`. I'll edit the commit to mention this - I thought since the change is so minimal it'd make more sense to bundle it in this change to the map setter functions than to have it be its own commit. Though if that's preferred, I dont' mind separating this to its own commit :)

libbpf-rs/src/map.rs
insearchoflosttime: Great question. path is strdup-ed internally by libbpf (https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/tree/tools/lib/bpf/libbpf.c#n7893Z)

insearchoflosttime:
Thanks for taking a look @danielocfb !

libbpf-rs/src/map.rs
danielocfb: Makes sense. Personally I'd split it out just to make it obvious and have basically the description you have at hand, but it's probably fine either way.

libbpf-rs/src/map.rs
danielocfb: Ah, thanks for the pointer! Makes sense now.

danielocfb (approved):
Thanks for the explanation! To me this change looks good.

insearchoflosttime:
update: remove naming change for "set_map_ifindex" -> "set_ifindex". Will make this naming change as part of its own commit.

PR #250, opened 2022-07-26
libbpf-rs: clean up unnecessary 'fd as i32' casting in maps

insearchoflosttime:
Map's fd is already defined as i32. We do not need to cast it
explicitly with "fd as i32".


danielocfb:
It seems as if formatting should now be changed slightly :-( Otherwise the change looks good to me. Have we considered hooking up `clippy` in CI, by any chance? It *should* be able to catch such minor oversights before check-in.

danielocfb (approved):

PR #251, opened 2022-07-26
libbpf-rs: Add autoload and log level APIs for programs

insearchoflosttime:
As part of the initiative to reach feature-parity with libbpf-1.0,
this commit adds wrapper functions around:

    * bpf_program__set_log_level
    * bpf_program__log_level
    * bpf_program__autoload

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -119,6 +119,12 @@ impl OpenProgram {
         }
     }
 
+    /// Set the log level for the bpf program
+    pub fn set_log_level(&mut self, log_level: u32) -> Result<()> {
+        let ret = unsafe { libbpf_sys::bpf_program__set_log_level(self.ptr, log_level) };
+        util::parse_ret(ret)
+    }
+
     /// Name of the section this `OpenProgram` belongs to.
     pub fn section(&self) -> &str {
         &self.section
@@ -133,6 +139,8 @@ impl OpenProgram {
             .map_err(|e| Error::Internal(e.to_string()))
     }
 
+    /// Set whether a bpf program should be automatically loaded by default
+    /// when the BPF object is loaded
     pub fn set_autoload(&mut self, autoload: bool) -> Result<()> {
         let ret = unsafe { libbpf_sys::bpf_program__set_autoload(self.ptr, autoload) };
         util::parse_ret(ret)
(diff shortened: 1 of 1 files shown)


libbpf-rs/src/program.rs
```diff
@@ -419,4 +432,8 @@ impl Program {
     pub fn insn_cnt(&self) -> usize {
         unsafe { libbpf_sys::bpf_program__insn_cnt(self.ptr) as usize }
     }
+
+    pub fn unload(&mut self) {
```
dxuuu, line 436: Do we need to consume `self` here? Does closing the prog break any invariants held by `Program`?

dxuuu (commented):
Please add docs to public methods

Side note: maybe also consider adding the `#![deny(missing_docs)]` at the crate level to enforce that all public APIs have docs.

libbpf-rs/src/program.rs
anakryiko, line 436: @insearchoflosttime can you please drop unload (libbpf doesn't expose it anymore), add doc comments and resubmit?

libbpf-rs/src/program.rs
```diff
@@ -376,6 +384,16 @@ impl Program {
         }
     }
 
+    /// Return true if the bpf program is set to autoload else false
```
danielocfb, line 387: Consider using markdown for boolean values:
```patch
-/// Return true if the bpf program is set to autoload else false
+/// Return `true` if the bpf program is set to autoload, `false` otherwise
```

libbpf-rs/src/program.rs
```diff
@@ -376,6 +384,16 @@ impl Program {
...
+    pub fn autoload(&self) -> bool {
+        unsafe { libbpf_sys::bpf_program__autoload(self.ptr) }
+    }
+
+    /// Return the bpf program log level
```
danielocfb, line 392:
```patch
-/// Return the bpf program log level
+/// Return the bpf program's log level
```

libbpf-rs/src/program.rs
```diff
@@ -119,6 +119,12 @@ impl OpenProgram {
         }
     }
 
+    /// Set the log level for the bpf program
+    pub fn set_log_level(&mut self, log_level: u32) -> Result<()> {
```
danielocfb, line 123: Can we document what `log_level` is? It's not obvious to users. Seems to be some bitmap kind of thing, is that right?

danielocfb (commented):
Thanks for working on exposing these functions! I mostly have one comment below.

danielocfb:
Will finish this up.

PR #254, opened 2022-07-27
libbpf-rs: Add wrapper for ringbuffer__epoll_fd

insearchoflosttime:
As part of the initiative to reach feature-parity with libbpf-1.0, this
commit adds a wrapper for the ringbuffer__epoll_fd API.

### DIFF
--- a/libbpf-rs/src/ringbuf.rs
+++ b/libbpf-rs/src/ringbuf.rs
@@ -146,6 +146,13 @@ impl<'a> RingBuffer<'a> {
 
         util::parse_ret(ret)
     }
+
+    /// Get an fd that can be used to sleep until data is available
+    pub fn epoll_fd(&self) -> i32 {
+        assert!(!self.ptr.is_null());
+
+        unsafe { libbpf_sys::ring_buffer__epoll_fd(self.ptr) }
+    }
 }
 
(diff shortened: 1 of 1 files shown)


libbpf-rs/src/ringbuf.rs
```diff
@@ -146,6 +146,15 @@ impl<'a> RingBuffer<'a> {
 
         util::parse_ret(ret)
     }
+
+    /// Get an fd that can be used to sleep until data is available
+    pub fn epoll_fd(&self) -> Result<()> {
```
danobi, line 151: Shouldn't this be returning an FD?

And does it need a `Result` wrapper? Doesn't the `RingBuffer` hold the invariant that the underlying `struct ring_buffer` is valid?

insearchoflosttime: The fd may be an error value if it failed to create the epoll instance (https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/ringbuf.c#L177)

insearchoflosttime: Nvm, just realized that epoll_fd will be called only when fd is value. I'll update this commit :) thanks for catching this!

danobi (approved):

PR #257, opened 2022-08-11
Used contained value's default for Arrays when implementing Default

uepoch:
This makes sure that the type identifier contained in the array is constructed with a valid default.

Since we were hardcoding `::default()` after the type identifier, it was not working well with nested arrays

Example:
```
  uint8_t pcomms[5][16];
```
Would fail because
```
[[u8; 16]::default(); 5]
```
is invalid.

This would turn the impl in:
```
[[u8::default; 16]; 5]
```

It's a fairly minor change that should not impact the majority of types

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -7,7 +7,7 @@ use std::mem::size_of;
 use std::os::raw::{c_char, c_ulong};
 use std::slice;
 
-use anyhow::{bail, ensure, Result};
+use anyhow::{anyhow, bail, ensure, Result};
 use scroll::Pread;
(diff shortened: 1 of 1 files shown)


insearchoflosttime:
LGTM. it looks like the build is failing because of some clippy warnings, so after https://github.com/libbpf/libbpf-rs/pull/258 gets landed, I'll merge this in.

danielocfb:
> LGTM. it looks like the build is failing because of some clippy warnings, so after #258 gets landed, I'll merge this in.

I retriggered the CI run now that #258 landed. Let's see whether everything is green now.

uepoch:
Alrighty

Issue #260, opened 2022-08-28
Derive Debug


mdaverde:
Possible to derive [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) trait output for some of the public-facing structs?

Besides the ability to debug from the output, when we build higher-level structs that use libbpf-rs's types internally, we can't derive `Debug` for our custom structs because we get a compiler error notifying us that libbpf-rs's types themselves don't implement `Debug`. 

It seems like we already do this for [some](https://github.com/libbpf/libbpf-rs/blob/master/libbpf-rs/src/tc.rs#L36) [structs](https://github.com/libbpf/libbpf-rs/blob/master/libbpf-rs/src/tc.rs#L219) but not others.

danielocfb:
Agreed, we most likely should do that.

danielocfb:
@mdaverde Are you interested in creating a pull request adding these derives? Should probably include warning on the `missing_debug_implementations` lint.

danielocfb:
With https://github.com/libbpf/libbpf-rs/pull/266 merged the scope of this issue got reduced to adding a few more `Debug` implementations to `struct`s containing pointers.

PR #266, opened 2022-09-07
libbpf-rs: Add derive(Debug) to query-related structs and enums

alexgartrell:
This is a simple quality-of-life improvement to make it a bit easier to use and discover the library.

```
fn main() {
    for prog in libbpf_rs::query::ProgInfoIter::default() {
        println!("{:#?}", prog);
    }
    for prog in libbpf_rs::query::LinkInfoIter::default() {
        println!("{:#?}", prog);
    }
    for prog in libbpf_rs::query::MapInfoIter::default() {
        println!("{:#?}", prog);
    }
}
```

### DIFF
--- a/libbpf-rs/src/iter.rs
+++ b/libbpf-rs/src/iter.rs
@@ -9,6 +9,7 @@ use crate::*;
 /// This implements [`std::io::Read`] for reading bytes from the iterator.
 /// Methods require working with raw bytes. You may find libraries such as
 /// [`plain`](https://crates.io/crates/plain) helpful.
+#[derive(Debug)]
 pub struct Iter {
     fd: i32,
 }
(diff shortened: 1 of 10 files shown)


danielocfb:
Looks good to me. Is this change sufficient to enable the `missing_debug_implementations` lint, crate-wide, so that we don't reintroduce public types without that derive, @alexgartrell ? And if so, would you mind adding it? Also see https://github.com/libbpf/libbpf-rs/issues/260.

alexgartrell:
@danielocfb it wasn't sufficient but I went ahead and instrumented everything else in a subsequent diff to add the warning.

danielocfb (approved):
Seems fine, thanks.

PR #268, opened 2022-09-08
Update minimum supported Rust version from 1.56 to 1.58

insearchoflosttime:
Version 1.58 is the minimum version that supports using captured identifiers in format strings, which is used within the libbpf-rs codebase.

This patch upgrades the minimum Rust version from 1.56 to 1.58.


danielocfb:
Looks good to me, though there is a CI failure. Can we try going to `1.58.1`, which should fix this problem: https://blog.rust-lang.org/2022/01/20/Rust-1.58.1.html#whats-in-1581-stable ?

PR #269, opened 2022-09-09
libbpf-rs: optionally return full vecs from query

alexgartrell:
NOTE: this is totally API-breaking, but the initial API wasn't really useful.

This avoids allocation overhead and guessing by reading the values and re-invoking the system call. Each array is blocked by a feature flag from the iterator.

Here's an example program using iced-x86 to disassembly the jited code:

```
use iced_x86::Formatter;

fn main() {
    for prog in libbpf_rs::query::ProgInfoIter::default().include_jited_prog_insns(true) {
        let mut d = iced_x86::Decoder::new(32, &prog.jited_prog_insns, 0);
        let mut f = iced_x86::GasFormatter::new();
        while d.can_decode() {
            let insn = d.decode();
            let mut f_insn = String::new();
            f.format(&insn, &mut f_insn);
            println!("{}", f_insn);
        }
    }
}
```

### DIFF
--- a/examples/bpf_query/Cargo.toml
+++ b/examples/bpf_query/Cargo.toml
@@ -9,3 +9,6 @@ edition = "2021"
 libbpf-rs = { path = "../../libbpf-rs" }
 nix = { version = "0.26", default-features = false, features = ["net", "user"] }
 clap = { version = "4.0.32", default-features = false, features = ["std", "derive", "help", "usage"] }
+
+[target.'cfg(target_arch = "x86_64")'.dependencies]
+iced-x86 = "1.20.0"
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -12,7 +12,9 @@
 
 use core::ffi::c_void;
 use std::convert::TryFrom;
+// TODO: convert remaining instances of size_of to size_of_val for clarity
 use std::mem::size_of;
+use std::mem::size_of_val;
 use std::os::raw::c_char;
 use std::string::String;
 use std::time::Duration;
@@ -25,6 +27,7 @@ use crate::util;
 use crate::MapType;
 use crate::ProgramAttachType;
 use crate::ProgramType;
+use crate::Result;
 
 macro_rules! gen_info_impl {
     // This magic here allows us to embed doc comments into macro expansions
@@ -103,6 +106,38 @@ fn name_arr_to_string(a: &[c_char], default: &str) -> String {
     }
 }
 
+#[derive(Clone, Debug)]
+/// BTF Line information
+pub struct LineInfo {
+    /// Offset of instruction in vector
+    pub insn_off: u32,
+    /// File name offset
+    pub file_name_off: u32,
+    /// Line offset in debug info
+    pub line_off: u32,
+    /// Line number
+    pub line_num: u32,
+    /// Line column number
+    pub line_col: u32,
+}
+
+impl From<&libbpf_sys::bpf_line_info> for LineInfo {
+    fn from(item: &libbpf_sys::bpf_line_info) -> Self {
+        LineInfo {
+            insn_off: item.insn_off,
+            file_name_off: item.file_name_off,
+            line_off: item.line_off,
+            line_num: item.line_col >> 10,
+            line_col: item.line_col & 0x3ff,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+#[repr(C)]
+/// Bpf identifier tag
+pub struct Tag([u8; 8]);
+
 /// Information about a BPF program
 #[derive(Debug, Clone)]
 // TODO: Document members.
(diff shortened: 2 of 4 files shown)


alexgartrell:
I don't think that this is a straightforward change so I'd consider this pull request more of an RFC. The tl;dr; is that it's really not valuable to just get the number of instructions if you actually want to introspect. The btf stuff is extra not-valuable for the same reason.

This probably ultimately requires actually writing out the code for the iterators instead of a big chunky macro to generate them, but I don't think that's the end of the world.

libbpf-rs/src/query.rs
```diff
@@ -98,101 +98,305 @@ fn name_arr_to_string(a: &[c_char], default: &str) -> String {
...
+        let mut line_info: Vec<LineInfoRaw> = Vec::new();
+        let mut func_info: Vec<FuncInfo> = Vec::new();
+        let mut jited_func_lens: Vec<u32> = Vec::new();
+        let mut prog_tags: Vec<Tag> = Vec::new();
+        let mut jited_ksyms: Vec<*const c_void> = Vec::new();
+
+        let item_ptr: *mut libbpf_sys::bpf_prog_info = &mut item;
+        let mut len = size_of::<libbpf_sys::bpf_prog_info>() as u32;
+
+        loop {
```
anakryiko, line 271: the actual loop looks unnecessary. You shouldn't need more than two calls for bpf_obj_get_info_by_fd: first to get lengths of necessary buffers, second to fill those buffers out by kernel. And then skip the second step if none of options were set. It should also simplify update_vec_field!() macro, probably, as you know that you are resizing vec.

libbpf-rs/src/query.rs
```diff
@@ -98,101 +98,305 @@ fn name_arr_to_string(a: &[c_char], default: &str) -> String {
...
+
 /// Information about a BPF program
 #[derive(Debug)]
 pub struct ProgramInfo {
     pub name: String,
     pub ty: ProgramType,
-    pub tag: [u8; 8],
+    pub tag: Tag,
     pub id: u32,
-    pub jited_prog_len: u32,
```
anakryiko, line 108: a bit conflicted on removing len fields. They seem useful for cases when user doesn't use ProgInfoQueryOptions and wants to do something smarted and custom on their own after fetching lengths. But this might be a bit hypothethical problem... Having vec fields and separate lens doesn't seem like a big deal?

anakryiko (commented):
I like these changes, as you said makes for much more usable API. I don't think breaking API change is a big deal, so I'd go ahead with this.

I left few comments on details (e.g., two explicit steps vs a loop), but those are minor. Let's wait for @danielocfb  to return to properly review this and land.

Thanks for improving libbpf-rs! :)

libbpf-rs/src/query.rs
alexgartrell: In this model, the _len is just available by calling .len() on the resulting vector. So I do think it's worth it to just do it the clean way in this instance

libbpf-rs/src/query.rs
```diff
@@ -98,101 +98,305 @@ fn name_arr_to_string(a: &[c_char], default: &str) -> String {
...
+#[repr(C)]
+struct LineInfoRaw {
+    insn_off: u32,
+    file_name_off: u32,
+    line_off: u32,
+    line_col: u32,
+}
+
+#[derive(Debug)]
+pub struct LineInfo {
```
danielocfb, line 116: Why do we need this type and can't use https://docs.rs/libbpf-sys/1.2.1+v1.2.0/libbpf_sys/struct.bpf_line_info.html? Isn't that what it is?

libbpf-rs/src/query.rs
```diff
@@ -98,101 +98,305 @@ fn name_arr_to_string(a: &[c_char], default: &str) -> String {
...
+    include_field_setter!(include_line_info);
+    include_field_setter!(include_func_info);
+    include_field_setter!(include_jited_line_info);
+    include_field_setter!(include_jited_func_lens);
+    include_field_setter!(include_prog_tags);
+    include_field_setter!(include_jited_ksyms);
+}
+
+macro_rules! update_vec_field {
+    ($item:expr, $flag:expr, $field_and_vec:ident, $len_field:ident, $default:expr, $result_ready:expr) => {
```
danielocfb, line 226: Have you tried expressing this as a function? It seems to me as if mutable references to objects could be passed in and that would suffice?

danielocfb (changes requested):
Sorry for the delay. Left some comments. I'd really want for us to try and limit the number of macros; it is borderline unmaintainable. I don't think it's obvious what can be passed in to said macros and there is basically no contract that each argument has to fulfill.

libbpf-rs/src/query.rs
danielocfb: Sorry, this comment was meant to apply to `LineInfoRaw`.

libbpf-rs/src/query.rs
danielocfb: I have no strong feelings either way. Unless we have a good use case for the additional length fields this model of just the `Vec` carrying the length is fine with me. We label ourselves "opinionated eBPF tooling" after all :-)

danielocfb:
> NOTE: this is totally API-breaking

I think that is fine.

libbpf-rs/src/query.rs
alexgartrell: FWIW, upon reflection, the reason to do this would be to write an application that only takes the jited instructions if there are <= N of them (for example). This is likely niche enough to ignore but I'll let people argue with me in V2

libbpf-rs/src/query.rs
```diff
@@ -98,101 +98,305 @@ fn name_arr_to_string(a: &[c_char], default: &str) -> String {
...
+}
+
+impl From<&LineInfoRaw> for LineInfo {
+    fn from(item: &LineInfoRaw) -> Self {
+        LineInfo {
+            insn_off: item.insn_off,
+            file_name_off: item.file_name_off,
+            line_off: item.line_off,
+            line_num: item.line_col >> 10,
+            line_col: item.line_col & 0x3ff,
```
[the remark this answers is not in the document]
alexgartrell: These are defined within macros in libbpf, so there's no clean way to pull it out.

https://www.kernel.org/doc/html/next/bpf/btf.html#:~:text=%23define%20BPF_LINE_INFO_LINE_NUM(line_col)%20%20%20%20%20%20%20%20((line_col)%20%3E%3E%2010)%0A%23define%20BPF_LINE_INFO_LINE_COL(line_col)%20%20%20%20%20%20%20%20((line_col)%20%26%200x3ff)

libbpf-rs/src/query.rs
```diff
@@ -498,29 +498,80 @@ gen_info_impl!(
...
+        let mut name: Vec<u8> = Vec::new();
+
+        let item_ptr: *mut libbpf_sys::bpf_btf_info = &mut item;
+        let mut len = size_of::<libbpf_sys::bpf_btf_info>() as u32;
+
+        let ret =
+            unsafe { libbpf_sys::bpf_obj_get_info_by_fd(fd, item_ptr as *mut c_void, &mut len) };
+        if ret != 0 {
+            return None;
+        }
```
danielocfb, line 519: Same comment as earlier about propagating errors up.

libbpf-rs/src/query.rs
```diff
@@ -498,29 +498,80 @@ gen_info_impl!(
...
-            btf_size: s.btf_size,
-            id: s.id,
-        })
+    fn load_from_fd(fd: i32) -> Option<Self> {
+        let mut item: libbpf_sys::bpf_btf_info = unsafe { std::mem::zeroed() };
+        let mut btf: Vec<u8> = Vec::new();
+        let mut name: Vec<u8> = Vec::new();
+
+        let item_ptr: *mut libbpf_sys::bpf_btf_info = &mut item;
+        let mut len = size_of::<libbpf_sys::bpf_btf_info>() as u32;
```
danielocfb, line 513:
```suggestion
        let mut len = size_of_val(&item) as u32;
```

libbpf-rs/src/query.rs
```diff
@@ -498,29 +498,80 @@ gen_info_impl!(
 // TODO: Document members.
 #[allow(missing_docs)]
 pub struct BtfInfo {
-    pub btf: u64,
-    pub btf_size: u32,
+    pub name: String,
+    pub btf: Vec<u8>,
```
danielocfb, line 502: Would you mind terribly documenting these changed members? We try to document everything that is new at the very least in the hopes that eventually we can remove all `#[allow(missing_docs)]`.

libbpf-rs/src/query.rs
```diff
@@ -498,29 +498,80 @@ gen_info_impl!(
...
 
 impl BtfInfo {
-    fn from_uapi(_fd: i32, s: libbpf_sys::bpf_btf_info) -> Option<Self> {
-        Some(Self {
-            btf: s.btf,
-            btf_size: s.btf_size,
-            id: s.id,
-        })
+    fn load_from_fd(fd: i32) -> Option<Self> {
+        let mut item: libbpf_sys::bpf_btf_info = unsafe { std::mem::zeroed() };
```
danielocfb, line 508: Nit: let's import `std` items we use at the top for the sake of consistency with existing code:
```suggestion
        let mut item: libbpf_sys::bpf_btf_info = unsafe { mem::zeroed() };
```

libbpf-rs/src/query.rs
```diff
@@ -498,29 +498,80 @@ gen_info_impl!(
...
+        item.name = name.as_mut_ptr() as *mut c_void as u64;
+
+        btf.resize(item.btf_size as usize, 0u8);
+        item.btf = btf.as_mut_ptr() as *mut c_void as u64;
+
+        let ret =
+            unsafe { libbpf_sys::bpf_obj_get_info_by_fd(fd, item_ptr as *mut c_void, &mut len) };
+        if ret == 0 {
+            Some(BtfInfo {
+                name: String::from_utf8(name).unwrap_or_else(|_| "(?)".to_string()),
```
danielocfb, line 534: One more suggestion: how about we make `BtfInfo::name` a `CString`. This way we can forego this opinionated fallible conversion and let users choose how to handle errors and it's barely more than a `to_string_lossy` call on their end if they opt not to.
It's what we do in the existing BTF functionality: https://docs.rs/libbpf-rs/0.21.2/libbpf_rs/btf/struct.BtfType.html#method.name

danielocfb (commented):
The updated changes look great! Thanks (and thanks for illustrating API usage in the example)! I left a few more minor comments.
Other than that I'd just hope we could add one or more unit tests for the changed functionality. Do you see a way to go about that? At least something like query a program and check non-zero instruction count or similar. Could also include parts of the example. It's mostly that we have something runnable in an automated fashion.

libbpf-rs/src/query.rs
alexgartrell: For what it's worth, ProgramInfo::name is a String as well. Should that be changed?

danielocfb: Ah, I missed that. Yes, that would be great. Though I won't insist you do it as you didn't really touch that member. So I'll leave it up to you. I added it to my list of things to do as well.

examples/bpf_query/Cargo.toml
```diff
@@ -9,3 +9,6 @@ edition = "2021"
 libbpf-rs = { path = "../../libbpf-rs" }
 nix = { version = "0.26", default-features = false, features = ["net", "user"] }
 clap = { version = "4.0.32", default-features = false, features = ["std", "derive", "help", "usage"] }
+
+[target.'cfg(target_arch = "x86_64")'.dependencies]
+iced-x86 = "1.20.0"
```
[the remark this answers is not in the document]
alexgartrell: It's a runtime dependency for the example in particular. Not a new dependency for libbpf-rs

libbpf-rs/src/query.rs
alexgartrell: Fixing this instance, but there are others in this file for a subsequent diff

examples/bpf_query/Cargo.toml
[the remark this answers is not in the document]
danielocfb: Ah yes, I somehow read this as `libbpf-rs`'s `Cargo.toml`.

alexgartrell:
I agree the lack of testing is a problem, but I propose we provide it in another diff. The real path here is to write cfg(test) stubs for the libbpf calls and mock it up completely, but that will be a monster change.

Absent that, you end up with tests that need sudo, which isn't super necessary/desirable here IMHO

libbpf-rs/src/query.rs
```diff
@@ -260,32 +497,84 @@ gen_info_impl!(
...
+                }
+                return None;
+            }
+
+            let info = BtfInfo::load_from_fd(fd);
+            let _ = close(fd);
+
+            match info {
+                Ok(i) => return Some(i),
+                Err(e) => eprintln!("Failed to load btf information: {}", e),
```
danielocfb, line 573: Let's please not print stuff in a library. It will end up in callers without any context in places where it is unlikely to be expected.

libbpf-rs/src/query.rs
```diff
@@ -260,32 +497,84 @@ gen_info_impl!(
 
 /// Information about BPF type format
 #[derive(Debug, Clone)]
-// TODO: Document members.
 #[allow(missing_docs)]
 pub struct BtfInfo {
-    pub btf: u64,
-    pub btf_size: u32,
+    /// The name associated with this btf information in the kernel
+    pub name: String,
```
danielocfb, line 503: Hm, I thought we agreed on making this `CString`. But now not even the comment shows up anymore...

danielocfb (approved):
Seems mostly fine now. I don't think you addressed all the points or perhaps I dreamed about pointing them out or GitHub swallowed them. Will create a fixup commit and add a CHANGELOG entry, but merging this change. Thanks!

danielocfb:
> Absent that, you end up with tests that need sudo, which isn't super necessary/desirable here IMHO

Half our tests already require `sudo`.

PR #274, opened 2022-10-01
Fix build for 32bit targets

### LINKED ISSUE #273: Can't build on 32bit architectures
Because of two hard coded `u64` in [`libbpf-rs/src/tc.rs`](https://github.com/libbpf/libbpf-rs/blob/de704c9e560d7b5679f9525ebf6d73ff8b05717a/libbpf-rs/src/tc.rs#L50) this library cannot be built for 32bit targets (see [here](https://gitlab.alpinelinux.org/crapStone/aports/-/jobs/858146#L538)).

Is it ok to include `libc` in the dependencies and convert the `usize` to a `libc::c_ulong`?

crapStone:
closes #273 

The hard coded `u64` in `tc.rs`.
This PR adds `libc` as dependency and changes the `u64` to `libc::c_ulong` to enable building for 32bit architectures.


danielocfb:
Could we use [`std::ffi::c_long`](https://doc.rust-lang.org/std/ffi/type.c_long.html) and forgo the `libc` dependency?

utky:
Is it suitable for this case to use [libbpf_sys::size_t](https://docs.rs/libbpf-sys/1.0.1+v1.0.0/libbpf_sys/type.size_t.html) instead of `u64`?

PR #276, opened 2022-10-06
0.19

danielocfb:
Bump version to `0.19.0` in preparation for a release using `libbpf-sys` `1.0`.

Let me know if we should postpone for other features to be included.


danielocfb:
Also, I am not intending on bumping (and releasing) `libbpf-cargo`. Let me know if that is necessary for some reason.

anakryiko (changes requested):
Let's bump libbpf-sys dependency to v1.0.1+v1.0.0 and release new versions of both libbpf-rs and libbpf-cargo. Given their dependency changed, I think both versions should be changed.

danielocfb:
> Let's bump libbpf-sys dependency to v1.0.1+v1.0.0 and release new versions of both libbpf-rs and libbpf-cargo. Given their dependency changed, I think both versions should be changed.

Sure thing, done.

anakryiko (changes requested):
there is still `libbpf-sys = { version = "1.0.0" }` in .tomls

danielocfb:
> there is still `libbpf-sys = { version = "1.0.0" }` in .tomls

Sure, but why should we want to change it? It's not like the crate doesn't work with version `1.0.0` (as evidenced by the fact that we just used that until now). The statement defines the minimum version. Control which version to use within the range of suitable ones lies with the consuming libraries. Bumping the minimum version only makes sense if we pull in a new feature that requires a higher version from my perspective. Otherwise we needlessly restrict the choices that downstream consumers can make.

anakryiko:
@danielocfb, because of the bug fixes in 1.0.1 (https://github.com/libbpf/libbpf/releases/tag/v1.0.1), there was a memory layout regression in v1.0 for bpf_object_open_opts. So let's bump to the very latest version.

danielocfb:
> So let's bump to the very latest version.

Done!

danielocfb:
We discussed the matter offline and identified a few other issues we should try to get done before release. `libbpf-sys` `1.0.1` actually uses `libbpf` `1.0.0` still. So we first intend to [bump](https://github.com/libbpf/libbpf-sys/pull/45) that. Then, because `libbpf` `1.0.1` actually broke compatibility with `1.0.0`, we may want to consider yanking `libbpf-sys` `1.0.x+1.0.0` from `crates.io` altogether (but that's @alexforster's call). In any event, we will have to wait for the new release of `libbpf-sys` first. Converting to draft for now.

alexforster:
Published `1.0.2+v1.0.1` – you should be good to go.

danielocfb:
Thanks Alex! I adjusted the pull request.

PR #277, opened 2022-10-11
libbpf-rs: Add support for USDT probes

ueno:
This adds a wrapper around bpf_program__attach_usdt introduced in libbpf-sys v1.0.0, which allows BPF programs to be attached to USDT probe points.

Signed-off-by: Daiki Ueno <dueno@redhat.com>


ueno:
@danielocfb I have added a test using the [probe](https://crates.io/crates/probe) crate. It works locally but might need some adjustment to cope with older Rust toolchain without `asm!`.

danielocfb (changes requested):
Your changes look great! Thanks for adding a test as well. I have one small request before we go ahead and merge, which is to please just bump our Rust version to use to `1.60` instead of making the test run conditionally (and drop the last commit altogether). Right now we are not overly concerned about the minimum Rust version we require and so it's fine to bump. [Here](https://github.com/libbpf/libbpf-rs/commit/838aca4afd9f5282a5fb433028e15c377cc2800c) is a commit showing the steps necessary.

ueno:
@danielocfb thank you for the pointer; I have updated the minimum required version (it was actually 1.59.0 that inline assembly got supported) and dropped the conditional compilation code.

danielocfb (approved):
Looks great to me! Thanks again.

PR #278, opened 2022-10-12
Update checkout and toolchain actions

danielocfb:
GitHub is complaining about usage of old Node.js on every GitHub Actions run:
> Node.js 12 actions are deprecated. For more information see:
> https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
> Please update the following actions to use Node.js 16: actions/checkout,
> actions-rs/toolchain, actions/checkout

Update the actions in question to get rid of those warnings.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,9 +15,9 @@ jobs:
         # Please adjust README when bumping version.
         rust: [stable, 1.58.1]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Install Rust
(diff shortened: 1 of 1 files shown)


danielocfb:
Well, it won't solve the warning coming from `toolchain` but it's a step in the right direction. I am not going to fix them up at this point as well.

anakryiko (approved):

danielocfb:
FWIW, [here](https://github.com/actions-rs/toolchain/pull/220) is the pull request to update `actions-rs/toolchain`.

PR #279, opened 2022-10-12
libbpf-cargo: Fix leaks of bpf_object

danielocfb:
Running the test suite with leak detection enabled makes it light up like a Christmas tree. This change fixes leaks of `bpf_object` objects in our generation code.


danielocfb:
We are also going to need a new patch release containing the leak fixes. I added a commit.

danielocfb:
Also updating `libbpf-sys` again to address https://github.com/libbpf/libbpf-rs/issues/280.

libbpf-cargo/src/btf/btf.rs
```diff
@@ -97,7 +103,7 @@ impl<'a> Btf<'a> {
             types: vec![BtfType::Void],
             ptr_size: ptr_size as u32,
             string_table: str_data,
-            bpf_obj,
+            _bpf_obj: bpf_obj,
```
anakryiko, line 106: it's actually weird and wrong for BTF object to contain bpf_object inside. But I don't think we need to do that here if we just make our own copy of raw_data and keep *that* in Btf. WDYT?

danielocfb: I've adjusted the code. Let me know if that's not what you had in mind.

anakryiko (approved):
LGTM as a general approach.

PR #282, opened 2022-10-14
Run leak sanitizer in CI

danielocfb:
Over in https://github.com/libbpf/libbpf-rs/pull/279 we fixed a regression caused by leakage of certain BPF objects. With this change we try to prevent such regressions from creeping in by enabling leak sanitizer when running our test suite in CI. Note that because we rely on unstable flags for said enablement, we need to run the nightly toolchain.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -13,8 +13,29 @@ jobs:
       fail-fast: false
       matrix:
         # Please adjust README when bumping version.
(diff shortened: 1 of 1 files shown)


danielocfb:
Here is a failure caused by one of the earlier fixes backed out: https://github.com/danielocfb/libbpf-rs/actions/runs/3252768280/jobs/5339337623#step:10:321

PR #284, opened 2022-10-21
examples: tproxy: Add BPF tproxy example

danobi:
This example application implements the classic iptables TPROXY target
using BPF. This is a nice example b/c it demonstrates a non-trivial BPF +
networking use case using libbpf-rs.

Signed-off-by: Daniel Xu <dxu@dxuuu.xyz>


anakryiko:
this looks short and simple enough to be part of libbpf-bootstrap maybe?

danobi:
Yeah, could be libbpf-bootstrap.

But just wanna add that one thing I really like about rust ecosystem is that examples in a project are generally in a standard location. libbpf-rs/libbpf-cargo doesn't strictly follow that pattern b/c we need to run build scripts per-example, but at least we have the `examples/` directory in the root.

For example, when I was developing `below` and using `cursive` (TUI crate), I found it incredibly helpful that they had a large directory of examples: https://github.com/gyscos/cursive/tree/main/cursive/examples . So I expect that libbpf-rs users might find a healthy examples/ directory here helpful as well.

.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.[ch]]
```
insearchoflosttime, line 1: is "root = true" also needed?

examples/tproxy/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "tproxy"
+version = "0.1.0"
+authors = ["Daniel Xu <dxu@dxuuu.xyz>"]
+edition = "2021"
```
insearchoflosttime, line 5: 2022? :)

danobi: Doesn't look like there's a 2022 edition yet: https://doc.rust-lang.org/edition-guide/introduction.html

.editorconfig
danobi: Looks like not needed but helpful.

PR #287, opened 2022-12-05
Update time crate dependency to 0.3.15

insearchoflosttime:
In libbpf-rs, the minimum supported Rust version is 1.59. libbpf-rs triggers builds in CI for version 1.59.

The latest compatible candidate version of the time crate in 1.59 is 0.3.15.

Without this change, the time crate may be upgraded to versions beyond 0.3.15, which will trigger a build failure for version 1.59.

### DIFF
--- a/examples/capable/Cargo.toml
+++ b/examples/capable/Cargo.toml
@@ -10,7 +10,7 @@ libbpf-rs = { path = "../../libbpf-rs" }
 libc = "0.2"
 phf = { version = "0.10", features = ["macros"] }
 plain = "0.2"
-time = { version = "0.3", features = ["formatting", "local-offset", "macros"]}
+time = { version = "=0.3.15", features = ["formatting", "local-offset", "macros"]}
 clap = { version = "3.1", default-features = false, features = ["std", "derive"] }
 
 [build-dependencies]
--- a/examples/runqslower/Cargo.toml
+++ b/examples/runqslower/Cargo.toml
@@ -9,7 +9,7 @@ anyhow = "1.0"
 libbpf-rs = { path = "../../libbpf-rs" }
 libc = "0.2"
 plain = "0.2"
-time = { version = "0.3", features = ["formatting", "local-offset", "macros"]}
+time = { version = "=0.3.15", features = ["formatting", "local-offset", "macros"]}
 clap = { version = "3.1", default-features = false, features = ["std", "derive"] }
 
 [build-dependencies]


insearchoflosttime:
This is the error I was seeing when time was upgraded to 0.3.17: https://github.com/libbpf/libbpf-rs/actions/runs/3536060022/jobs/5934747733

danielocfb:
> Without this change, the time crate may be upgraded to versions beyond 0.3.15, which will trigger a build failure for version 1.59.

> This is the error I was seeing when time was upgraded to 0.3.17: https://github.com/libbpf/libbpf-rs/actions/runs/3536060022/jobs/5934747733

For my understanding, in this run you specifically bumped time to `0.3.17`, right? The reason I am asking is because this is basically what `Cargo.lock` is used for here: it pins dependencies for CI runs (but of course also in general). So when I schedule a CI run on `master` it succeeded, because updating a pinned dependency is a manually requested action (well, it could also happen if another dependency *requires* a newer version, but that does not seem to be the case here I guess, as there exists a resolution with `0.3.15` as your change shows).

I agree that theoretically people could use a newer version of `time`. But that's fine, because they could also use a newer version of the toolchain as well. The way I see it, version `1.59` is a lower bound and so as long as there is *one* (or more) suitable version of all dependencies that compiles we are good. In general I believe a version specification (as in: the thing in `Cargo.toml`, not in `Cargo.lock`) pinning to a single version is frowned upon in the Rust eco system, because it restricts which versions clients can use tremendously, and can easily result in build breakage. Consider what happens if another dependency had a constraint `time = ">=0.3.16"` -- there is no candidate that matches all constraints now. It's not a big issue here, because we only do that for examples, which are binaries (and hence there are no downstream consumers in a strict sense), but I don't believe that it's really a necessary step either way.

So my question is: why not just *not* update `time` in `Cargo.lock`?

insearchoflosttime:
> For my understanding, in this run you specifically bumped time to 0.3.17, right? 

IIRC, time was automatically bumped up to 0.3.17 when I ran `cargo build`.

> Consider what happens if another dependency had a constraint time = ">=0.3.16" -- there is no candidate that matches all constraints now;

If another dependency had a constraint where time >=0.3.16, there would be no candidate that matches all constraints, even if time was not pinned to 0.3.15. I'm not sure what the correct answer here is, this would be incompatible regardless of what we do.

> So my question is: why not just not update time in Cargo.lock?

Ah, maybe I'm misunderstanding what Cargo.lock is then. From my understanding, Cargo.lock is automatically generated by cargo and not intended to be manually changed.

insearchoflosttime:
Reading through the docs, it also looks like Cargo.toml supports a [package] rust-version field to specify the minimum supported Rust version for a crate (https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html#cargo-rust-version) - adding this might be a better approach instead of pinning the crate to a specific version?

PR #288, opened 2022-12-14
Enforce minimum Rust version from Cargo

danielocfb:
Compilation problems caused by unsupported Rust features due to an old toolchain can be confusing. They also make it non-obvious what the minimum Rust version to upgrade to is.
Starting with Rust 1.56, Cargo honors the rust-version field [0] that causes it to fail compilation with a meaningful error message if the toolchain is not recent enough.
Set it for libbpf-rs and libbpf-cargo to have nicer errors reported to users in case of old toolchain usage.

[0] https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Please adjust README when bumping version.
+        # Please adjust README and rust-version field in Cargo.toml files when
+        # bumping version.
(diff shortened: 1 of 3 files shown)


anakryiko (approved):
Thanks!

danielocfb:
FWIW, this is what the error message will look like when attempting to build on an older toolchain: https://github.com/danielocfb/libbpf-rs/actions/runs/3699341115/jobs/6266599053#step:7:84

PR #291, opened 2022-12-15
Remove unnecessary 'err as i32' casts

insearchoflosttime:
In instances where err is already a 32-bit int, 'err as i32' is an unnecessary cast.


danielocfb:
Ugh, seems as if more casts were shadowed by the original failure.

danielocfb:
> Ugh, seems as if more casts were shadowed by the original failure.

Let's merge this first and the next one feeling like it can clean up the remaining bits. Seems to have been caused by a newer toolchain having better checks.

PR #292, opened 2022-12-16
Topic/btf rework

danielocfb:


danielocfb:
Also, for the interested, [here](https://morestina.net/blog/1868/self-referential-types-for-fun-and-profit) is a pretty accessible (IMO) article about self-referential types that closely reflect what is being done here.

PR #294, opened 2022-12-19
Stop alignment truncation

danielocfb:
The Rust code generation logic contains some alignment truncation logic -- limiting reported alignment to four on all systems -- whose purpose is unclear and the associated comment describes what is being done but not why. History of that logic is a single commit [0] importing the world.
The logic causes a bunch of padding to be added to Rust definitions that is not present when dumping the same BTF object in C (using bpftool, for example), making it unnecessary hard to compare emitted definitions.
Remove it. I was unable to find similar code in the Linux BTF and C code generation logic and have verified that bpftool does not emit padding for any of the types affected in tests [1].

[0] https://github.com/libbpf/libbpf-rs/commit/1f0be0fb192e61ee21b77325df5d6e2c1df78dd0 [1] https://gist.github.com/danielocfb/89e8bd7f28b5283ac94891984ff1742a

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -702,18 +702,9 @@ impl Btf {
         let align = if packed {
             1
         } else {
-            // Assume 32-bit alignment in case we're generating code for 32-bit
-            // arch. Worst case is on a 64-bit arch the compiler will generate
-            // extra padding. The final layout will still be identical to what is
-            // described by BTF.
             let a = self.align_of(type_id)? as usize;
             ensure!(a != 0, "Failed to get alignment of type_id: {}", type_id);
-
-            if a > 4 {
-                4
-            } else {
-                a
-            }
+            a
         };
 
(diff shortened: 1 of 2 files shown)


danobi:
@d-e-s-o IIRC the original reason was for compat with 32bit systems. Take for example:

```
struct Foo {
    uint32_t a;    /* BTF says offset for a == 0 */
    uint64_t b;    /* BTF says offset for b == 8 */
}
```

The above works as-is for 64 bit systems. But on 32-bit systems, you get:

```
struct Foo {
    uint32_t a;    /* offset == 0 */
    uint64_t b;    /* offset == 4 */
}
```

It's not super clear to me from the gist how bpftool is handling this. And I believe that assume-32-bit logic was transplanted from libbpf anyways. Having trouble finding it atm, but maybe @anakryiko remembers where it is.

danobi:
Ah yeah, here it is: https://github.com/torvalds/linux/commit/0f993845d723c8765655

danielocfb:
Interesting. Thanks for the pointer! Will leave this change out for now then.

PR #295, opened 2022-12-20
Add probes, and Program::insns

yuval-k:
motivation for this PR: I'm adding an option to my loader to check that **all** helpers in the program are supported by the kernel.
For this to work, i need to iterate the program instructions, get the called helpers and probe them.

To that end I have:
- added `insn` to OpenProgram (and for completeness, to Program too)
- added `prog_type` and `name` to OpenProgram
- Added a safe version libbpf probes in probe.rs

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -1,4 +1,5 @@
 use std::convert::TryFrom;
+use std::ffi::CStr;
 use std::path::Path;
 use std::ptr;
 
@@ -30,6 +31,14 @@ impl OpenProgram {
         }
     }
 
+    // The `ProgramType` of this `OpenProgram`.
+    pub fn prog_type(&self) -> ProgramType {
+        match ProgramType::try_from(unsafe { libbpf_sys::bpf_program__type(self.ptr) }) {
+            Ok(ty) => ty,
+            Err(_) => ProgramType::Unknown,
+        }
+    }
+
     pub fn set_attach_type(&mut self, attach_type: ProgramAttachType) {
         unsafe {
             libbpf_sys::bpf_program__set_expected_attach_type(self.ptr, attach_type as u32);
@@ -42,9 +51,18 @@ impl OpenProgram {
         }
     }
 
-    /// Name of the section this `Program` belongs to.
+    /// Name of the section this `OpenProgram` belongs to.
     pub fn section(&self) -> &str {
         &self.section
+    }
+
+    /// The name of this `OpenProgram`.
+    pub fn name(&self) -> Result<&str> {
+        let name_ptr = unsafe { libbpf_sys::bpf_program__name(self.ptr) };
+        let name_c_str = unsafe { CStr::from_ptr(name_ptr) };
+        name_c_str
+            .to_str()
+            .map_err(|e| Error::Internal(e.to_string()))
     }
 
     pub fn set_autoload(&mut self, autoload: bool) -> Result<()> {
@@ -88,6 +106,21 @@ impl OpenProgram {
     ///
     pub fn insn_cnt(&self) -> usize {
         unsafe { libbpf_sys::bpf_program__insn_cnt(self.ptr) as usize }
+    }
+
+    /// Gives read-only access to BPF program's underlying BPF instructions.
+    ///
+    /// Keep in mind, libbpf can modify and append/delete BPF program's
+    /// instructions as it processes BPF object file and prepares everything for
+    /// uploading into the kernel. So [`OpenProgram::insns`] and [`Program::insns`] may return
+    /// different sets of instructions. As an example, during BPF object load phase BPF program
+    /// instructions will be CO-RE-relocated, BPF subprograms instructions will be appended, ldimm64
+    /// instructions will have FDs embedded, etc. So instructions returned before load and after it
+    /// might be quite different.
+    pub fn insns(&self) -> &[libbpf_sys::bpf_insn] {
+        let count = self.insn_cnt();
+        let ptr = unsafe { libbpf_sys::bpf_program__insns(self.ptr) };
+        unsafe { std::slice::from_raw_parts(ptr, count) }
     }
 }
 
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/probe.rs
```diff
@@ -0,0 +1,57 @@
...
+//! Note that running many of these probes in a short amount of time can cause
+//! the kernel to reach the maximal size of lockable memory allowed for the
+//! user, causing subsequent probes to fail. In this case, the caller may want
+//! to adjust that limit with setrlimit().
+
+use super::map::MapType;
+use super::program::ProgramType;
+use libbpf_sys::bpf_func_id;
+
+use crate::*;
```
danielocfb, line 12: I understand there is precedent to the contrary here, but can we please stop using glob imports like that? It's fine for pulling in `super` definitions in a unit test module, but in the core library I'd prefer we list the imports explicitly, so that it's easier to follow where something is coming from or what it refers to. It also prevents naming conflicts from the get go. What do you think?

Also see note here: https://phaiax.github.io/mdBook/rustbook/ch07-03-importing-names-with-use.html#glob-imports-with-

libbpf-rs/src/probe.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Make sure the process has required set of CAP_* permissions (or runs as
+/// root) when performing feature checking.
+pub fn probe_prog_type(prog_type: ProgramType) -> Result<bool> {
+    let ret = unsafe { libbpf_sys::libbpf_probe_bpf_prog_type(prog_type as u32, std::ptr::null()) };
+    match ret {
+        0 => Ok(false),
+        1 => Ok(true),
+        _ => Err(Error::System(-ret)),
+    }
+}
```
danielocfb, line 26: I can see why you'd arrange functionality the way you did, but perhaps an alternative design would be to extend `ProgramType` with a method to query whether it is supported (similar for other types)? Then users would not need to import this functionality explicitly, it's just generally easier discoverable, and it would allow for much more explicit naming (e.g., `impl ProgramType { fn is_supported(&self) -> Result<bool> { ... } }` or `fn is_supported_on_host(&self)`; in addition to `is_helper_supported(&self, helper_id...) ...`).

I am not a fan of the probe terminology as the sole indication of what a function is doing, as it is ambiguous. Again, I can see where it's coming from, but I don't think we should sacrifice explicitness & proper integration (native feel) with the binding language, just to preserve design decision made in a different language and that users may have no awareness of.

How does that sound to you?

libbpf-rs/src/program.rs
```diff
@@ -86,6 +102,23 @@ impl OpenProgram {
...
         unsafe { libbpf_sys::bpf_program__insn_cnt(self.ptr) as usize }
     }
+
+    /// Gives read-only access to BPF program's underlying BPF instructions.
+    ///
+    /// Keep in mind, libbpf can modify and append/delete BPF program's
+    /// instructions as it processes BPF object file and prepares everything for
+    /// uploading into the kernel. So [`OpenProgram::insns`] and [`Program::insns`] may return
+    /// different sets of
+    /// instructions. As an example, during BPF object load phase BPF program
```
danielocfb, line 112: Why is the line broken in the middle?

libbpf-rs/src/program.rs
```diff
@@ -44,6 +52,14 @@ impl OpenProgram {
         &self.section
     }
 
+    pub fn name(&self) -> Result<&str> {
```
danielocfb, line 55: Would you mind documenting the method? Same comment for `prog_type` above.

danielocfb (changes requested):
I think it makes sense to integrate this functionality. Thanks for the pull request. I've left a few comments, please let me know if they make sense to you.

libbpf-rs/src/probe.rs
yuval-k: no prob, i just tried to follow existing style. will change

libbpf-rs/src/probe.rs
yuval-k: good idea - i'll give it a try

libbpf-rs/src/program.rs
yuval-k: bad copy past from libbpf; will fix

yuval-k:
hi @danielocfb updated the PR with your feedback. based on the contribution doc, i updated the existing commits and forced pushed.

libbpf-rs/src/program.rs
```diff
@@ -42,11 +51,20 @@ impl OpenProgram {
...
         &self.section
     }
 
+    /// The name of this `OpenProgram`.
+    pub fn name(&self) -> Result<&str> {
+        let name_ptr = unsafe { libbpf_sys::bpf_program__name(self.ptr) };
+        let name_c_str = unsafe { CStr::from_ptr(name_ptr) };
+        name_c_str
+            .to_str()
+            .map_err(|e| Error::Internal(e.to_string()))
```
danielocfb, line 65: Nit: Should we use `c_ptr_to_string` here?

libbpf-rs/src/program.rs
```diff
@@ -89,6 +107,21 @@ impl OpenProgram {
...
     }
+
+    /// Gives read-only access to BPF program's underlying BPF instructions.
+    ///
+    /// Keep in mind, libbpf can modify and append/delete BPF program's
+    /// instructions as it processes BPF object file and prepares everything for
+    /// uploading into the kernel. So [`OpenProgram::insns`] and [`Program::insns`] may return
+    /// different sets of instructions. As an example, during BPF object load phase BPF program
+    /// instructions will be CO-RE-relocated, BPF subprograms instructions will be appended, ldimm64
+    ///  instructions will have FDs embedded, etc. So instructions returned before load and after it
```
danielocfb, line 118: Nit: Remove space before `instructions`.

libbpf-rs/src/program.rs
```diff
@@ -496,4 +529,14 @@ impl Program {
     pub fn insn_cnt(&self) -> usize {
         unsafe { libbpf_sys::bpf_program__insn_cnt(self.ptr) as usize }
     }
+
+    /// Gives read-only access to BPF program's underlying BPF instructions.
+    ///
+    /// Please see note in [`OpenProgram::insns`].
+    ///
```
danielocfb, line 536: -`///`

danielocfb (approved):
Looks good overall, thanks again! The one thing I forgot to ask earlier was: can you please add a few tests for this functionality as well? I think then we can get it merged.

libbpf-rs/src/program.rs
yuval-k: we could; the main difference is that `c_ptr_to_string` returns a `Result<String>` and this returns `Result<&str>` which doesn't copy.
LMK which do you prefer.

danielocfb: Ah, true. No need to allocate. Let's keep what you have.

yuval-k:
Hi @danielocfb I believe I now addressed all comments. 
Note that I also added `map_type` and `name` to OpenMap (similar to what i did for OpenProgram)

danielocfb:
> Hi @danielocfb I believe I now addressed all comments. Note that I also added `map_type` and `name` to OpenMap (similar to what i did for OpenProgram)

Looks good to me. Thanks for seeing this through!

PR #296, opened 2022-12-22
Check documentation in CI

danielocfb:

### DIFF
--- a/libbpf-rs/src/ringbuf.rs
+++ b/libbpf-rs/src/ringbuf.rs
@@ -35,6 +35,7 @@ pub struct RingBufferBuilder<'a> {
 }
 
 impl<'a> RingBufferBuilder<'a> {
+    /// Create a new `RingBufferBuilder` object.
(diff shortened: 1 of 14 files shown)


libbpf-rs/src/ringbuf.rs
```diff
@@ -35,6 +35,7 @@ pub struct RingBufferBuilder<'a> {
 }
 
 impl<'a> RingBufferBuilder<'a> {
+    /// Create a new `RingBufferBuilder` not containing any callbacks.
```
anakryiko, line 38: "not containing any callbacks" seems like an unnecessary part? Builder is not a final RingBuffer handler, so obviously it starts with essentially empty set callbacks?

anakryiko (approved):
LGTM!

libbpf-rs/src/ringbuf.rs
danielocfb: I removed this part. Thanks for taking a look.

Issue #270, opened 2022-09-20
libbpf-cargo does not honor alignment of struct


zhaiyan920:
Hi, I've encountered this when using SkeletonBuilder: if the bpf code has "aligned" attribute on struct, then generated type does not correctly correspond to that.

I created a reproducible here: https://github.com/zhaiyan920/reproducible/tree/main/libbpf-cargo-not-honor-alignment, with libbpf-rs=0.18.0 and libbpf-cargo=0.12.0.

When running the build, it turns out

```
struct TestData {
  int data;
} __attribute__((aligned(16)));

struct TestData data[1];
```
is generated as
```
pub mod test_bss_types {
    #[derive(Debug, Copy, Clone)]
    #[repr(C)]
    pub struct bss {
        pub data: [TestData; 1],
    }
    #[derive(Debug, Default, Copy, Clone)]
    #[repr(C)]
    pub struct TestData {
        pub data: i32,
    }
}
```

BTF entry itself reflects the content correctly
```
[1] PTR '(anon)' type_id=0
[2] FUNC_PROTO '(anon)' ret_type_id=3 vlen=1
        'ctx' type_id=1
[3] INT 'int' size=4 bits_offset=0 nr_bits=32 encoding=SIGNED
[4] FUNC 'prog' type_id=2 linkage=global
[5] STRUCT 'TestData' size=16 vlen=1
        'data' type_id=3 bits_offset=0
[6] ARRAY '(anon)' type_id=5 index_type_id=7 nr_elems=1
[7] INT '__ARRAY_SIZE_TYPE__' size=4 bits_offset=0 nr_bits=32 encoding=(none)
[8] VAR 'data' type_id=6, linkage=global
[9] DATASEC '.bss' size=0 vlen=1
        type_id=8 offset=0 size=16 (VAR 'data')
```

So it seems like an issue in code generation.

danielocfb:
Hi @zhaiyan920 , thanks for the report. We have also hit this issue but haven't gotten around looking into a fix. Do you have the cycles to take this on, by any chance?

danielocfb:
@zhaiyan920 can you check if https://github.com/danielocfb/libbpf-rs/tree/topic/fix-align works for you?

zhaiyan920:
hi @danielocfb, sorry for missing previous message. I just briefly tested with my repo, pointing to your branch didn't address the issue. Here is what it generated 
`
struct TestData {
  int data;
} __attribute__((aligned(16)));
`
[...]

PR #297, opened 2022-12-22
Add necessary padding to overall struct type size

danielocfb:
When a C struct is defined with greater-than-natural alignment (by tagging it with the respective attribute), our emitted Rust types end up being incorrect and not necessarily matching kernel types. The problem is that while we support adding padding between members as necessary, we do not make sure to pad to the BTF defined type size (which is how alignment is conveyed via BTF).
Let's ensure we pad up to this size, to ensure better (well, hopefully perfect...) match up between generated types and kernel definitions.

Fixes: #270

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -880,6 +880,18 @@ impl Btf {
                         agg_content.push(format!(r#"    pub {}: {},"#, field_name, field_ty_str));
                     }
 
+                    if t.is_struct {
+                        let struct_size = t.size as usize;
+                        let padding =
+                            self.required_padding(offset, struct_size, type_id, packed)?;
+                        if padding != 0 {
+                            agg_content.push(format!(r#"    __pad_{offset}: [u8; {padding}],"#,));
+                            impl_default.push(format!(
+                                r#"            __pad_{offset}: [u8::default(); {padding}]"#,
+                            ));
+                        }
+                    }
+
                     if !gen_impl_default && t.is_struct {
                         writeln!(def, r#"#[derive(Debug, Default, Copy, Clone)]"#)?;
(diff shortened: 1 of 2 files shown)


danielocfb:
Grml...I forgot to adjust all those others tests...

libbpf-cargo/src/btf/btf.rs
```diff
@@ -880,6 +880,18 @@ impl Btf {
                         agg_content.push(format!(r#"    pub {}: {},"#, field_name, field_ty_str));
                     }
 
+                    if t.is_struct && !packed {
```
anakryiko, line 883: why `!packed` check? self.required_padding should take care of this?

not sure if it makes any difference, but it looks unnecessary and thus suspicious

danielocfb: Yeah, it's not strictly necessary. I had it there to make the fact that no padding is desired when a struct is packed more explicit, as it's not obvious otherwise. But given that `required_padding` takes `packed` as an input that's probably good enough. Removed.

anakryiko: >  I had it there to make the fact that no padding is desired when a struct is packed more explicit

You can have a packed struct where the last field ends way before the struct ends, so padding would still be necessary. E.g., try adding few more tests:

```
struct blah {
    char x;
    int y __attribute__((packed));
    long:64;
}
```

You can come up with some crazy cases. And if you manipulate BTF itself you can have even crazier ones.  Just look at https://patchwork.kernel.org/project/netdevbpf/patch/20221212211505.558851-7-andrii@kernel.org/ and the whole patch set.

Obviously, no one sane would be doing this explicitly in practice. But again, why making unnecessary simplifying assumptions, right?

anakryiko (approved):

PR #299, opened 2023-01-03
Add derive(Clone) to the structs in query.rs.

yuval-k:
This makes it easier to use these structs in other data structures.


danielocfb (approved):
Seems fine to me, thanks!

PR #301, opened 2023-01-05
Make `rustfmt` dependency optional

danielocfb:
Please see individual commits for description.


insearchoflosttime:
LGTM!

danielocfb:
FWIW, here is a skeleton generated without `rustfmt` applied to it: https://gist.github.com/danielocfb/1c1acd0612fb68f87e5803bf143afca7

I think it looks entirely reasonable. There are some whitespace errors, the program binary dump is occupying a single line, and we could certainly improve it in minor ways, but overall I'd definitely consider it readable. cc @anakryiko

PR #309, opened 2023-01-12
Hook up USDT opts

danielocfb:


libbpf-rs/src/program.rs
```diff
@@ -9,6 +10,28 @@ use strum_macros::Display;
 
 use crate::*;
 
+/// Options to optionally be provided when attaching to a USDT.
+#[derive(Clone, Debug, Default)]
+pub struct UsdtOpts {
+    /// Custom user-provided value accessible through `bpf_usdt_cookie`.
+    pub cookie: u64,
+    #[doc(hidden)]
+    pub _non_exhaustive: (),
```
insearchoflosttime, line 19: What does _non_exhaustive do? I see https://doc.rust-lang.org/reference/attributes/type_system.html,is this the same as marking the struct with the "#[non_exhaustive]" attribute?

danielocfb: It's just an "empty" (`()`) member that is undocumented and which conveys that more stuff may be added at any time. In short, the intention is to strongly urge you to use `Default` for initializing anything you don't care about. It's similar in intent to `#[non_exhaustive]`, yeah, but the latter has the problem that you prevent users entirely from creating an object based on initialization of individual members, as the documentation you linked states:

> Non-exhaustive variants ([struct](https://doc.rust-lang.org/reference/items/structs.html) or [enum variant](https://doc.rust-lang.org/reference/items/enumerations.html)) cannot be constructed with a [StructExpression](https://doc.rust-lang.org/reference/expressions/struct-expr.html)

```
error[E0639]: cannot create non-exhaustive struct using struct expression
   --> libbpf-rs/tests/test.rs:804:13
    |
804 | /             UsdtOpts {
805 | |                 cookie: cookie_val.into(),
806 | |                 ..UsdtOpts::default()
807 | |             },
    | |_____________^

For more information about this error, try `rustc --explain E0639`.
```

So that's really a deal breaker. I didn't see the point in introducing a complex builder infrastructure or something like that, so that seems like the best compromise.

We could leave it exhaustive, but it seems reasonable to be anticipating additions given that the underlying `libbpf` type does so as well.

danielocfb: It's similar to the pattern described here: https://rust-unofficial.github.io/patterns/idioms/priv-extend.html#alternative-private-fields-for-structs but given that we want constructability by clients, I think the member has to be made `pub`.

insearchoflosttime:
LGTM

PR #310, opened 2023-01-13
Provide Debug impl for remaining public types

danielocfb:
It is generally considered good practice to implement Debug on all public types of a crate [[0]][0] [[1]][1], in order to facilitate easier debugging by users.
Let's do just that for our remaining types that don't do it yet.

[0] https://rust-lang.github.io/api-guidelines/debuggability.html [1] https://rust-lang.github.io/api-guidelines/checklist.html

Closes: #260


insearchoflosttime:
Nice! LGTM

PR #317, opened 2023-01-18
Hook up uprobe attaching with additional opts

danielocfb:


libbpf-rs/src/util.rs
```diff
@@ -4,6 +4,12 @@ use std::path::Path;
 
 use crate::*;
 
+/// A type holding either a `L` or `R`.
+pub enum Either<L, R> {
```
insearchoflosttime, line 8: Is there a benefit to defining it here ourselves vs. using the either crate (https://docs.rs/either/1.1.0/either/enum.Either.html)?

libbpf-rs/src/program.rs
```diff
@@ -436,6 +489,36 @@ impl Program {
...
+        retprobe: bool,
+        pid: i32,
+        binary_path: impl AsRef<Path>,
+        func_offset: usize,
+    ) -> Result<Link> {
+        self.attach_uprobe_impl(
+            pid,
+            binary_path.as_ref(),
+            func_offset,
+            Either::Right(retprobe),
```
insearchoflosttime, line 505: I'm not sure if Either is a common Rust paradigm, but I think it's clearer from a readability standpoint if we just directly call "libbpf_sys::bpf_program__attach_uprobe()" here instead of having an attach_uprobe_impl() internal function.

libbpf-rs/src/util.rs
danielocfb: It's just another dependency to have publicly listed and care about, for what boils down to four lines of code. It did not seem worth it to me to pull it from the web. No other reason.

PR #318, opened 2023-01-18
Add CI step to build against minimum version of dependencies

danielocfb:

### DIFF
--- a/libbpf-rs/src/print.rs
+++ b/libbpf-rs/src/print.rs
@@ -83,7 +83,6 @@ extern "C" fn outer_print_cb(
 /// To pass all messages to the `log` crate:
 ///
 /// ```
-/// use log;
 /// use libbpf_rs::{PrintLevel, set_print};
 ///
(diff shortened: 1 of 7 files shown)


libbpf-rs/src/print.rs
```diff
@@ -83,7 +83,6 @@ extern "C" fn outer_print_cb(
 /// To pass all messages to the `log` crate:
 ///
 /// ```
-/// use log;
```
insearchoflosttime, line 86: I think we still need this line in order to use the log crate?

danielocfb: The crate is always available by virtue of it being listed in `Cargo.toml`. `use` only really makes sense if we were to import a specific function from it into our namespace. But the example references functions/macros by full path: `log::debug` etc. In any event, CI builds and [runs doctests](https://github.com/libbpf/libbpf-rs/actions/runs/3953874904/jobs/6770606319#step:7:290), so we can be sure it's not necessary.

Issue #224, opened 2022-05-06
`*mut perf_buffer` cannot be sent between threads safely within `PerfBuffer<'_>`


chengshuyi:
I want to create a thread to receive perf data and pass the data to the main thread for processing through crossbeam_channel. But there was trouble.

below is my code.
```c
thread::spawn(move || loop {
        perf.poll(Duration::from_millis(100)).unwrap();
    });
```

below is the error message.

```c
`dyn libbpf_rs::perf_buffer::SampleCb<for<'r> Output = ()>` cannot be sent between threads safely
the trait `Send` is not implemented for `dyn libbpf_rs::perf_buffer::SampleCb<for<'r> Output = ()>`
required because of the requirements on the impl of `Send` for `Unique<dyn libbpf_rs::perf_buffer::SampleCb<for<'r> Output = ()>>`
required because of the requirements on the impl of `Send` for `Unique<libbpf_rs::perf_buffer::CbStruct<'_>>`
required because of the requirements on the impl of `Send` for `Mutex<PerfBuffer<'_>>`
required because it appears within the type `[closure@src/main.rs:72:19: 74:6]`rustc[E0277](https://doc.rust-lang.org/error-index.html#E0277)
```

Maybe we can add this line of code to solve the problem, what do you think?
```rust
unsafe impl Send for PerfBuffer {}
```

Thanks in advance!

insearchoflosttime:
I don't think perf buffers can be transferred across thread boundaries because internally the perfbuffer's circular buffers are per-CPU, and there is no guarantee that the two threads will run on the same CPU (please correct me if I'm wrong here though @anakryiko). I don't think we can PerfBuffer can have the Send trait.

Ringbuffers though can be safely shared across multiple CPUs simultaneously. Maybe the ringbuffer would solve your use case instead of the perfbuffer? https://nakryiko.com/posts/bpf-ringbuf/ is a great article explaining the differences between the ringbuf and perfbuf.

anakryiko:
Both ringbuf and perfbuf can poll from any thread, as long as it's a single thread. So it should be possible to Send them to another thread. As long as Send implies a move, and not sharing? I'm forgetting exact semantics for Send, but there is no problem in polling for updates from any thread.

insearchoflosttime:
Ah okay, I think I'm confusing how perfbuffers work internally then. I'll take another look at the libbpf implementation. thanks for clarifying

plxty:
[...]

PR #319, opened 2023-01-18
libbpf-rs: Implement `Send` for `PerfBuffer` and `RingBuffer`

danielocfb:
As per https://github.com/libbpf/libbpf-rs/issues/224#issuecomment-1119996446 ring buffers and perf buffers can be polled from any thread. Hence, it is safe for us to implement `Send` for the corresponding Rust types, `RingBuffer` and `PerfBuffer`.

Closes: #224

### DIFF
--- a/libbpf-rs/src/perf_buffer.rs
+++ b/libbpf-rs/src/perf_buffer.rs
@@ -220,10 +220,30 @@ impl<'b> PerfBuffer<'b> {
     }
 }
 
-impl<'b> Drop for PerfBuffer<'b> {
+// SAFETY: `perf_buffer` objects can safely be polled from any thread.
+unsafe impl Send for PerfBuffer<'_> {}
+
+impl Drop for PerfBuffer<'_> {
     fn drop(&mut self) {
         unsafe {
             libbpf_sys::perf_buffer__free(self.ptr);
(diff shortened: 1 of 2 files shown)


libbpf-rs/src/perf_buffer.rs
```diff
@@ -220,10 +220,30 @@ impl<'b> PerfBuffer<'b> {
...
+
+impl Drop for PerfBuffer<'_> {
     fn drop(&mut self) {
         unsafe {
             libbpf_sys::perf_buffer__free(self.ptr);
         }
     }
 }
+
+#[cfg(test)]
```
insearchoflosttime, line 234: What are your thoughts on moving the tests to the libbpf-rs/tests directory?

danielocfb: I don't feel strongly about it, but in general I'd think it makes the most sense to have integration style tests (that test the public API surface) reside inside `libbpf-rs/tests`. Also see [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html). This one is arguably only a build time check anyway, so it feels most natural to keep it closest to the source code, at least to me. Happy to move it if that's what you prefer, though.

PR #321, opened 2023-01-19
Add license information to capable example

danielocfb:
It's not good style for us to have example code lying around without a license, to the point that it renders the example code completely useless, because it is unlikely to be safely reusable. This change proposes to license capable example under LGPL-2.1 OR BSD-2-Clause, similar to the main library.

### DIFF
--- a/examples/capable/Cargo.toml
+++ b/examples/capable/Cargo.toml
@@ -2,6 +2,7 @@
 name = "capable"
 version = "0.1.0"
 authors = ["Devasia Thomas <https://www.linkedin.com/in/devasiathomas>"]
+license = "LGPL-2.1 OR BSD-2-Clause"
 edition = "2021"
 
 [dependencies]


danielocfb:
@DevasiaThomas may I ask you to please take a look?

DevasiaThomas:
@danielocfb thanks for the tag 🙂.. I believe all ebpf example code in this repo are dual licensed in nature. The ebpf code itself is GPL 2.0 as indicated by the SPDX license identifier. The rust code takes on the license of libbpf-rs as it was contributed here. Since these examples are not meant to be packaged and shipped individually - is there a way we can have have the top level ``cargo.toml`` file contain and be the overarching license for all rust code in this repo? 

@anakryiko thoughts? Do you wish the examples be licensed separately?

danielocfb:
It's mostly about having that formally enshrined in all `Cargo.toml` files, as that is what some tooling flags otherwise. Depending on what lawyer you ask, you'll probably end up with different answers as to what supersedes what. So if in doubt, we should probably err on the side of being more explicit.

DevasiaThomas:
@danielocfb Alright then, LGTM!

Issue #177, opened 2022-01-11
Statically link BPF object files to avoid having DWARF info


jyn514:
As a summary of the conversation in https://github.com/libbpf/libbpf-rs/pull/169, the approach we're going to go with is statically linking BPF object files, which will strip out DWARF info as a side-effect

_Originally posted by @insearchoflosttime in https://github.com/libbpf/libbpf-rs/issues/163#issuecomment-1006820411_

javierhonduco:
This would be amazing. I am currently not checking in the generated skeleton in git due to this

PR #169, opened 2022-01-03
Use `-ffile-prefix-map` to get consistent debug info even across different users and machines

jyn514:
This is a more principled fix to #163 than #168. It avoids non-determinism from the tempdir where
the header files are stored, as well as differences from user home directories and project
directories.

Ideally, this would just omit the debug info altogether, but libbpf-sys gives a hard error if it's missing.
I tried stripping the object files with `-Wl,-s`, but it didn't remove the non-deterministic data.

Fixes #163. Closes #168.

Signed-off-by: Joshua Nelson <jnelson@cloudflare.com>

PR #323, opened 2023-01-20
libbpf-cargo: Strip DWARF information from generated object files

danielocfb:
Clang generated BPF object files may contain DWARF information that
references paths to temporary directories, for example. That ends up
rendering our generated skeleton unstable, as it potentially changes
between invocations [0]. That can be unfortunate if the skeleton is
meant to be checked into version control.
Andrii suggested [1] stripping said DWARF information by linking the
object file, which strips it as a side effect. This change does so.

[0] https://github.com/libbpf/libbpf-rs/issues/163
[1] https://github.com/libbpf/libbpf-rs/pull/169#issuecomment-1006074629

Closes: https://github.com/libbpf/libbpf-rs/issues/177

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -44,13 +44,14 @@ jobs:
         override: true
     - name: Install deps
       run: |
-        sudo apt-get install -y clang-14 libelf-dev zlib1g-dev
+        sudo apt-get install -y clang-14 libelf-dev zlib1g-dev linux-headers-$(uname -r)
+        sudo ln -s /usr/include/asm-generic /usr/include/asm
         sudo rm -f /bin/clang && sudo ln -s /usr/bin/clang-14 /bin/clang
     - name: Build
       run: cargo build --verbose --workspace --exclude runqslower
     - name: Run tests
       # Skip BTF tests which require sudo
-      run: cargo test --verbose --workspace --exclude runqslower -- --skip test_object --skip test_tc
+      run: cargo test --verbose --workspace --exclude runqslower -- --skip test_object --skip test_tc --include-ignored
     - name: Run BTF tests
       run: cd libbpf-rs && cargo test --verbose -- test_object test_tc
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -224,6 +224,7 @@ dependencies = [
  "cargo_metadata",
  "clap",
  "goblin",
+ "libbpf-rs",
  "libbpf-sys",
  "memmap2",
  "num_enum",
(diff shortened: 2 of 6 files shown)


danielocfb:
I verified that the added test failed beforehand.

danielocfb:
Grml...unsure where the failure is coming from, off hand. Will take another look.

danielocfb:
That being said, if there are comments on the approach feel free to post them.

danielocfb:
> ---- test::test_skeleton_builder_deterministic stdout ----
> clang is version 14.0.0
> Building /tmp/.tmpJeIbmy/proj/src/bpf/prog.bpf.c
> thread 'test::test_skeleton_builder_deterministic' panicked at 'called `Result::unwrap()` on an `Err` value: Build("Failed to compile obj=/tmp/.tmpu1nEWg/prog.o with status=exit status: 1\n stdout=\n \n stderr=\n In file included from /tmp/.tmpJeIbmy/proj/src/bpf/prog.bpf.c:2:\nIn file included from /usr/include/linux/bpf.h:11:\n/usr/include/linux/types.h:5:10: fatal error: 'asm/types.h' file not found\n#include <asm/types.h>\n         ^~~~~~~~~~~~~\n1 error generated.\n\n")', libbpf-cargo/src/test.rs:955:10

It appears as if the linux-headers package may be broken on Ubuntu? `/usr/include/linux/types.h` references `asm/types.h`, but that does not exist. What does exist is `/usr/include/asm-generic/types.h`. Both `/usr/include/linux/types.h` and `/usr/include/asm-generic/types.h` seem to be provided by the `linux-headers` package. There are other versions and variants of this virtual package available, but none seem to make a difference. Logically, it seems wrong for a header in a package to reference something outside of this package but not have that declared as an explicit dependency. Things work fine on my local system.
I also tried installing `linux-source` but that did not help.

I think it's probably not worth dealing with issues like that. So I think we should just make the test ignored. Then it won't be run by default but we can still do so on demand.

danielocfb:
I ended up symlinking `asm/` to `asm-generic/` and it works. Still keeping the test as ignored, but explicitly overwriting that in CI now. So we should be all good.

Some links touching on the mess:
https://www.mail-archive.com/redhat-list%40redhat.com/msg49603.html
https://stackoverflow.com/questions/65591871/asm-vs-asm-generic-in-linux-headers-are-they-same
https://stackoverflow.com/questions/14795608/asm-errno-h-no-such-file-or-directory

insearchoflosttime:
LGTM

PR #324, opened 2023-01-20
Enable dependency caching in CI

danielocfb:
Just stumbled over the Swatinem/rust-cache action which seems to be a low-key way for caching dependencies once built. As per my brief testing here [0] [1] the relevant runtime of the two jobs affected decrease from ~4:30 min down to ~3:30, if not better. Let's give it a try.

[0] https://github.com/danielocfb/libbpf-rs/actions/runs/3970473231 [1] https://github.com/danielocfb/libbpf-rs/actions/runs/3970565645

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,6 +46,7 @@ jobs:
       run: |
         sudo apt-get install -y clang-14 libelf-dev zlib1g-dev
         sudo rm -f /bin/clang && sudo ln -s /usr/bin/clang-14 /bin/clang
+    - uses: Swatinem/rust-cache@v2.2.0
     - name: Build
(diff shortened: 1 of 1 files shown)


insearchoflosttime:
Nice! How did you discover Swatinem's rust cache?

danielocfb:
> Nice! How did you discover Swatinem's rust cache?

I saw it being used [here](https://github.com/cobalt-org/liquid-rust/blob/b6e1d4e6288ce534b5ab1c59f618c99d63a51734/.github/workflows/ci.yml#L44), where I landed after reading https://www.reddit.com/r/rust/comments/8a24qk/dependabot_announcing_rust_support/ for trying to better understand the [Dependabot stuff](https://github.com/libbpf/libbpf-rs/pull/325) configuration.

PR #334, opened 2023-01-26
Ignore clippy lint warning for derive default

mimullin-bbry:
Rust version 1.62 made it possible to use #[derive(Default)] (see: https://github.com/rust-lang/rust/pull/94457)

As of rust version 1.68, the default clippy configuration will print a warning for manually impl Derive on enums. As such, users of libbpf-cargo will begin to experience clippy warnings when they migrate to version 1.68.

libbpf-cargo's minimum rust version is 1.58, thus the recommended method to remove the clippy warnings, is not yet an option for the project.

Add a #[allow(clippy::derivable_impls)], to suppress the clippy warning on all generated skeletons.
Add a TODO to remove this once the project moves to a minimum rust version of 1.62

### DIFF
--- a/libbpf-cargo/src/btf/btf.rs
+++ b/libbpf-cargo/src/btf/btf.rs
@@ -988,6 +988,9 @@ impl Btf {
 
                     // write an impl Default for this enum
                     if !t.values.is_empty() {
+                        // TODO: remove #[allow(clippy::derivable_impls)]
+                        //       once minimum rust at 1.62+
+                        writeln!(def, r#"#[allow(clippy::derivable_impls)]"#)?;
                         writeln!(def, r#"impl Default for {name} {{"#, name = t.name)?;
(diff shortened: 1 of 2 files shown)


danielocfb (commented):
Looks good to me, @mimullin-bbry . Thanks! I am just wondering, why did you decide not add the `allow` annotation to the generated definition/impl directly but applied it to pretty much everything generated?

mimullin-bbry:
> I am just wondering, why did you decide not add the `allow` annotation to the generated definition/impl directly but applied it to pretty much everything generated?

Because I didn't even think of that solution.  What you describe is much much better and I'll see if I can implement it instead.

danielocfb (approved):
Looks good to me. Thanks again! I did not specifically test that change, but did not find other paths where we generate an `enum` definition.

PR #336, opened 2023-01-26
libbpf-rs: Few changes to recent patches

danielocfb:
This patch changes a few nuisances from recent pull requests:
- Rename `Linker::add()` to `add_file` to follow libbpf naming more
  closely (I still kept `Linker::link()` (as opposed to finalize)) as it
  seems apt.
- Fix `SEC` used in `test_skeleton_builder_deterministic`. The program
  would no longer load with libbpf 1.0.
- Use `vmlinux.h` instead of `<linux/bpf.h>` for consistency with other
  skeletons.

### DIFF
--- a/examples/capable/build.rs
+++ b/examples/capable/build.rs
@@ -11,5 +11,5 @@ fn main() {
         .source(SRC)
         .build_and_generate(&out)
         .unwrap();
-    println!("cargo:rerun-if-changed={}", SRC);
+    println!("cargo:rerun-if-changed={SRC}");
 }
--- a/examples/capable/src/main.rs
+++ b/examples/capable/src/main.rs
@@ -167,7 +167,7 @@ fn _handle_event(opts: Command, event: capable_bss_types::event) {
 }
 
 fn handle_lost_events(cpu: i32, count: u64) {
-    eprintln!("Lost {} events on CPU {}", count, cpu);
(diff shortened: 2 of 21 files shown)


anakryiko (approved):
so this concern about failed linking step and maybe (in the future) still needing linker instance makes sense. Let's not make .link() take self and keep it as is. On libbpf side, calling bpf_linker__add_file after bpf_linker__finalize will keep returning -EINVAL, so no corruption or anything like that will happen.

Other than that, LGTM.

danielocfb:
> so this concern about failed linking step and maybe (in the future) still needing linker instance makes sense. Let's not make .link() take self and keep it as is. On libbpf side, calling bpf_linker__add_file after bpf_linker__finalize will keep returning -EINVAL, so no corruption or anything like that will happen.

Sounds good. I reverted that part of the change. An alternative would be to include `self` in the error variant that we return. I did not want to do that as it makes working with the method more cumbersome (because in addition to the instance we also get the actual error, but now we no longer can just propagate errors with `?`, for example), but it is what some other consuming Rust APIs do on error (and mostly if they do *not* have to include an additional error message).

danielocfb:
Rebased against https://github.com/libbpf/libbpf-rs/pull/334. Will merge once everything is green.

PR #338, opened 2023-01-26
Fix licenses

michel-slm:
- The LGPL license file had a typo, rename
- Fix symlinks in the libbpf-* crates
- Add missing license symlinks from examples
- Update `Cargo.toml` files; `LGPL-2.1` is a deprecated SPDX tag. Assuming the intent is `LGPL-2.1-or-later` for easier use by downstream projects

Signed-off-by: Michel Alexandre Salim <salimma@fedoraproject.org>


danielocfb (commented):
Thanks for the pull request! It seems fine in principle, but `LGPG-2.1` seems to be a deprecated alias for [GNU Lesser General Public License v2.1 **only**](https://spdx.org/licenses/LGPL-2.1.html), so I don't think we can just assume that the intent was `LGPL-2.1-or-later`, unfortunately.

michel-slm:
Ah, you're of course right. Let me rework. Since it's dual licensed with BSD, I think it does not matter if we stick to `-only` anyway.

danielocfb:
Awesome, thanks! One last question: is the separator used in the license file name of relevance? I see `LICENSE.BSD-2-Clause` uses a dot but `LICENSE-LGPL-2.1` a dash? Is that intentional or can we agree on a single one for the sake of consistency?

michel-slm:
@danielocfb that's... a typo when I scripted making the links, sorry. Updating

Issue #187, opened 2022-02-12
Cargo audit should be added to the github workflow


mimullin-bbry:
PR https://github.com/libbpf/libbpf-rs/pull/186 will give the libbpf-rs project a clean run of "cargo audit"

The project may wish to utilize cargo audit in the github workflow.

Using cargo audit, will give users of libbpf-rs piece of mind that the project isn't pulling in any potential vulnerabilities.

Cargo audit be run on a schedule (weekly/daily?) along with the regular PR request cycle in order to catch new vulnerabilities introduced by dependencies.

syscll:
https://github.com/actions-rs/audit-check could probably do the job.

You can also use the [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) trigger in GitHub actions to run it as a cron.

I can PR something if that helps?

mimullin-bbry:
Change #190 is certainly something that I, a user of libbpf-rs, would appreciate.

However, adding cargo audit as a scheduled event will add a maintenance burden upon the project maintainers.  I think it's a small burden w.r.t. scheduled events; but it's not my call to make.

To follow up #190, do you know if there is a way to add cargo audit to the CI pipeline, but make it such that it doesn't "fail" a PR build?  It would be good for reviewers of code to see if any newly added dependencies pull in potential security violations, but shouldn't interrupt project momentum for already known, but as-yet unaddressed violations.

syscll:
Apologies if I jumped the gun by creating the PR.

>However, adding cargo audit as a scheduled event will add a maintenance burden upon the project maintainers. I think it's a small burden w.r.t. scheduled events; but it's not my call to make.

If a scheduled event fails, it will only update the new shield in the README. But I do see how this could be added overhead to maintainers.

>To follow up https://github.com/libbpf/libbpf-rs/pull/190, do you know if there is a way to add cargo audit to the CI pipeline, but make it such that it doesn't "fail" a PR build?

Unfortunately, that isn't possible right now. There is an [open feature request](https://github.com/actions-rs/audit-check/issues/216) to add it, but as of right now, it would fail the PR status checks.

loshz:
I wanted to touch base on this again as I've been having good success with [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) recently, and it addresses all of the above issues we encountered before. In fact, [cargo-deny is now the official RustSec frontend](https://github.com/EmbarkStudios/cargo-deny/issues/194).

[...]

PR #340, opened 2023-01-27
Build with --locked in CI

danielocfb:
If we bump dependencies manually in `Cargo.toml` we may end up forgetting about adjusting `Cargo.lock` in the process, depending on how we go about it. That's not great for users as it will cause spurious changes when they build a checkout.
Luckily for us, it's an easy thing to detect: we should build with `--locked`, which will cause the build process to fail if the two files do not agree. Use it in our CI workflow accordingly.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -49,12 +49,12 @@ jobs:
         sudo rm -f /bin/clang && sudo ln -s /usr/bin/clang-14 /bin/clang
     - uses: Swatinem/rust-cache@v2.2.0
     - name: Build
-      run: cargo build --verbose --workspace --exclude runqslower
+      run: cargo build --locked --verbose --workspace --exclude runqslower
     - name: Run tests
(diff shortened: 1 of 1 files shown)


insearchoflosttime:
nice! out of curiosity, how did you discover the locked flag for cargo?

danielocfb:
> nice! out of curiosity, how did you discover the locked flag for cargo?

Thanks for the review! I think I had been aware of it for a while, but for this particular use case it may have been the [comments](https://github.com/libbpf/libbpf-rs/issues/187#issuecomment-1382580735) over in #187 that mentioned `--locked` that lead me into looking into its utility for our use cases a bit more and that made me recollect that 4bd2a132b234b7e798995427356e55ff6da13781 managed to forget updating `Cargo.lock` and so 52ab2505285099534e165776b58a03d7ee9e9715 included a bunch of unrelated changes more or less by accident.

PR #348, opened 2023-01-31
Perf enhancements

arilou:
Add some more low level API access to the perf buffer
1. Add ability to setup wakeup event count
2. Add low level API to access the underlying buffer, if one wants to implement a different draining mechanism than the default one in libbpf
3. Make Skel Sync


libbpf-rs/src/perf_buffer.rs
```diff
@@ -142,6 +161,49 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
         }
     }
 
+    unsafe extern "C" fn call_event_cb(
+        ctx: *mut c_void,
+        cpu: i32,
+        hdr: *mut libbpf_sys::perf_event_header,
+    ) -> i32 {
+        #[repr(C)]
+        #[derive(Debug, Default, Copy, Clone)]
```
danielocfb, line 170: Shall we minimize the number of derives to what is actually necessary?

libbpf-rs/src/perf_buffer.rs
```diff
@@ -142,6 +161,49 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
...
+        struct perf_sample_raw {
+            pub header: libbpf_sys::perf_event_header,
+            pub size: u32,
+            pub data: [u8; 0],
```
danielocfb, line 174: `pub` seems unnecessary. Same below.

libbpf-rs/src/perf_buffer.rs
```diff
@@ -104,29 +120,32 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
...
             lost_cb: self.lost_cb,
         }));
 
+        let mut attr = unsafe {
+            libbpf_sys::perf_event_attr {
+                type_: libbpf_sys::PERF_TYPE_SOFTWARE,
+                config: libbpf_sys::PERF_COUNT_SW_BPF_OUTPUT as u64,
+                sample_type: libbpf_sys::PERF_SAMPLE_RAW as u64,
+                __bindgen_anon_1: libbpf_sys::perf_event_attr__bindgen_ty_1 {
+                    sample_period: self.wakeup_events as u64,
```
danielocfb, line 134: Not really familiar with the perf event API and documentation seems non-existent: why are we setting a sampling period here? Can you please clarify and add a comment?

libbpf-rs/src/perf_buffer.rs
```diff
@@ -142,6 +161,49 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
...
+        }
+
+        match (*hdr).type_ {
+            libbpf_sys::PERF_RECORD_SAMPLE => {
+                let s = hdr as *mut perf_sample_raw;
+                let event_sample = &mut *s;
+                Self::call_sample_cb(
+                    ctx,
+                    cpu,
+                    std::ptr::addr_of_mut!(event_sample.data) as *mut c_void,
```
danielocfb, line 193: Can you please use `ptr::` for consistency? Here and for all other uses you introduce.

libbpf-rs/src/perf_buffer.rs
```diff
@@ -87,6 +91,18 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
             pages,
             sample_cb: self.sample_cb,
             lost_cb: self.lost_cb,
+            wakeup_events: self.wakeup_events,
+        }
+    }
+
+    /// Minimum amount of events before waking up
```
danielocfb, line 98: `amount` -> `number`?

Terminate comment with period?

libbpf-rs/src/perf_buffer.rs
```diff
@@ -142,6 +161,49 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
         }
     }
 
+    unsafe extern "C" fn call_event_cb(
```
danielocfb, line 164: Can you please add `#[deny(unsafe_op_in_unsafe_fn)]` here and then add individual `unsafe` blocks?

libbpf-rs/src/perf_buffer.rs
```diff
@@ -87,6 +91,18 @@ impl<'a, 'b> PerfBufferBuilder<'a, 'b> {
             pages,
             sample_cb: self.sample_cb,
             lost_cb: self.lost_cb,
+            wakeup_events: self.wakeup_events,
+        }
+    }
+
+    /// Minimum amount of events before waking up
+    pub fn wakeup_events(self, wakeup_events: u32) -> PerfBufferBuilder<'a, 'b> {
```
danielocfb, line 99: Seems inconsistent to accept `u32` here, given that the `buffer_buffer` API you add next accepts a `usize` index and then casts to `i32`, no?

libbpf-rs/src/perf_buffer.rs
```diff
@@ -282,6 +282,26 @@ impl<'b> PerfBuffer<'b> {
             unsafe { libbpf_sys::perf_buffer__buffer_fd(self.ptr, buf_idx as libbpf_sys::size_t) };
         util::parse_ret_i32(ret)
     }
+
+    pub fn buffer_buffer(&self, buf_idx: usize) -> Result<&mut [u8]> {
```
danielocfb, line 286: I don't believe this API is sound. If you call it multiple times with the same index you get multiple mutable references to the same area of memory, right?

Also, documentation it missing.

libbpf-cargo/src/gen.rs
```diff
@@ -785,6 +785,7 @@ fn gen_skel_contents(_debug: bool, raw_obj_name: &str, obj_file_path: &Path) ->
         }}
 
         unsafe impl<'a> Send for {name}Skel<'a> {{}}
+        unsafe impl<'a> Sync for {name}Skel<'a> {{}}
```
danielocfb, line 788: It would be great if you were to lay out your reasoning for this change. Also see [here](https://github.com/libbpf/libbpf-rs/blob/master/CONTRIBUTING.md#submitting-a-pull-request): "Include useful details in the commit messages". Thanks.

danielocfb (changes requested):
Left a couple of comments. Can you please make sure to add tests for the newly added functionality? Thanks.

libbpf-rs/src/perf_buffer.rs
anakryiko: @arilou if this is useful option, it probably would be better to add it on libbpf side into perf_buffer_opts, and then keep using non-raw variant in libbpf-rs. Would you be able to submit a patch to bpf@vger.kernel.org for this?

anakryiko (changes requested):
nit: you have typo in commit subject: PrefBufferBuilder 

more importantly, I think we should avoid libbpf-rs using perf_buffer__new_raw (and all the added code and complexity that comes with it), and instead keep both libbpf and libbpf-rs consistent in terms of perf_buffer options that are supported. This is a trivial change on libbpf side, let's just do it?

libbpf-rs/src/perf_buffer.rs
anakryiko: and the name should be just "buffer"?

libbpf-cargo/src/gen.rs
arilou: Done

libbpf-rs/src/perf_buffer.rs
arilou: Renamed to buffer, as for would you get multiple mutable references to the same area is true. i'll add a comment about that.

libbpf-rs/src/perf_buffer.rs
arilou: The reason it's u32 is because that's the type in libbpf
```
union {
    __u32       wakeup_events;    /* wakeup every n events */
    __u32       wakeup_watermark; /* bytes before wakeup   */
};
```

libbpf-rs/src/perf_buffer.rs
arilou: Done

libbpf-rs/src/perf_buffer.rs
arilou: Done

libbpf-rs/src/perf_buffer.rs
arilou: Done

libbpf-rs/src/perf_buffer.rs
arilou: I think I'll remove the wakeup_events patch from this PR and after ill send a patch to libbpf and libbpf-sys will move to where that patch is, I'll re-submit a new PR

libbpf-rs/src/perf_buffer.rs
arilou: Done

libbpf-rs/src/perf_buffer.rs
arilou: Done

arilou:
Thank you for the fast reply, I have added comments on top of yours and made the required changes, I will though following the request from @anakryiko remove the wakeup_events from this PR, and extend libbpf first in order
to avoid the usage of __raw API.

libbpf-rs/src/perf_buffer.rs
danielocfb: I understand. But it's also an implementation detail. It's obviously minor, but what I was trying to point out is that `buf_idx` as accepted [here](https://github.com/libbpf/libbpf-rs/pull/348/files#diff-753b0a913ef36046b385a6709fa7541513a1ace7a6bfa176fc55aae9c2c02c1aR228) is `usize` but it is cast to a `i32` for the call to `perf_buffer__buffer` as that is the internally used type. So if `libbpf` usage is the argument being made, why do it differently there but not here? Just trying to understand what rationale you apply.

I think either usage is a trade-off ultimately (one makes it harder to call while the other potentially gets us into trouble when casting) and so intuitively I'd treat both approaches as equally bad and find a third argument as a tiebreaker: in this case consistency. We expose `pages` as `usize` and then cast internally so it may be best to do the same here? Let me know what you think.

libbpf-rs/src/perf_buffer.rs
danielocfb: Sounds good, thanks!

libbpf-rs/src/perf_buffer.rs
danielocfb: But given that you removed the code in question I think we can resolve this issue :-)

libbpf-cargo/src/gen.rs
danielocfb: Thanks! Long term we should probably make `libbpf_rs::OpenObject` and `libbpf_rs::skeleton::ObjectSkeletonConfig<'a>` `Sync` (or at least convince ourselves that it would be a bad idea), but this seems reasonable.

libbpf-rs/src/perf_buffer.rs
```diff
@@ -218,6 +219,31 @@ impl<'b> PerfBuffer<'b> {
             unsafe { libbpf_sys::perf_buffer__buffer_fd(self.ptr, buf_idx as libbpf_sys::size_t) };
         util::parse_ret_i32(ret)
     }
+
+    /// Get a mutable access to the underlying per-cpu ring buffer data, this API can be used
+    /// to implement a different draining mechanisim for the perf buffer data.
+    ///
+    /// Notice: The API will return a mutable reference to the same memory region in case
+    ///         it's called more than once with the same index, this due to it's nature.
```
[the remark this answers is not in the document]
arilou: The reason I went with &self, as this would eventually have the same affect on the object as a call to consume/consume_buffer which both get &self and not &mut self.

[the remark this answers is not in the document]
danielocfb: Okay, but `consume`/`consume_buffer` are sound because they don't return mutable references to internal data, `buffer` is not because it does.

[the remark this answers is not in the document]
arilou: I think it will be wrong to make this API &mut self, I do see your point in marking it as an unsafe function.

I'll give you an example why I think this will make this API much harder to use, and in case we go down this road
it means we should probably implement all the other helpers Vec has like split_at_mut, I think it's better to think of this API like get_unchecked_mut, perhaps it should be renamed to buffer_unchecked_mut and marked as unsafe.

As for an example, let's pretend we have a structure called RingBuffer, which implements an iterator over the event in the perf buffer, by making this &self mut, it means you can only have a single instance of this RingBuffer, but you have N cores, so you want to create a RingBuffer structure per perf-buffer.

Will you be willing to accept renaming this API to buffer_unchecked_mut and marked as unsafe?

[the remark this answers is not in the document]
danielocfb: > As for an example, let's pretend we have a structure called RingBuffer, which implements an iterator over the event in the perf buffer, by making this &self mut, it means you can only have a single instance of this RingBuffer, but you have N cores, so you want to create a RingBuffer structure per perf-buffer.

Thanks for the example. I certainly can see why you would want this API, I am just trying to figure out if there is a better design possible. In my opinion and assuming I am understanding correctly, the more appropriate way to go about expressing the contract in code would be to have something like:

```rust
impl<'b> PerfBuffer<'b> {
  pub fn into_cpu_buffers(self) -> Vec<CpuBuffer<'b>>;
}
```
The problem will be that there is a bunch of infrastructure that you will need on `CpuBuffer` that currently only exists on `PerfBuffer` and would need to be duplicated in one form or another. In short, it's not trivial to retro-fit that design on the existing one and given that it's more or less a niche use (most users probably won't use that lower level infrastructure), it may not be worth going down that road.

> I think it's better to think of this API like get_unchecked_mut, perhaps it should be renamed to buffer_unchecked_mut and marked as unsafe.

Do keep in mind that `get_unchecked_mut` still has a `&mut self` receiver, which eliminates a slew of intricacies in its usage.

> Will you be willing to accept renaming this API to buffer_unchecked_mut and marked as unsafe?

That would definitely be better. We will need to slap a bunch of warnings on it. And again, a test case is still missing.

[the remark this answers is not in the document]
arilou: > Do keep in mind that get_unchecked_mut still has a &mut self receiver, which eliminates a slew of intricacies in its usage.

Ye, you are right my bad, that why i edited my original reply :P, but it sounds like you understand my point, as for your suggestion yes, that's possible but like you said it will still be a bit annoying as there are stuff that in PerfBuffer which would be useful to keep.

As for the name of the API if you will review my (edited reply) I suggest it would be buffer_mut since the word unchecked is usually meant for index checking, where here we still do them.
so it would look like
`pub unsafe fn buffer_mut(&self, buf_idx: usize) -> Result<&mut [u8]>`

What kind of test did you have in mind? and any chance we have some test I can just extend or use as a sample?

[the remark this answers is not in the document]
danielocfb: > As for the name of the API if you will review my (edited reply) I suggest it would be buffer_mut since the word unchecked is usually meant for index checking, where here we still do them.
> so it would look like
> `pub unsafe fn buffer_mut(&self, buf_idx: usize) -> Result<&mut [u8]>`

Yep, I think that sounds good.

> What kind of test did you have in mind? and any chance we have some test I can just extend or use as a sample?

Anything that demonstrates that `buffer_mut` works as expected will do. So I'd think that writing a value to a ring buffer from a BPF program and then scanning all buffers from user space to make sure it arrived somewhere, perhaps? Should be possible to distill parts of what `capable` is doing, for example, and enhance it a bit before putting it into `libbpf-rs/tests/test.rs`.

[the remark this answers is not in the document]
arilou: I have submitted a test patch, where I produce a know value and look for it in the perf buffer

libbpf-rs/tests/test.rs
```diff
@@ -1044,3 +1044,47 @@ fn test_object_link_files() {
...
+        .build()
+        .expect("Failed to build");
+
+    // Make the an action that the tracepoint will see
+    let _pid = unsafe { libc::getpid() };
+
+    let found_cookie = (0..perf.buffer_cnt()).into_iter().any(|buf_idx| {
+        let buf = unsafe {
+            perf.buffer_mut(buf_idx)
+                .expect("Failed to retreieve buffer")
```
danielocfb, line 1084: typo: `retrieve`

libbpf-rs/tests/test.rs
```diff
@@ -1044,3 +1044,47 @@ fn test_object_link_files() {
...
+    let map = obj.map("pb").expect("Failed to get perf-buffer map");
+
+    let cookie_bytes = cookie_val.to_ne_bytes();
+    let searcher = TwoWaySearcher::new(&cookie_bytes[..]);
+
+    let perf = libbpf_rs::PerfBufferBuilder::new(map)
+        .build()
+        .expect("Failed to build");
+
+    // Make the an action that the tracepoint will see
```
danielocfb, line 1078: typo: -the

libbpf-rs/Cargo.toml
```diff
@@ -38,6 +38,7 @@ probe = "0.3"
 scopeguard = "1.1"
 serial_test = "0.5"
 tempfile = "3.3"
+memmem = "0.1.1"
```
danielocfb, line 41: Can you please keep the list sorted? Thanks!

danielocfb (commented):
Looks great, thanks! Can you please rebase? It seems you targeted an old revision.

I have one more question, after getting one night's sleep over this proposal: What do you think about the following?
- instead of introducing `buffer_mut`, we expose:
```rust
    /// Retrieve the underlying [`libbpf_sys::perf_buffer`].
    pub fn as_libbpf_perf_buffer_ptr(&self) -> *mut libbpf_sys::perf_buffer {
        self.ptr
    }
```

This way, we avoid the the .5 cal footgun cocked and aimed at everyone's proverbial complete legs that circumvents all safeguards that Rust provides, you can still do the same thing, and we can sleep at night :-)

danielocfb:
> if someone wants they can always implement buffer_mut them self (I guess that's where you were going with this right?)

Exactly. Up to you, but feel free to leave the test case in, though, and just move the relevant code from the current `buffer_mut` directly into it. This way, we have some kind of example on how to use this API, which I suspect may be useful to users.

libbpf-rs/Cargo.toml
arilou: Done

libbpf-rs/tests/test.rs
arilou: Done

arilou:
I have revised to patches so now libbpf-rs only exposes the function you have suggested, and moved buffer_mut to the test only code.

Thanks and have a good weekend,
-- Jon.

danielocfb:
> I have revised to patches so now libbpf-rs only exposes the function you have suggested, and moved buffer_mut to the test only code.

Thanks! I merged it with a few changes.

PR #353, opened 2023-02-04
Add GitHub Actions workflow to publish new versions

danielocfb:

Here is a sample run: https://github.com/danielocfb/libbpf-rs/actions/runs/4088926223

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,24 @@
+name: Publish
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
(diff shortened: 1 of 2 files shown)


danielocfb:
> To clarify my understanding of git workflows: where would the publish workflow show up? Will this be under the "Actions" tab and then if you want to publish the new libbpf-rs and libbpf-cargo crates, you just press that button and it gets published?

Yep and yep. Please see https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow. Basically, a button appears there by virtue of us enabling `workflow_dispatch`. We push it stuff gets published (or so is the hope...)

PR #369, opened 2023-02-27
link: constructor from the pin path must be a  static method

vankoven:


danielocfb (approved):
Looks good to me. Thanks!

PR #370, opened 2023-02-27
link: pass program as mut reference to update_prog

vankoven:
`Program` is a re-usable object and must not be consumed while updating the `link`.

Follow-up for #369

When an eBPF program is loaded and attached by a custom loader based on `libbpf-rs`, one or more `Link` objects are created and pinned to bpffs. Once the loader is restarted to attach another program, it needs to load existing `Link` pins and call `update_prog()` method.
Since a single eBPF program can be assigned to more than one hook and since dropping of `Program` may invoke its unloading, the `update_prog()` must not consume the `Program` instance, but must use it as a mut reference. Just Like the underlying `libbpf-sys` function.

### DIFF
--- a/libbpf-rs/src/link.rs
+++ b/libbpf-rs/src/link.rs
@@ -41,7 +41,7 @@ impl Link {
     }
 
     /// Replace the underlying prog with `prog`.
-    pub fn update_prog(&mut self, prog: Program) -> Result<()> {
+    pub fn update_prog(&mut self, prog: &Program) -> Result<()> {
         let ret = unsafe { libbpf_sys::bpf_link__update_program(self.ptr, prog.ptr) };
         util::parse_ret(ret)
     }


danielocfb:
I agree that we should not consume the `Program`. Why do you think we need a mutable reference to it, though, @vankoven?

vankoven:
Good question, @danielocfb . I used `mut` reference here because of two reasons.

The first and most important point, `Link::update_prog()` calls `libbpf_sys::bpf_link__update_program()`, which uses `prog` as a `*mut Prog`. I borrowed' Program' as mutable to avoid a situation where the mutable raw pointer is created from a non-mutable reference.

Second, I took a look at the existing API of the `Program`. It has a huge family of `attach_*()` functions. All of those use mutable references to `Program`, even when underlying `libbpf_sys` functions don't require mutable raw pointers and no operations that change `Program` content happen.

Although the mutability of `Program` in `Program::attach*()` is not technically required, it looked like structures in `libbpf-rs` have "logical" mutability: they're borrowed as mutable every time we intend to change the behaviour of the object/program, including attaching and pinning. Thus I have followed the same concept in `update_prog()`. 

I may be wrong in the second case, but I am more confident in the first one.

anakryiko:
> Good question, @danielocfb . I used `mut` reference here because of two reasons.
> 
> The first and most important point, `Link::update_prog()` calls `libbpf_sys::bpf_link__update_program()`, which uses `prog` as a `*mut Prog`. I borrowed' Program' as mutable to avoid a situation where the mutable raw pointer is created from a non-mutable reference.

it should be `const struct bpf_program *prog` on libbpf side, it's just an omission, which we should fix upstream. There is no modification of underlying `struct bpf_program` in this operation.

Haven't thought much about `bpf_program__attach*()` and if they have to be mutable pointer/ref, but for `bpf_link__update_prog()` it definitely doesn't require mutability, even at conceptual level.

vankoven:
Thanks for the clarification! I have updated the PR and removed the mutability.

danielocfb (approved):

Issue #245, opened 2022-07-25
`*mut bpf_object` cannot be sent between threads safely


heinrich5991:
Is it intentional that structs like `WlistSkel` don't implement `Send` and `Sync`? Because they contain raw pointers, they're automatically marked as non-`Send`able and non-`Sync`hronized, but is it actually the case that one cannot use the data structures from multiple threads?

I'm trying to build a HTTP API around libbpf, and `warp` wants to use multiple threads to handle HTTP requests, thus the types would need to be shareable between threads. Otherwise, I'd have to spawn a separate thread that serializes and handles all the communication with libbpf, which would be less efficient.

Errors look like this:
```
error[E0277]: `*mut bpf_object` cannot be sent between threads safely
   --> src/main.rs:42:23
    |
42  |     let routes = root.or(get_ips_route).with(warp::cors().allow_any_origin());
    |                       ^^ `*mut bpf_object` cannot be sent between threads safely
    |
    = help: within `whitelist::wlist::WlistSkel<'_>`, the trait `Send` is not implemented for `*mut bpf_object`
    = help: the trait `warp::filter::FilterBase` is implemented for `warp::filter::map::Map<T, F>`
    = note: required because it appears within the type `Object`
note: required because it appears within the type `whitelist::wlist::WlistSkel<'_>`
   --> src/bpf/.output/wlist.skel.rs:188:12
    |
188 | pub struct WlistSkel<'a> {
    |            ^^^^^^^^^
    = note: required because of the requirements on the impl of `Sync` for `std::sync::Mutex<whitelist::wlist::WlistSkel<'_>>`
    = note: required because of the requirements on the impl of `Send` for `&std::sync::Mutex<whitelist::wlist::WlistSkel<'_>>`
    = note: required because it appears within the type `[closure@src/main.rs:30:65: 40:6]`
    = note: required because of the requirements on the impl of `warp::filter::FilterBase` for `warp::filter::map::Map<warp::filter::and::And<warp::filter::and::And<warp::filter::and::And<impl warp::filter::FilterBase<Extract = (), Error = Infallible> + warp::Filter + Copy, Exact<warp::path::internal::Opaque<__StaticPath>>>, impl warp::filter::FilterBase<Extract = (), Error = Rejection> + warp::Filter + Copy>, impl warp::filter::FilterBase<Extract = (), Error = Rejection> + warp::Filter + Copy>, [closure@src/main.rs:30:65: 40:6]>`
```

[...]

PR #374, opened 2023-03-08
Mark set_print as unsafe

mendess:
Unfortunately `set_print` cannot be safely wrapped as the underlying call is not thread safe, as documented in another `libbpf_` function [here](https://github.com/libbpf/libbpf/blob/cf46d44f0a06aa8b9400691ea3eb86ca4f066d5c/src/libbpf.h#L1602-L1607)

As such I think it needs to be marked `unsafe` for this library to be considered a safe wrapper around `libbpf`

I also deleted the "Temporarily disable printing" example as it seem unlikely that usage pattern is realistic outside the most trivial of programs.

### DIFF
--- a/examples/capable/src/main.rs
+++ b/examples/capable/src/main.rs
@@ -175,7 +175,12 @@ fn main() -> Result<()> {
 
     let mut skel_builder = CapableSkelBuilder::default();
     if opts.debug {
-        skel_builder.obj_builder.debug(true);
+        unsafe {
+            // SAFETY:
+            // no other thread is running which could cause undefined behaviour due
+            // to this call.
+            skel_builder.obj_builder.debug(true);
+        }
     }
 
     bump_memlock_rlimit()?;
--- a/examples/runqslower/src/main.rs
+++ b/examples/runqslower/src/main.rs
@@ -79,7 +79,12 @@ fn main() -> Result<()> {
 
     let mut skel_builder = RunqslowerSkelBuilder::default();
     if opts.verbose {
-        skel_builder.obj_builder.debug(true);
+        unsafe {
+            // SAFETY:
+            // no other thread is running which could cause undefined behaviour due
+            // to this call.
+            skel_builder.obj_builder.debug(true);
+        }
     }
 
     bump_memlock_rlimit()?;
--- a/examples/tproxy/src/main.rs
+++ b/examples/tproxy/src/main.rs
@@ -53,7 +53,12 @@ fn main() -> Result<()> {
 
     let mut skel_builder = TproxySkelBuilder::default();
     if opts.verbose {
-        skel_builder.obj_builder.debug(true);
+        unsafe {
+            // SAFETY:
+            // no other thread is running which could cause undefined behaviour due
+            // to this call.
+            skel_builder.obj_builder.debug(true);
+        }
     }
 
     // Set constants
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -32,14 +32,18 @@ impl ObjectBuilder {
 
     /// Option to print debug output to stderr.
     ///
+    /// # Safety
+    ///
+    /// See [`set_print`](set_print#safety).
+    ///
+    /// Also it is guarenteed that the only parts of the `ObjectBuilder` that interacts with libbpf
+    /// are the [`open_file`](Self::open_file) and [`open_memory`](Self::open_memory) methods.
+    ///
     /// Note: This function uses [`set_print`] internally and will overwrite any callbacks
     /// currently in use.
-    pub fn debug(&mut self, dbg: bool) -> &mut Self {
-        if dbg {
-            set_print(Some((PrintLevel::Debug, |_, s| print!("{s}"))));
-        } else {
-            set_print(None);
-        }
+    pub unsafe fn debug(&mut self, dbg: bool) -> &mut Self {
+        let callback = dbg.then(|| (PrintLevel::Debug, (|_, s| print!("{s}")) as PrintCallback));
+        unsafe { set_print(callback) };
         self
     }
 
--- a/libbpf-rs/src/print.rs
+++ b/libbpf-rs/src/print.rs
@@ -69,6 +69,9 @@ extern "C" fn outer_print_cb(
     0 // return value is ignored by libbpf
 }
 
+// allow main to reinforce the idea that set print should be used before any other interaction
+// with the library.
+#[allow(clippy::needless_doctest_main)]
 /// Set a callback to receive log messages from libbpf, instead of printing them to stderr.
 ///
 /// # Arguments
@@ -77,6 +80,12 @@ extern "C" fn outer_print_cb(
 ///   log message to handle, or `None` to disable all printing.
 ///
 /// This overrides (and is overridden by) [`ObjectBuilder::debug`]
+///
+/// # Safety
+///
+/// This function mutates a global variable without syncronization inside libbpf. This means that
+/// it is not thread safe and **cannot** be called concurrently with any other function
+/// in this library.
 ///
 /// # Examples
 ///
@@ -93,26 +102,26 @@ extern "C" fn outer_print_cb(
 ///     }
 /// }
 ///
-/// set_print(Some((PrintLevel::Debug, print_to_log)));
+/// fn main() {
+///     unsafe {
+///         // SAFETY: this is being setup before interacting with any other part of the library
+///         set_print(Some((PrintLevel::Debug, print_to_log)))
+///     };
+/// }
 /// ```
 ///
 /// To disable printing completely:
 ///
 /// ```
 /// use libbpf_rs::set_print;
-/// set_print(None);
+/// fn main() {
+///     unsafe {
+///         // SAFETY: this is being setup before interacting with any other part of the library
+///         set_print(None)
+///     };
+/// }
 /// ```
-///
-/// To temporarliy suppress output:
-///
-/// ```
-/// use libbpf_rs::set_print;
-///
-/// let prev = set_print(None);
-/// // do things quietly
-/// set_print(prev);
-/// ```
-pub fn set_print(
+pub unsafe fn set_print(
     mut callback: Option<(PrintLevel, PrintCallback)>,
 ) -> Option<(PrintLevel, PrintCallback)> {
     let real_cb: libbpf_sys::libbpf_print_fn_t = callback.as_ref().and(Some(outer_print_cb));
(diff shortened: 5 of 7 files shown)


anakryiko:
There is nothing unsfae about set_print just because it's not thread-safe, or am I missing something? Should any non-synchronized API be marked unsafe? That seems weird.

Please don't do this.

danielocfb:
Thanks for the pull request, @mendess! I don't believe I follow the argumentation in favor of making it `unsafe`, to be honest. The [link](https://github.com/libbpf/libbpf/blob/cf46d44f0a06aa8b9400691ea3eb86ca4f066d5c/src/libbpf.h#L1602-L1607) you reference talks about thread safety. That is expressed in Rust via the `Sync` (and to some degree `Send`) marker traits.

Unsafety, in the Rust sense, refers to violation of data race freedom and or memory safety (out of bounds accesses etc.). I don't believe either is possible here. Can you please clarify why you believe this to be necessary?

mendess:
> Should any non-synchronized API be marked unsafe?
if it mutates a global variable then, per rust's rules, yes

This api will mutate a global variable in C, global variable which will in turn be read by almost all the functions in `libbpf`. This means that effectively a mutable borrow and an immutable borrow of that same global variable can exist simultaneously, which violates Rust's rules. This is the reasoning I followed to arrive at the conclusion that this had to sadly be marked unsafe.

If this doesn't make sense please tell me where I made the mistake.

danielocfb:
> > Should any non-synchronized API be marked unsafe?
> > if it mutates a global variable then, per rust's rules, yes
> 
> This api will mutate a global variable in C, global variable which will in turn be read by almost all the functions in `libbpf`. This means that effectively a mutable borrow and an immutable borrow of that same global variable can exist simultaneously, which violates Rust's rules. This is the reasoning I followed to arrive at the conclusion that this had to sadly be marked unsafe.
> 
> If this doesn't make sense please tell me where I made the mistake.

The point is (as belabored at length [here](https://github.com/libbpf/libbpf-rs/issues/245#issuecomment-1460594559)) that the C stuff you are referring to effectively maps to Rust's [AtomicUsize](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html). And you can indeed use it safely in a global context: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=89e2b5d8074073e1f2727efa9810894c

mendess:
according to the [C standard](https://open-std.org/JTC1/SC22/WG14/www/docs/n1570.pdf), In section 5.1.2.4 "Multi-threaded executions and data races" you can read this in paragraph 25:
> The execution of a program contains a data race if it contains two conflicting actions in different threads, at least one of which is not atomic, and neither happens before the other. Any such data race results in undefined behavior.

In paragraph 4 you can get the definition of "conflict":
> Two expression evaluations conflict if one of them modifies a memory location and the other one reads or modifies the same memory location.

In paragraph 5 you can get the definition of "atomic operation", which doesn't include "word sized, aligned writes/reads".

So according to the standard it's undefined behavior to have a read and a write to the same location from different threads when no ordering is specified.

Now, I know that on x86/x64 architectures they might actually be atomic, but that only applies to the assembly generated, when you're writing C you're writing code for the C abstract machine, which has an optimizer that can reorder statements or even delete part of the code as I'm sure we all know. As such I believe you should always use true atomic operations when writing/reading to global memory locations.
If a function can invoke UB in C then it naturally follows that calling that function from rust can invoke UB in rust, as such, I decided it was best to mark it as `unsafe`.

danielocfb:
> As such I believe you should always use true atomic operations when writing/reading to global memory locations.

I agree (as I stated at least twice beforehand over in https://github.com/libbpf/libbpf-rs/issues/245), which is why I advise you to bring this up with `libbpf` folks. From my perspective, it has no bearing on `libbpf-rs`, however. Yes, this is a potential violation, but this library cannot and will not vouch for current implementations details of the C implementation and if there are bugs in said C implementation, they should be fixed there and not papered over here.

mendess:
> this library cannot and will not vouch for current implementations details of the C implementation
But this is not an implementation detail, the library itself states that `libbpf_` functions are not thread safe, thus, if we can't safely wrap them we should make them as unsafe.

> if there are bugs in said C implementation, they should be fixed there and not papered over here
I agree that libbpf bugs should not be papered over here but this isn't a `libbpf` bug, it's a documented part of the API. The only reason I dug around to know why it wasn't thread safe was just to help in understanding of the reasons behind the marking as `unsafe`.

danielocfb:
> > this library cannot and will not vouch for current implementations details of the C implementation
> 
> But this is not an implementation detail, the library itself states that `libbpf_` functions are not thread safe, thus, if we can't safely wrap them we should mark them as unsafe.

From https://github.com/libbpf/libbpf-rs/issues/245#issuecomment-1459025019:
So Andrii's stance seems to be that aligned pointer reads and writes are always atomic and so there is no issue. Personally, I think it should be an explicit atomic read and write, but I guess whatever. That reasoning would then presumably also apply to use_debugfs(). Nevertheless, if functionality is not explicitly green lit as being thread safe I'd be reluctant to expose it as Sync just because it currently is. That just invites subtle issues down the line.


> > if there are bugs in said C implementation, they should be fixed there and not papered over here
> 
> I agree that libbpf bugs should not be papered over here but this isn't a `libbpf` bug, it's a documented part of the API. The only reason I dug around the implementation details to know why it wasn't thread safe was just to help in understanding of the reasons behind the marking as `unsafe`.

So I agree that it appears seems as if [this comment](https://github.com/libbpf/libbpf/blob/cf46d44f0a06aa8b9400691ea3eb86ca4f066d5c/src/libbpf.h#L1602-L1607) and [this one](https://github.com/libbpf/libbpf-rs/pull/374#issuecomment-1460532163) contradict what was conveyed above. And I cannot reconcile the two without tea leaf reading (though I will say that I have a hunch...). So @anakryiko can you please confirm your intentions or clarify how I misrepresented what we discussed offline?

anakryiko:
> > > this library cannot and will not vouch for current implementations details of the C implementation
> > 
> > 
> > But this is not an implementation detail, the library itself states that `libbpf_` functions are not thread safe, thus, if we can't safely wrap them we should mark them as unsafe.
> 
> From [#245 (comment)](https://github.com/libbpf/libbpf-rs/issues/245#issuecomment-1459025019): So Andrii's stance seems to be that aligned pointer reads and writes are always atomic and so there is no issue. Personally, I think it should be an explicit atomic read and write, but I guess whatever. That reasoning would then presumably also apply to use_debugfs().

I don't mind making libbpf_set_print() using atomic exchange internally. Let's do that change. We can even say that it's thread-safe at that point, though libbpf won't provide any strict ordering guarantees between libbpf_set_print() and seeing new print callback on another thread. Either way application shouldn't rely on such tiny races for correctness.

> 
> > > if there are bugs in said C implementation, they should be fixed there and not papered over here
> > 
> > 
> > I agree that libbpf bugs should not be papered over here but this isn't a `libbpf` bug, it's a documented part of the API. The only reason I dug around the implementation details to know why it wasn't thread safe was just to help in understanding of the reasons behind the marking as `unsafe`.
> 
> So I agree that it appears seems as if [this comment](https://github.com/libbpf/libbpf/blob/cf46d44f0a06aa8b9400691ea3eb86ca4f066d5c/src/libbpf.h#L1602-L1607) and [this one](https://github.com/libbpf/libbpf-rs/pull/374#issuecomment-1460532163) contradict what was conveyed above. And I cannot reconcile the two without tea leaf reading (though I will say that I have a hunch...). So @anakryiko can you please confirm your intentions or clarify how I misrepresented what we discussed offline?

libbpf API not being thread-safe is a general statement about all the APIs. And that's true for most stuff (especially when we are talking about bpf_object and its methods, you definitely don't want to modify bpf_object state from two separate threads). But for stuff like `libbpf_set_print()`, yes, there is no synchronization, but there is also no inherent unsafety, as it's doing primitive aligned single-word reads/writes. As I said above, atomic exchange makes sense here as well, and we can even specify that this specific API is thread-safe. A bunch of feature-detection APIs can also be marked thread-safe. Please contribute patches to mark such APIs as thread-safe. As a general statement, though, all libbpf APIs are not thread-safe, unless explicitly marked otherwise, of course.

One set of APIs where libbpf-rs probably might want to enforce synchronization is `libbpf_register_prog_handler` and `libbpf_unregister_prog_handler`. Those are truly stateful and very non-atomic, and are not synchronized, so I'd prefer libbpf-rs to take some global lock for this pair of operations.

There might be few more APIs like this, though not very likely (I don't remember all of them off the top of my head, but we can audit all this). We can go audit them one by one and make a decision.

But for stuff like use_debugfs() and kernel_supports(), we don't need atomic operations. It works and will work correctly for the purposes it's being used for, even if some portion of C standard (sorry, not an expert on standards) declares some of the aspects as undefined behavior. If compilers somehow exploit such UBs, they will break, say, Linux kernel (there are lots of not explicitly atomic operations done there, see READ_ONCE/WRITE_ONCE), and they will just undo such "optimizations".

In any case, this is internal libbpf implementation details that won't lead to any crashes or misbehaviors, it's all "eventually consistent" caches. Worst case we'll repeat some feature detection on two parallel threads, which is not a bug or a problem. But this is all implementation details, and they are called from APIs that indeed are not thread-safe (as I mentioned, bpf_object manipulations are not, generally speaking, thread-safe).

But also note that it's a bit more nuanced still, as, say,  bpf_object__load() *can* be called by multiple parallel threads provided it's called for two independent instances of `struct bpf_object *`.

I agree we can document all that better on libbpf side, of course. As usual, patches are very welcome.

danielocfb:
Thanks Andrii. That confirms the mental model about `libbpf` I assumed above. So with respect to the safety/unsafety of `set_print`, I suggest we make `libbpf_set_print` use atomic instructions as mentioned. @mendess are you interested in contributing this change to `libbpf`? Otherwise I will put it on my plate.

Once that is done I believe we should be in agreement that there are no issues and that `set_print` is sound being a safe function, correct, @mendess?

mendess:
Yeah, sounds good to me. I would contribute but I've personally never contributed to anything in the Linux kernel tree (yet), so it might take a bit longer to get acquainted with the mailing list workflow. Feel free to do it if you want :)

danielocfb:
> Yeah, sounds good to me. I would contribute but I've personally never contributed to anything in the Linux kernel tree (yet), so it might take a bit longer to get acquainted with the mailing list workflow. Feel free to do it if you want :)

It may be a good chance to get acquainted with how things are done there, so if you have entertained the thought of contributing there, I'd be happy to leave it to you. From my perspective, we are not blocked on it, so take all the time you need. Just let me know how you decide.

anakryiko:
@danielocfb I have someone on Meta side interested in making relevant changes, so hopefully we'll get this fixed and better documented on libbpf side

PR #375, opened 2023-03-08
Remove undefined behavior from ObjectSkeletonConfig.

mendess:
According to the [documentation](https://doc.rust-lang.org/alloc/boxed/index.html#considerations-for-unsafe-code), any use of aliases of a box, be it raw pointers or references, is invalid after the box has been moved/mutated/borrowed, thus the box aliasing that was happening here was Undefined Behaviour. Since the only reason the box was kept was so that it would be dropped in the destructor, I removed it from the fields of the struct and then dropped it manually in the destructor.

I also added the one missing NonNull check. This will guarantee that bugs in the generated code fail faster since uninitialized pointers will be detected sooner from a panic instead of a seg-fault

### DIFF
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -187,8 +187,8 @@ impl OpenObject {
     ///     - points to a loaded `bpf_object`
     ///
     /// It is not safe to manipulate `ptr` after this operation.
-    pub unsafe fn from_ptr(ptr: *mut libbpf_sys::bpf_object) -> Result<Self> {
-        unsafe { Self::new(NonNull::new_unchecked(ptr)) }
+    pub unsafe fn from_ptr(ptr: NonNull<libbpf_sys::bpf_object>) -> Result<Self> {
+        unsafe { Self::new(ptr) }
     }
 
     /// Takes underlying `libbpf_sys::bpf_object` pointer.
--- a/libbpf-rs/src/skeleton.rs
+++ b/libbpf-rs/src/skeleton.rs
@@ -4,7 +4,7 @@ use std::boxed::Box;
 use std::ffi::CString;
 use std::mem::size_of;
 use std::os::raw::{c_char, c_ulong};
-use std::ptr;
+use std::ptr::{self, NonNull};
 
 use libbpf_sys::{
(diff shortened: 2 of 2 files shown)


danielocfb (commented):
Thanks for fixing this up! What you do seems fine in principle, however, it is not entirely clear to me why we need a box to begin with? Can't we just have a `obj: *mut *mut bpf_object` member added? What am I missing?

Also, could you please split this pull request into commits each with a single purpose and description? I am thinking one for each
- the `libbpf-rs/src/object.rs` change
- the `build_maps` receiver change
- the `build_progs` receiver change
- the rest

mendess:
@danielocfb done!

> Can't we just have a obj: *mut *mut bpf_object member added?
We can, but I don't see what is gained from having such a member, imho it makes the code harder to understand.

mendess:
I also wanted to tag @danobi here since they wrote the code I'm changing, just to make sure I didn't make any mistakes in my interpretation of the code.

danielocfb (approved):
Thanks @mendess ! Looks great to me now. Will give Dan a chance to take a look as well, as I agree this code is tricky.

danielocfb:
> > Can't we just have a obj: *mut *mut bpf_object member added?

(I meant to say `obj: *mut bpf_object` member, which I believe is what it would need to be)

> We can, but I don't see what is gained from having such a member, imho it makes the code harder to understand.

We wouldn't need a heap allocation and it would be clear that the memory location is associated with this object (no risk of leakage etc.). It would also curb any questions as to why we need a heap allocation to begin with. But I tend to agree it's not a clear cut improvement and I don't feel overly strongly ¯\_(°ペ)_/¯

mendess:
I think we need to heap allocate here from what I could understand from the C code, because this code is not using the C api as it was intended iirc so it has to cheat a bit. 

In the C version `s.obj` points to a field in a another struct, but this version doesn't use that other struct so it needs to allocate on the heap to act as a sort of fake "field of that struct". It's all really convoluted tbh so I didn't even dare investigate if I could get rid of the allocation.

danielocfb:
@danobi do you intend to take a look at this pull request? Otherwise we will go ahead and merge it.

danielocfb:
I have no other way to contact Daniel. I am going ahead with the merge given that the code has been reviewed and as we cannot be blocked on this indefinitely.

PR #378, opened 2023-03-09
Allow Map::update*() on shared reference

danielocfb:
Map updates are system calls and the kernel ensures safety in the context of multiple concurrent updates. Reflect that in the signatures of the `update*()` methods by allowing for a `&self` receiver. Requiring a mutable reference there leads to awkward usage patterns in certain scenarios.


danielocfb:
There actually are more than just the `update*()` methods that should work on a shared reference. Let me create a new pull request...

PR #379, opened 2023-03-09
Allow various Map system call wrappers to work on shared references

danielocfb:
Map updates in various shapes and forms are just system call wrappers and the kernel ensures safety in the context of multiple concurrent updates. As such, it makes sense for us to reflect that in the signatures of the corresponding methods by allowing for a &self receiver. Requiring a mutable reference there leads to awkward usage patterns in certain scenarios.


anakryiko (approved):
LGTM. I'm guessing original thinking was that conceptually we are indeed modifying map contents (even if it's in the kernel), but I agree, there is no need to force unnecessary restrictions when it's just a handle to kernel object

Issue #281, opened 2022-10-14
Releases


mdaverde:
Am wondering around the general appetite for more standard release practices. Actions such as:

- Git tags per version
- Github releases
- Changelog generation 

Any of these would reduce friction in determining what changed between crate releases

danielocfb:
I agree that it would be useful to include these bits. Do you have a suggestion on how to go about changelog generation?

PR #385, opened 2023-03-13
Enhance publish workflow

danielocfb:
A while back we added the publish workflow for publishing new releases of the workspace's crates to crates.io. This change enhances this workflow with the following logic:
1) we check that both the libbpf-rs and libbpf-cargo version are the
   same; we have decided to bump both in unison as that simplifies our
   lives
2) once we successfully published new versions of said crates, we create
   a git tag along with a GitHub release
   
 References: #281

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,10 +4,31 @@ on:
   workflow_dispatch:
 
 jobs:
+  version:
+    name: Check workspace member versions
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


danielocfb:
I triggered the workflow on my own fork and saw it create a GitHub release: https://github.com/danielocfb/libbpf-rs/actions/runs/4409873868

anakryiko (approved):
Fancy. :)

PR #387, opened 2023-03-15
Check package creation as part of publish workflow

danielocfb:
Currently publishing multiple crates in a workspace is not an atomic operation. That will hopefully change eventually on the Cargo side, but currently it means that we may end up publishing one crate version and then failing to publish the other in our workflow. To minimize the chances of that happening, let's include a cargo package run, which does some basic sanity checking, before evening attempting to publish.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,6 +38,8 @@ jobs:
         profile: minimal
         toolchain: stable
         override: true
+    - name: Dry-run package creation
+      run: cargo package --no-verify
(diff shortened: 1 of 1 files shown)


danielocfb:
I had a test run here: https://github.com/danielocfb/libbpf-rs/actions/runs/4431017309/jobs/7773539390 and it would have caused the issue fixed by https://github.com/libbpf/libbpf-rs/pull/386 before actually publishing anything.

anakryiko (approved):

PR #388, opened 2023-03-17
Expose underlying raw pointer in `Object` and `Map`

### LINKED ISSUE #360: How can I access `*const bpf_object`, `*const bpf_map` from corresponding Rust structs?
Like the title. I'm going to get the raw pointer and do some operations manually with `libbpf-sys`, but I didn't find any way to access underlying raw pointer in `Object` or `Map`

heyrutvik:
Closes #360

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -602,6 +602,11 @@ impl Map {
             Link::new(ptr)
         })
     }
+
+    /// Retrieve the underlying [`libbpf_sys::bpf_map`].
+    pub fn as_libbpf_bpf_map_ptr(&self) -> Option<NonNull<libbpf_sys::bpf_map>> {
+        self.ptr
+    }
 }
 
 bitflags! {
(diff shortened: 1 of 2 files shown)


libbpf-rs/src/map.rs
```diff
@@ -602,6 +602,11 @@ impl Map {
             Link::new(ptr)
         })
     }
+
+    /// Retrieve the underlying [`libbpf_sys::bpf_map`].
+    pub fn as_libbpf_bpf_map_ptr(&self) -> Option<NonNull<libbpf_sys::bpf_map>> {
```
anakryiko, line 607: curious why this one returns Option<> while as_libbpf_bpf_object_ptr() doesn't?

heyrutvik: See,
https://github.com/heyrutvik/libbpf-rs/blob/f9b3d9d01b467c1a561a49dc2c40d7a235926c0f/libbpf-rs/src/map.rs#L161-L162 

The `ptr` could be null if the map is created from userspace. I kept the return type `Option<_>` since I was not sure how to handle the call if `ptr` is not updated with correct value.

heyrutvik: @anakryiko Let me know if you have any suggestions about it.

PR #394, opened 2023-03-22
Use nightly rustfmt in CI

danielocfb:
I forgot to include that earlier, but with commit bf09e444d410 ("Use single line use statements") we include rustfmt configuration options that are only used on nightly.
With this change we adjust CI to use a nightly toolchain so that we can actually enforce the new style.


danielocfb:
I had that tested here: https://github.com/danielocfb/libbpf-rs/actions/runs/4494961074/jobs/7908080417

anakryiko (approved):

PR #397, opened 2023-03-27
Add apis that uses `bpf_map_info`

Officeyutong:
`libbpf` provides a api `bpf_obj_get_info_by_fd` which can get `bpf_map_info` (containing details of a certain map) only providing its file descriptor. 

This PR adds the rust wrapper of `bpf_obj_get_info_by_fd`, containing a struct `MapInfo` and some functions.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -2,17 +2,24 @@ use core::ffi::c_void;
 use std::convert::TryFrom;
 use std::ffi::CStr;
 use std::fmt::Debug;
+use std::os::unix::prelude::AsRawFd;
+use std::os::unix::prelude::BorrowedFd;
 use std::path::Path;
 use std::ptr;
 use std::ptr::null;
 use std::ptr::NonNull;
+use std::slice::from_raw_parts;
 
 use bitflags::bitflags;
+use libbpf_sys::bpf_map_info;
+use libbpf_sys::bpf_obj_get_info_by_fd;
 use nix::errno;
 use nix::unistd;
 use num_enum::IntoPrimitive;
 use num_enum::TryFromPrimitive;
 use strum_macros::Display;
+
+use std::mem;
 
 use crate::util;
 use crate::Error;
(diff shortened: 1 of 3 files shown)


danielocfb:
Hi @Officeyutong , thanks for the pull request. Can you please clean up the history? I seem to see a single relevant commit and then a bunch of fixes. Can they be combined?

Officeyutong:
> Hi @Officeyutong , thanks for the pull request. Can you please clean up the history? I seem to see a single relevant commit and then a bunch of fixes. Can they be combined?

I've just squashed the commits :)

libbpf-rs/src/map.rs
```diff
@@ -731,3 +734,117 @@ impl<'a> Iterator for MapKeyIter<'a> {
...
+/// Provide ability to retrive the details of a certain map with its fd only
+#[derive(Debug)]
+pub struct MapInfo {
+    info: bpf_map_info,
+}
+
+impl MapInfo {
+    /// Create a MapInfo struct from a fd.
+    pub fn new(fd: i32) -> Result<Self> {
+        let mut map_info = unsafe { std::mem::zeroed::<bpf_map_info>() };
```
danielocfb, line 747: Please use `use` declarations at the top: `use std::mem` (here and everywhere). Please also add a [`SAFETY` comment](https://std-dev-guide.rust-lang.org/documentation/safety-comments.html#inside-safe-elements).

libbpf-rs/src/map.rs
```diff
@@ -731,3 +734,117 @@ impl<'a> Iterator for MapKeyIter<'a> {
...
+    pub fn map_type(&self) -> MapType {
+        match MapType::try_from(self.info.type_) {
+            Ok(t) => t,
+            Err(_) => MapType::Unknown,
+        }
+    }
+    /// Get the map id
+    pub fn id(&self) -> u32 {
+        self.info.id
+    }
```
danielocfb, line 778: In my opinion we shouldn't be adding wrappers for each and every member of the C struct. Would it be sufficient to have another `as_libbpf_*` cast? Or just make the `info` part public. It's not like it's an implementation detail, really.

libbpf-rs/src/map.rs
```diff
@@ -731,3 +734,117 @@ impl<'a> Iterator for MapKeyIter<'a> {
...
+        let ret = unsafe {
+            bpf_obj_get_info_by_fd(
+                fd,
+                &mut map_info as *mut bpf_map_info as *mut c_void,
+                &mut size as *mut u32,
+            )
+        };
+        if ret < 0 {
+            return Err(Error::System(errno::errno()));
+        }
```
danielocfb, line 758: Please use `parse_ret_i32` helper.

libbpf-rs/src/map.rs
```diff
@@ -731,3 +734,117 @@ impl<'a> Iterator for MapKeyIter<'a> {
...
+/// Represent a wrapper `bpf_map_info` struct pointer
+/// Provide ability to retrive the details of a certain map with its fd only
+#[derive(Debug)]
+pub struct MapInfo {
+    info: bpf_map_info,
+}
+
+impl MapInfo {
+    /// Create a MapInfo struct from a fd.
+    pub fn new(fd: i32) -> Result<Self> {
```
danielocfb, line 746: Let's stop using "raw" file descriptors and use `std::os::fd::BorrowedFd` or `OwnedFd` here as applicable.

Officeyutong:
> I am reluctant to accept this functionality, as it's not clear to me why we need it, to be honest. What would be the use case for it? Along those lines, we should also add tests for it.

For example, I have a library that will perform map operations based on a user provided file descriptor. In this case I'll have to ensure the buffer have enough space for the map key and value (if the user want to lookup, or get next key, etc), so I need to know the key size and map size. The library can be seen [here](https://github.com/eunomia-bpf/wasm-bpf/blob/b76be32d44c2ec1933ca28eab875b50e713855b8/runtime/wasm-bpf-rs/src/bpf/map_operate.rs#L51). 

In the previous versions of the library mentioned, I don't know that the kernel and libbpf provide such abilities, so I just iterate over all maps and find the one matched the fd, and get its's key size and value size through the bpf_map*. After knowing that libbpf provodes a function `bpf_obj_get_info_by_fd` and the bpf_map_info structure, I just moved to that. I also think that other ones may also have similar needs, so I want to add `bpf_map_info` to libbpf-rs.

libbpf-rs/src/map.rs
Officeyutong: The fd may be not obtained from an bpf_map* struct, and even can be provided by the user. So maybe it's not a good idea to use wrapped file descriptors?

libbpf-rs/src/map.rs
Officeyutong: This function returns `Result<Self>`, not `Result<i32>`or `Result<usize>`. May be add a new `parse_ret_XXX` function?

Officeyutong:
> I am reluctant to accept this functionality, as it's not clear to me why we need it, to be honest. What would be the use case for it? Along those lines, we should also add tests for it.

Besides, I really appreciate the review you wrote! This is my first time contributing to libbpf-rs, so there's so much I'm not familiar with, so thank you!

libbpf-rs/src/map.rs
Officeyutong: I make `info` publicized now

libbpf-rs/src/map.rs
Officeyutong: Fixed now

libbpf-rs/src/map.rs
danielocfb: > The fd may be not obtained from an bpf_map* struct, and even can be provided by the user.

Yeah, but that's why I'd say it's important to make the contract clear whether it's consuming the file descriptor or not.

libbpf-rs/src/map.rs
```diff
@@ -734,3 +738,121 @@ impl<'a> Iterator for MapKeyIter<'a> {
...
+#[derive(Debug)]
+pub struct MapInfo {
+    /// The inner bpf_map_info object
+    pub info: bpf_map_info,
+}
+
+impl MapInfo {
+    /// Create a MapInfo struct from a fd.
+    pub fn new(fd: i32) -> Result<Self> {
+        // SAFETY: create a zeroed C stryct, and call a FFI function to fill it.
```
danielocfb, line 752: *struct

libbpf-rs/src/map.rs
```diff
@@ -734,3 +738,121 @@ impl<'a> Iterator for MapKeyIter<'a> {
...
+    pub fn map_type(&self) -> MapType {
+        match MapType::try_from(self.info.type_) {
+            Ok(t) => t,
+            Err(_) => MapType::Unknown,
+        }
+    }
+    /// Get the map id
+    pub fn id(&self) -> u32 {
+        self.info.id
+    }
```
danielocfb, line 785: Do we still need all those getters now that `info` is public?

danielocfb:
Can you please squash the changes properly as per the [contributor's guide](https://github.com/libbpf/libbpf-rs/blob/master/CONTRIBUTING.md) and fix CI?

danielocfb:
> > I am reluctant to accept this functionality, as it's not clear to me why we need it, to be honest. What would be the use case for it? Along those lines, we should also add tests for it.
> 
> For example, I have a library that will perform map operations based on a user provided file descriptor. In this case I'll have to ensure the buffer have enough space for the map key and value (if the user want to lookup, or get next key, etc), so I need to know the key size and map size. The library can be seen [here](https://github.com/eunomia-bpf/wasm-bpf/blob/b76be32d44c2ec1933ca28eab875b50e713855b8/runtime/wasm-bpf-rs/src/bpf/map_operate.rs#L51).
> 
> In the previous versions of the library mentioned, I don't know that the kernel and libbpf provide such abilities, so I just iterate over all maps and find the one matched the fd, and get its's key size and value size through the bpf_map*. After knowing that libbpf provodes a function `bpf_obj_get_info_by_fd` and the bpf_map_info structure, I just moved to that. I also think that other ones may also have similar needs, so I want to add `bpf_map_info` to libbpf-rs.

Gotcha. I guess we can include it. Thanks!

libbpf-rs/src/map.rs
Officeyutong: I removed the single-line getter now

libbpf-rs/src/map.rs
Officeyutong: fixed

libbpf-rs/src/map.rs
Officeyutong: > > The fd may be not obtained from an bpf_map* struct, and even can be provided by the user.
> 
> Yeah, but that's why I'd say it's important to make the contract clear whether it's consuming the file descriptor or not.

I'm using `BorrowedFd` now.

PR #402, opened 2023-03-29
Update phf dependency to 0.11

danielocfb:
Now that we use `1.63` as the minimum supported Rust version, we can bump `phf` up to `0.11`. In so doing we eliminate a few transitive dependencies while getting the newest and greatest.


anakryiko (approved):
what are we using this perfect hashing library for?...

danielocfb:
> what are we using this perfect hashing library for?...

The `capable` example seems to use it for a static map: https://github.com/libbpf/libbpf-rs/blob/master/examples/capable/src/main.rs#L25-L67

Issue #337, opened 2023-01-26
probably needs to pass -fno-stack-protector to clang


gyakovlev:
example build failure in https://github.com/facebookincubator/below/blob/45d53ebd9f44768654b1baa1be02b4f8e3eb9924/below/build.rs#L33
``` 
--- stderr                                                                                                                                                                                                                                                    
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Build("Failed to compile obj=/var/tmp/portage/sys-process/below-0.6.3/temp/.tmpptrY1G/exitstat.o with status=exit status: 1\n stdout=\n \n stderr=\n ./src/bpf/exitstat.bpf.c:57:5: er
ror: A call to built-in function '__stack_chk_fail' is not supported.\nu32 task_cpu(void *arg) {\n    ^\n./src/bpf/exitstat.bpf.c:72:5: error: A call to built-in function '__stack_chk_fail' is not supported.\nint tracepoint__sched__sched_process_exit(\n 
```

on this system (gentoo hardened) clang  and gcc are configured to pass
```
-fstack-clash-protection
-fstack-protector-strong
```

by default.
which doesn't work with BPF

https://lkml.org/lkml/2020/2/21/1000


so `SkeletonBuilder` (and maybe something else) needs to take care of this at the source I think, instead of downstreams adding it to build scripts.

WDYT?

gyakovlev:
another recent related change:
https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=878625e1c7a10dfbb1fdaaaae2c4d2a58fbce627

danielocfb:
Based on the links you provided, that flag seems like a reasonable thing to add to the `clang` invocation indeed. Do you intend to work on a patch to address the issue, @gyakovlev ?

I am thinking that perhaps it would would also be nice to honor the `CFLAGS` environment variable in order to provide some more control for users to unblock themselves for cases like this.

PR #409, opened 2023-04-04
Invoke `clang` with `-fno-stack-protector`

d-e-s-o:
Compilation of libbpf-cargo generated skeletons on a Gentoo system may fail, as stack protection may be enabled implicitly there, but it is not supported by BPF [0].
Make sure to invoke clang with -fno-stack-protector to prevent compilation failures.

[0] https://lkml.org/lkml/2020/2/21/1000

Closes: #337

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -265,11 +265,16 @@ pub fn build_single(
     check_clang(debug, &clang, skip_clang_version_checks)?;
     let header_parent_dir = tempdir()?;
     let header_dir = extract_libbpf_headers_to_disk(header_parent_dir.path())?;
-    let compiler_options = if let Some(dir) = &header_dir {
+    let mut compiler_options = if let Some(dir) = &header_dir {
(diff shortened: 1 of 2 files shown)


libbpf-cargo/src/build.rs
```diff
@@ -270,6 +270,11 @@ pub fn build_single(
     } else {
         options.to_string()
     };
+
+    // Explicitly disable stack protector logic, which doesn't work with
+    // BPF. See https://lkml.org/lkml/2020/2/21/1000.
+    let compiler_options = format!("{compiler_options} -fno-stack-protector");
```
mendess, line 276: Nit: avoid dropping the original options by mutating the original options. Also makes it more readable, but that's just my opinion

```suggestion
    compiler_options += " -fno-stack-protector";
```
or
```suggestion
    compiler_options.push_str(" -fno-stack-protector");
```

d-e-s-o: Sure thing, done!

anakryiko (approved):
makes sense

danielocfb:
Merged.

PR #410, opened 2023-04-04
Fix pointer mismatch causing build failure on `aarch64`

d-e-s-o:

See https://github.com/d-e-s-o/libbpf-rs/actions/runs/4602666687/jobs/8131885274 for failure.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,6 +97,36 @@ jobs:
           override: true
       - uses: Swatinem/rust-cache@v2.2.0
       - run: RUSTFLAGS="$RUSTFLAGS -L /usr/lib/x86_64-linux-gnu" cargo build --locked --package capable --features=static
+  build-aarch64:
+    name: Build for aarch64
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


anakryiko (approved):
TIL `*mut _` (type placeholder) trick!

As for arm64, I think libbpf is doing it somewhat differently (https://github.com/libbpf/libbpf/blob/master/.github/workflows/build.yml#L73-L91). See if you can use anything from libbpf's approach. We have more architectures than just x86-64 and arm64.

Either way, great improvement, ship it!

danielocfb:
> As for arm64, I think libbpf is doing it somewhat differently (https://github.com/libbpf/libbpf/blob/master/.github/workflows/build.yml#L73-L91). See if you can use anything from libbpf's approach. We have more architectures than just x86-64 and arm64.

Ah, I didn't know that `libbpf` had a cross compilation job, thanks for the pointer. I actually copied bits of it from `libbpf-sys`'s CI, actually (otherwise it would have taken me much longer to figure out the right incantation, especially given that it's all remote with a cycle time measured in minutes).

PR #411, opened 2023-04-04
Correctly cast *const c_char to *mut c_char without changing the type

mendess:
The code as it stands now (in master) doesn't compile on platforms where c_char is an alias to u8.

This bug could have been avoided by using [`pointer::cast_mut`](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast_mut) instead of manually casting the pointer, but back in the orginal PR we didn't want to raise the MSRV.

Is this enough of a reason to raise the MSRV to 1.65?

### DIFF
--- a/libbpf-rs/src/btf/mod.rs
+++ b/libbpf-rs/src/btf/mod.rs
@@ -22,6 +22,7 @@ use std::fmt::Display;
 use std::marker::PhantomData;
 use std::mem::size_of;
 use std::ops::Deref;
+use std::os::raw::c_char;
 use std::os::raw::c_void;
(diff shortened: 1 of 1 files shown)


danielocfb:
Thanks for the fix! I also have a change out addressing this issue here https://github.com/libbpf/libbpf-rs/pull/410

Is it okay to take that one @mendess , as it includes a regression test as well?

> Is this enough of a reason to raise the MSRV to 1.65?

I wanted to avoid doing that at this point to give consumers some more slack.

mendess:
Yep, sounds good! That one is better.

PR #412, opened 2023-04-05
Add utilities to work with pinned maps without going through `Object`

mendess:
In order to work with pinned maps or already loaded maps, you don't need the entire bpf_object, you can just interact with them by filesystem path or id. So I added some new constructors and getters for additional useful information.

This might have warranted different PRs, since the getters aren't 100% related to the main point of the PR (the constructors) but I need these getters and they seemed too small for a standalone PR. Hopefully that's okay.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -166,7 +166,7 @@ impl OpenMap {
 pub struct Map {
     fd: i32,
     name: String,
-    ty: libbpf_sys::bpf_map_type,
+    ty: MapType,
     key_size: u32,
     value_size: u32,
 
@@ -194,7 +194,8 @@ impl Map {
             return Err(Error::System(-fd));
         }
 
-        let ty = unsafe { libbpf_sys::bpf_map__type(ptr.as_ptr()) };
+        let ty = MapType::try_from(unsafe { libbpf_sys::bpf_map__type(ptr.as_ptr()) })
+            .unwrap_or(MapType::Unknown);
         let key_size = unsafe { libbpf_sys::bpf_map__key_size(ptr.as_ptr()) };
         let value_size = unsafe { libbpf_sys::bpf_map__value_size(ptr.as_ptr()) };
 
@@ -230,10 +231,7 @@ impl Map {
 
     /// Retrieve type of the map.
     pub fn map_type(&self) -> MapType {
-        match MapType::try_from(self.ty) {
-            Ok(t) => t,
-            Err(_) => MapType::Unknown,
-        }
+        self.ty
     }
 
     /// Key size in bytes
@@ -564,7 +562,6 @@ impl Map {
             None => (util::str_to_cstring("")?, "".to_string()),
         };
 
-        let map_type = map_type.into();
         let map_name_ptr = {
             if map_name_str.as_bytes().is_empty() {
                 null()
@@ -575,7 +572,7 @@ impl Map {
 
         let fd = unsafe {
             libbpf_sys::bpf_map_create(
-                map_type,
+                map_type.into(),
                 map_name_ptr,
                 key_size,
                 value_size,
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -208,6 +208,16 @@ impl Map {
         })
     }
 
+    /// Fetch extra map information
+    pub fn info(&self) -> Result<MapInfo> {
+        MapInfo::new(unsafe {
+            // SAFETY
+            // This BorrowedFd instance doesn't live longer than this scope, so it satisfies its
+            // invariants.
+            BorrowedFd::borrow_raw(self.fd)
+        })
+    }
+
     /// Retrieve the `Map`'s name.
     pub fn name(&self) -> &str {
         &self.name
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -1,10 +1,12 @@
 use core::ffi::c_void;
 use std::convert::TryFrom;
 use std::ffi::CStr;
+use std::ffi::CString;
 use std::fmt::Debug;
 use std::mem;
 use std::os::unix::prelude::AsRawFd;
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/map.rs
```diff
@@ -194,20 +196,34 @@ impl Map {
...
-            value_size,
+            ty: info.type_ as _,
+            key_size: info.key_size,
+            value_size: info.value_size,
+            max_entries: info.max_entries,
             ptr: Some(ptr),
         })
     }
 
+    fn info(fd: i32) -> Result<libbpf_sys::bpf_map_info> {
```
danielocfb, line 267: We just recently added the `MapInfo` type and it should be usable here instead, I would think.

https://github.com/libbpf/libbpf-rs/blob/30ae9cd0c8f5ba643f219432d9fb7778cc2db0e8/libbpf-rs/src/map.rs#L744-L767

libbpf-rs/src/map.rs
```diff
@@ -211,6 +214,56 @@ impl Map {
         })
     }
 
+    /// Open a previously pinned map from it's path.
```
danielocfb, line 217: it's -> its

libbpf-rs/src/map.rs
```diff
@@ -211,6 +214,56 @@ impl Map {
...
+        if p.last() != Some(&0) {
+            p.push(0);
+        }
+        let p = CString::from_vec_with_nul(p).expect("path contained null bytes");
+        let fd = parse_ret_i32(unsafe {
+            // SAFETY
+            // p is never null since we allocated ourselves.
+            libbpf_sys::bpf_obj_get(p.as_ptr())
+        })?;
+        Self::from_fd(fd)
```
danielocfb, line 232: Would you mind using the non-generic inner function pattern here to curb potential bloat? https://www.possiblerust.com/pattern/non-generic-inner-functions

libbpf-rs/src/map.rs
```diff
@@ -211,6 +214,56 @@ impl Map {
         })
     }
 
+    /// Open a previously pinned map from it's path.
+    ///
+    /// # Panics
+    /// If the path contains null bytes.
+    pub fn from_pinned_path<P: AsRef<Path>>(path: P) -> Result<Self> {
```
danielocfb, line 221: Would it be possible to add a test for this constructor so that we can see its usage?

danielocfb, line 221: Would it be possible to add a test for this constructor so that we can see its usage?

libbpf-rs/src/map.rs
```diff
@@ -211,6 +214,56 @@ impl Map {
...
+        let fd = parse_ret_i32(unsafe {
+            // SAFETY
+            // p is never null since we allocated ourselves.
+            libbpf_sys::bpf_obj_get(p.as_ptr())
+        })?;
+        Self::from_fd(fd)
+    }
+
+    /// Open a loaded map from it's map id.
+    pub fn from_map_id(id: u32) -> Result<Self> {
```
danielocfb, line 236: Similar comment to above: would be great to have a test that illustrates usage and checks workings.

libbpf-rs/src/map.rs
```diff
@@ -171,6 +171,7 @@ pub struct Map {
     key_size: u32,
     value_size: u32,
     max_entries: u32,
+    id: u32,
```
danielocfb, line 177: I am not opposed to this addition, just curious why you think it's needed? Given the `MapInfo` API we could always query it on demand, right? That seems preferable to duplicating members needlessly.

Similar comment about `max_entries`, actually.

libbpf-rs/src/map.rs
```diff
@@ -271,6 +273,11 @@ impl Map {
         self.max_entries
     }
 
+    /// Returns the map's id.
+    pub fn id(&self) -> u32 {
```
danielocfb, line 330: Let's tag it `#[inline]` given the brevity of the method and to make it eligible for inlining across crates.

danielocfb (commented):
Seems reasonable conceptually, but I think we should try and use the `MapInfo` API introduced [recently](https://github.com/libbpf/libbpf-rs/pull/397). Can you please check if it simplifies things to reduce duplication, @mendess ? Left a few more comments.

mendess:
I totally missed the `MapInfo` api. I reworked the entire PR so hopefully it's a lot better now.

One thing I went back and forth on is whether I should add getters like this
```rs
impl MapInfo {
    fn id(&self) -> u32 { self.info.id }
}
```
Reason being that, `self.info` is `pub` so the getter is redundant, on the other hand `map.info()?.info.id` looks a little bit silly. What do you think?

mendess:
I can't figure out why the tests are failing, it panics in the `bump_rlimit_mlock` function but I don't know what's different in my tests that could possibly cause that to fail in CI. It doesn't fail locally.

danielocfb:
> I totally missed the `MapInfo` api. I reworked the entire PR so hopefully it's a lot better now.
> 
> One thing I went back and forth on is whether I should add getters like this
> 
> ```rust
> impl MapInfo {
>     fn id(&self) -> u32 { self.info.id }
> }
> ```
> 
> Reason being that `self.info` is `pub` so the getter is redundant, on the other hand `map.info()?.info.id` looks a little bit silly. What do you think?

Yeah, I asked the submitter to remove the getters as these were 5+ one line access methods in an earlier revision of the patch and instead make the `map_info` member public. Let's leave it like this, I would say.

> I can't figure out why the tests are failing, it panics in the `bump_rlimit_mlock` function with a permission denied errno value but I don't know what's different in my tests that could possibly cause that to fail in CI. It doesn't fail locally.

There is nothing different other than the name, it's just that the other tests are not run without `root` access. That's ensured by the `--skip test_object` argument. I agree it's confusing and we should try and find a better way. Just add your test to the skip list for now or rename them to `test_object[...]`.

mendess:
@danielocfb done

danielocfb:
Looks great! I merged it with two minor changes. Thanks!

PR #419, opened 2023-04-12
Verify packages in workspace member directories

danielocfb:
Running cargo --no-verify in the root of the workspace checks packaging of all workspace members, including the examples.
We do not currently (and have no intention to) publish the examples, however. As such, let's verify only the members that we intend to publish. Also make use of Cargo's --package argument for the package and publish sub-commands we use, as that minimizes hard coded paths.


danielocfb:
Going ahead with the merge once CI passes as it's blocking the next release.

PR #431, opened 2023-04-26
program: support attach to BPF iterators.

WeiZhang555:
Add new function `attach_iter` to support BPF iter program, since iterator requires a map fd as attach options, general `attach` function can't work, we must add standalone `attach_iter` function.


libbpf-rs/src/program.rs
```diff
@@ -772,6 +772,28 @@ impl Program {
         )
     }
 
+    /// Attach this program to a
+    /// [BPF iter](https://www.kernel.org/doc/html/latest/trace/tracepoints.html).
+    /// The entry point of the program must be defined with `SEC("iter")` or `SEC("iter.s")`.
+    pub fn attach_iter(&mut self, map_fd: u32) -> Result<Link> {
```
danielocfb, line 778: We recently started using Rust's [`fd` abstractions](https://doc.rust-lang.org/std/os/fd/index.html), `BorrwedFd` and `OwnedFd`. Can we use them here please?

libbpf-rs/src/program.rs
```diff
@@ -772,6 +772,28 @@ impl Program {
         )
     }
 
+    /// Attach this program to a
+    /// [BPF iter](https://www.kernel.org/doc/html/latest/trace/tracepoints.html).
```
danielocfb, line 776: This link destination does not seem on topic. Is https://www.kernel.org/doc/html/latest/bpf/bpf_iterators.html more fitting?

danielocfb (commented):
Overall this seems like a reasonable addition, thanks! But can you please add a unit test for this functionality? It helps verify that everything is wired up correctly and illustrates usage.

libbpf-rs/src/program.rs
WeiZhang555: Yes, you're right. I copied/pasted a wrong link.

I will fix it in a few days once after I come back from  my vacation.

libbpf-rs/src/program.rs
WeiZhang555: Fixed. @danielocfb

danielocfb:
Thanks for the update! I merged the change with some tiny modifications.

WeiZhang555:
> Thanks for the update! I merged the change with some tiny modifications.

@danielocfb Thanks!

PR #433, opened 2023-04-27
 Add implementations of AsFd for Program 

chengshuyi:
This patch mainly implements AsFd for Program, and fd returns BorrowedFd. But I'm not so sure that the modifications of the examples make sense. Or maybe I should modify TcHookBuilder to accept the BorrowedFd parameter.

Best regards and thank you in advance.


chengshuyi:
There are some ci errors, after your review, I will modify them together, Thanks!

danielocfb:
Thanks for the pull request. I have two preliminary comments:
1) can you please use use traits from std::os::unix::io:: instead of std::os::fd:: so that we do not have to bump the minimum Rust version? It seems CI is rightly barking at us.
2) let's make sure that individual commits can be built properly. Currently it seems only the last commit allows for successful project compilation. Can you please adjust the history so that we don't have intermittent build breakages?

Will look through the core logic in more detail in a bit.

danielocfb (commented):
Other than the two aforementioned comments the changes look good. We can merge once addressed. Thanks!

chengshuyi:
> Thanks for the pull request. I have two preliminary comments:
> 
> 1. can you please use use traits from std::os::unix::io:: instead of std::os::fd:: so that we do not have to bump the minimum Rust version? It seems CI is rightly barking at us.

Ok.

> 2. let's make sure that individual commits can be built properly. Currently it seems only the last commit allows for successful project compilation. Can you please adjust the history so that we don't have intermittent build breakages?
> 

Ok, I'll merge them into one commit.

Thanks!

> Will look through the core logic in more detail in a bit.

danielocfb:
Thanks @chengshuyi , I added a CHANGELOG entry and merged the change!

PR #435, opened 2023-04-28
Include textual errno representation in Error::System Display impl

danielocfb:
An error such as "errno: 1" doesn't tell users particularly much. Provide more information by stringifying `errno` using `strerror_r` and including this information in the textual representation of the `Error::System` variant.

### DIFF
--- a/libbpf-rs/src/error.rs
+++ b/libbpf-rs/src/error.rs
@@ -1,3 +1,5 @@
+use std::ffi::CStr;
+use std::os::raw::c_char;
 use std::result;
 
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/error.rs
```diff
@@ -16,6 +18,90 @@ pub enum Error {
...
+
+    // TODO: Use `CStr::from_bytes_until_nul` once we are at Rust 1.69.
+    let cstr = unsafe { CStr::from_ptr(&buf as _) };
+    Some(String::from_utf8_lossy(cstr.to_bytes()).to_string())
+}
+
+/// Best-effort stringification of errno values to be included in textual
+/// representation of errors.
+fn error_suffix(errno: i32) -> String {
+    match (errno_to_str(errno), errno_to_text(errno)) {
```
anakryiko, line 81: while errno is always positive, errors returned from lots of libbpf APIs directly are negative. I found it useful in practice to not have to care/remember to negate error codes. So consider stringifying `abs(errno)` to handle both styles of error codes transparently

anakryiko (approved):

libbpf-rs/src/error.rs
danielocfb: Yeah, we have to make sure that we convert these return values correctly. I am not in favor of just papering over wrong conversions, to be honest.

PR #440, opened 2023-05-02
Improve btf_type_match! macro

mendess:
I got better at rust macros and so I came back to improve the `btf_type_match!` macro :smile: 

I think I covered pretty much every use case of a "real" `match` statement, thus making using this macro actually useful instead of annoying.

Please refer to all the tests cases I wrote to test the macro, the goal of the tests is to make sure the code compiles, hence the lack of asserts.


libbpf-rs/src/btf/types.rs
```diff
@@ -863,26 +844,333 @@ gen_collection_concrete_type! {
...
 /// ```compile_fail
 ///     BtfKind::Int => {
 ///         /* we don't have access to the variable, but we know the scrutinee is an Int */
 ///     }
 /// ```
+///
+/// Multiple Variants
+/// ```compile_fail
+///     BtfKind::Struct | BtfKind::Union => {
+///         /* we don't have access to the variable, but we know the scrutinee is either a Struct or a Union */
```
danielocfb, line 857: Nit: Let's not mix in C style comments and stick with `//` unless there is a good reason not to?

libbpf-rs/src/btf/types.rs
```diff
@@ -863,26 +844,333 @@ gen_collection_concrete_type! {
...
+            _ => false,
+        });
+        btf_type_match!(match ty {
+            BtfKind::Composite(c) => {
+                c.is_struct
+            }
+            _ => false,
+        });
+        // we ask rustfmt to not format this block so that we can keep the trailing `,` in the int
+        // branch.
```
danielocfb, line 1107: Comment seems misleading. There is no `int` branch.

libbpf-rs/src/btf/types.rs
```diff
@@ -863,26 +844,333 @@ gen_collection_concrete_type! {
...
+        dummy_type!(ty);
+
+        match ty.kind() {
+            crate::btf::BtfKind::Int => {
+                let i = crate::btf::types::Int::try_from(ty).unwrap();
+                std::hint::black_box(i);
+                "int"
+            }
+            _ => {
+                std::hint::black_box(());
```
danielocfb, line 1128: What purpose does this `black_box` call have here? It's not clear to me why we need it.

libbpf-rs/src/btf/types.rs
```diff
@@ -863,26 +844,333 @@ gen_collection_concrete_type! {
...
+                $action
+            }
+        }
+             $($rest)*
+        )
+    };
+    // default match case
+    //
+    // we only need the expression case here because this case is not followed by a $rest:tt like
+    // the others, which let's us use the $(,)? pattern.
```
danielocfb, line 974: Well, it could be ;-) Basically we support
```rust
        btf_type_match!(match ty {
            BtfKind::Int => "int",
            _ => "default",
        });
```
but not
```rust
        btf_type_match!(match ty {
            _ => "default",
            BtfKind::Int => "int",
        });
```
which I think is fair and there is no need, just pointing it out.

danielocfb (approved):
Looks good to me. Thanks for improving this macro! Left a few minor comments.

libbpf-rs/src/btf/types.rs
mendess: yeah that was on purpose, the second code will always generate a warning and in general be incorrect. Technically something the vanilla rust match supports but it seemed hardly worth supporting :grin:

libbpf-rs/src/btf/types.rs
mendess: I didn't want to use a println but wanted something that looked like a statement. I guess I could have used a simple assignment
```rs
let x = i;
x
```
instead of being all fancy with the black_box.

TL;DR: It serves no special purpose besides being a statement

danielocfb:
Merged. Thanks for the comprehensive set of tests covering the individual cases!

PR #441, opened 2023-05-05
Add several interfaces to restore bpf management data after user space process is restarted

yunbo-xufeng:
For high availability purpose, after user space process(e.g. the bpfloader agent) is restarted, we need to restore all the necessary bpf management data, such as map and program objects, so that we can cleanup properly when bpf programs are detached and unloaded.
For example, we store the bpf map id and program id into the disk filesystem, and after user space process restarted, it reads these data from persistent files and we reconstruct all the bpf management data from the map id, program id and others.
This patchset adds several interfaces to meet our current needs, and these interfaces generally read the bpf object information from the kernel.


danielocfb (changes requested):
Thanks for the pull request. I left a few comments. Can you please add tests for anything more than a trivial getter? Especially everything making `unsafe` calls into `libbpf`.

yunbo-xufeng:
> Thanks for the pull request. I left a few comments. Can you please add tests for anything more than a trivial getter? Especially everything making `unsafe` calls into `libbpf`.

Thanks for your review, I'll make the changes based on your comments and send a new pull request later.

libbpf-rs/src/program.rs
```diff
@@ -418,6 +421,34 @@ impl Program {
...
+        let ret = unsafe {
+            libbpf_sys::bpf_obj_get_info_by_fd(
+                fd.as_raw_fd(),
+                prog_info_ptr as *mut c_void,
+                &mut len,
+            )
+        };
+        if ret != 0 {
+            return Err(Error::Internal(format!("get bpf prog_info error: {}", ret)));
+        }
```
danielocfb, line 448: What's the reason for not using `parse_ret_*` functionality here? (as we do in `get_fd_by_id`)

libbpf-rs/src/program.rs
```diff
@@ -418,6 +421,34 @@ impl Program {
...
+        // A file descriptor coming from the bpf_prog_get_fd_by_id function is always suitable for
+        // ownership and can be cleaned up with close.
+        Ok(unsafe { OwnedFd::from_raw_fd(fd) })
+    }
+
+    /// Returns program id by fd
+    pub fn get_id_by_fd(fd: BorrowedFd) -> Result<u32> {
+        let mut prog_info = libbpf_sys::bpf_prog_info::default();
+        let prog_info_ptr: *mut libbpf_sys::bpf_prog_info = &mut prog_info;
+        let mut len = std::mem::size_of::<libbpf_sys::bpf_prog_info>() as u32;
```
danielocfb, line 438: Can you please use imports at the top instead of providing the full path here?

libbpf-rs/tests/test.rs
```diff
@@ -1254,3 +1254,22 @@ fn test_object_perf_buffer_raw() {
 
     assert!(found_cookie);
 }
+
+/// Check that we can get the map pin path from map object.
+#[test]
+fn test_map_get_pin_path() {
```
danielocfb, line 1263: Let's include this functionality into the existing `test_object_loading_pinned_map_from_path` test instead of creating a new one? It seems to fit fine in there, I'd say. Can we also include a call to `is_pinned`, please?

danielocfb (commented):
Seems good for the most part, I have a few more comments. Also, can you please make sure that CI passes with your changes? Otherwise we cannot land them.

libbpf-rs/src/map.rs
```diff
@@ -375,6 +375,41 @@ impl Map {
...
+                "map {} is not pinned",
+                self.name()
+            )));
+        }
+
+        let path_ptr = match self.ptr {
+            Some(ptr) => unsafe { libbpf_sys::bpf_map__pin_path(ptr.as_ptr()) },
+            None => return Err(Error::InvalidInput(("No map pointer found").to_string())),
+        };
+        let path_c_str = unsafe { CStr::from_ptr(path_ptr) };
```
danielocfb, line 409: Can't `bpf_map__pin_path` return `NULL`? It almost feels as if we shouldn't include any pinned checks here. `libbpf` should be the source of truth. So it would probably be better to return an `Option<&OsStr>`, I would say. What do you think?

yunbo-xufeng: > Can't `bpf_map__pin_path` return `NULL`? It almost feels as if we shouldn't include any pinned checks here. `libbpf` should be the source of truth. So it would probably be better to return an `Option<&OsStr>`, I would say. What do you think?

Yes, it's better to ignore pinned checks.

libbpf-rs/src/program.rs
yunbo-xufeng: > What's the reason for not using `parse_ret_*` functionality here? (as we do in `get_fd_by_id`)

I misunderstood the returned value(0 for success and otherwise fail), fixed.

yunbo-xufeng:
> Seems good for the most part, I have a few more comments. Also, can you please make sure that CI passes with your changes? Otherwise we cannot land them.

The CI fails at my added test case, but it's weird, it panicked at bump_rlimit_mlock(), do you have any thoughts?

danielocfb:
> > Seems good for the most part, I have a few more comments. Also, can you please make sure that CI passes with your changes? Otherwise we cannot land them.
> 
> The CI fails at my added test case, but it's weird, it panicked at bump_rlimit_mlock(), do you have any thoughts?

We ignore certain tests that require root when running the overall test suite and then run them from the `libbpf-rs/` directory (which uses `sudo`). See https://github.com/libbpf/libbpf-rs/blob/208d02e05b1ea436454f270b67ee01f32a629ae8/.github/workflows/test.yml#L61

That's based on name, currently. I suggest you start your new test with with `test_object_[...]` and it should work as expected. I haven't gotten around to overhauling this.

yunbo-xufeng:
> > > Seems good for the most part, I have a few more comments. Also, can you please make sure that CI passes with your changes? Otherwise we cannot land them.
> > 
> > 
> > The CI fails at my added test case, but it's weird, it panicked at bump_rlimit_mlock(), do you have any thoughts?
> 
> We ignore certain tests that require root when running the overall test suite and then run them from the `libbpf-rs/` directory (which uses `sudo`). See
> 
> https://github.com/libbpf/libbpf-rs/blob/208d02e05b1ea436454f270b67ee01f32a629ae8/.github/workflows/test.yml#L61
> 
> That's based on name, currently. I suggest you start your new test with with `test_object_[...]` and it should work as expected. I haven't gotten around to overhauling this.

Yes, now everything works fine, thank a lot for your suggestion!
BTW, I have added a new map pin test case because adding the test logic in the existing test_object_loading_pinned_map_from_path() involves convert map object from mutable to immutable.

danielocfb (approved):
Looks good to me, thanks!

Issue #447, opened 2023-05-08
`Program::attach_uprobe_with_opts()` fails when building with `--release`


0xB10C:
I've ran into an issue while deploying binary build with `cargo build --release` that attaches some uprobes to C++ functions. Locally, `cargo build` without `--release`, it works fine. To confirm the issue I've changed the libbpf CI to run with `--release` in https://github.com/0xB10C/libbpf-rs/commit/11b11f7d6dfd3055c8f5db2e54803d49437724b9. 

From [this](https://github.com/0xB10C/libbpf-rs/actions/runs/4916973854/jobs/8781435392#step:9:216) run:
```
...
libbpf: prog 'bpf_usdt_arg': relo #1: patched insn #9 (LDX/ST/STX) off 128 -> 128
libbpf: prog 'bpf_usdt_cookie': relo #2: <byte_off> [39] struct pt_regs.ip (0:16 @ offset 128)
libbpf: prog 'bpf_usdt_cookie': relo #2: matching candidate #0 <byte_off> [172] struct pt_regs.ip (0:16 @ offset 128)
libbpf: prog 'bpf_usdt_cookie': relo #2: patched insn #4 (LDX/ST/STX) off 128 -> 128
libbpf: bpf_uprobe_opts has non-zero extra bytes
thread 'test_object_uprobe_with_cookie' panicked at 'Failed to attach prog: Internal("bpf call \"libbpf_rs::program::Program::attach_uprobe_with_opts<std::path::PathBuf>::{{closure}}\" returned NULL")', libbpf-rs/tests/test.rs:1160:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::result::unwrap_failed
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/result.rs:1750:5
   3: core::ops::function::FnOnce::call_once
   4: core::ops::function::FnOnce::call_once
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

[...]

PR #450, opened 2023-05-15
Properly zero initialize uprobe opts `sz` member

danielocfb:
This change fixes the `sz` member initialization of `bpf_uprobe_opts` objects that we use in the uprobe attach code. The previously incorrect size could lead to libbpf failing the attachment on optimized builds, where the size of Program objects was different to dev ones.

Fixes: #447


anakryiko (approved):
nice find!

PR #451, opened 2023-05-17
Rename tests requiring root

danielocfb:
We have some complicated and non-obvious logic for skipping certain
tests when running cargo test from the root of the crate in CI, but then
an explicit invocation from the libbpf-rs/ sub-directory attempting to
run those skipped tests (now under sudo).
Simplify the logic and hopefully make it somewhat more obvious by
prefixing all tests that require root with `test_root_`.

Note: this is a workaround for what I'd call a bug in Cargo. I opened
https://github.com/rust-lang/cargo/issues/12147 in the hopes that this
will be addressed eventually (though I don't know how realistic that
is). If we are still getting confusion even with this more explicit
naming, we may want to consider moving (or copying) libbpf-rs's .cargo/
directory into the workspace root.


anakryiko (approved):
naming nit: "root" is a bit ambiguous (it could be interpreted as "root of file hierarchy", so "test_sudo_" would be better, IMO. But I don't feel strongly about this, so up to you.

danielocfb:
> naming nit: "root" is a bit ambiguous (it could be interpreted as "root of file hierarchy", so "test_sudo_" would be better, IMO. But I don't feel strongly about this, so up to you.

Sure, I changed the prefix.

PR #464, opened 2023-06-09
Add CRUD support for queue/stack map

### LINKED ISSUE #463: Allow manipulating map of type queue or stack
Hello,

I'm rewriting a cli implemented with Go based on cilium/ebpf to Rust, who applies firewall rules to container at cgroup level and monitoring the container's network flow. 
The latter feature engages a map type `queue` allowing the bpf program to record all ingress/egress info of the container which can be popped out to stdout by launching the cli.

But it seems like the lookup_and_delete method in libbpf_rs::Map doesn't yet support the operation on queue/stack, as mentioned in the comment and the outcome of my attempts.

ZhengjunHUO:
Fixes #463

According to the bpf [doc](https://www.kernel.org/doc/html/latest/bpf/map_queue_stack.html#userspace), for queue/stack map the userspace methods: bpf_map_update_elem(), bpf_map_lookup_elem(), bpf_map_lookup_and_delete_elem() take a nil pointer for key, and i did some changes in the corresponding methods in map.rs to make my scenario works as expected. (the repetition doesn't cool, it would be great if we could refacto the key checking part a bit).
- I added test in test.rs
- A working example is [here](https://github.com/ZhengjunHUO/ctnctl-rs/blob/main/src/actions.rs#L207)


libbpf-rs/src/map.rs
```diff
@@ -455,10 +455,18 @@ impl Map {
 
         let mut out: Vec<u8> = Vec::with_capacity(out_size);
 
+        let mut map_key = key.as_ptr() as *const c_void;
+        if self.key_size() == 0
+            && (matches!(self.map_type(), MapType::Queue)
+                || matches!(self.map_type(), MapType::Stack))
+        {
```
danielocfb, line 462:
```suggestion
if self.key_size() == 0 && matches!(self.map_type(), MapType::Queue | MapType::Stack) {
```

danielocfb: ^ here and elsewhere

libbpf-rs/src/map.rs
```diff
@@ -455,10 +455,18 @@ impl Map {
...
+            map_key = std::ptr::null();
```
danielocfb, line 463: Can you please import at top level?
```rust
use std::ptr;
```

Here and elsewhere.

libbpf-rs/src/map.rs
danielocfb: In general, would it be possible to introduce a function `map_key` that contains the newly added logic for map key inference?

libbpf-rs/tests/bin/src/cgroup-fw.bpf.c
```diff
@@ -0,0 +1,153 @@
...
+	s.port = p.dport;
+	isBannedL3 = bpf_map_lookup_elem(&ingress_blacklist, &iphd->saddr);
+	isBannedL4 = bpf_map_lookup_elem(&ingress_l4_blacklist, &s);
+    }else{
+        bpf_printk("Egress to %lu",iphd->daddr);
+	s.addr = p.daddr;
+	s.port = p.dport;
+	isBannedL3 = bpf_map_lookup_elem(&egress_blacklist, &iphd->daddr);
+	isBannedL4 = bpf_map_lookup_elem(&egress_l4_blacklist, &s);
+    }
```
danielocfb, line 128: Please don't mix tabs and spaces. Also, please fix formatting around `else`.

libbpf-rs/tests/bin/src/cgroup-fw.bpf.c
```diff
@@ -0,0 +1,153 @@
+#include <linux/bpf.h>
```
danielocfb, line 1: Thanks for adding this test object! It seems fine conceptually, but can we please boil it down to the essence? This seems more like an example than something we want in a basic unit test. E.g., I don't think we need all the filtering logic. It's probably easier to just use `__u32` values instead of 16 byte `pkt` ones.

Perhaps we don't need a new object altogether and can just add a queue/stack to, say, `tracepoint.bpf.c`? What do you think?

libbpf-rs/tests/test.rs
```diff
@@ -370,6 +370,55 @@ fn test_sudo_object_map_empty_lookup() {
         .is_none());
 }
 
+#[test]
+fn test_sudo_object_map_queue_crud() {
```
danielocfb, line 374: Let's add a summary comment what the test does?

danielocfb (commented):
This seems useful, thanks! I left a few comments. Can we simplify the testing infrastructure?

libbpf-rs/src/map.rs
ZhengjunHUO: hello @danielocfb, appreciate for the prompt review ! Yes I should have done that to avoid the repetition, it's fixed in the latest push.

libbpf-rs/src/map.rs
ZhengjunHUO: seems that it's already import at the top level, I did the change in the code.

libbpf-rs/tests/bin/src/cgroup-fw.bpf.c
ZhengjunHUO: tks to point out ! I removed this part as you advised below, but I'll apply a formatter in another repo where this file  comes from.

libbpf-rs/tests/bin/src/cgroup-fw.bpf.c
ZhengjunHUO: my bad, I just copied the file from the project I'm working on. I added a queue map and a stack map to the tracepoint.bpf.c

libbpf-rs/tests/test.rs
ZhengjunHUO: agree with that, I extended the test to cover the stack, with comment.

danielocfb (approved):
Looks good to me now, thanks! Will add a CHANGELOG entry and merge with a few minor changes.

danielocfb:
Split the test and merged the change. Thanks for your contribution!

PR #473, opened 2023-06-13
Split Maps into Maps and MapHandles (fixes #448)

### LINKED ISSUE #448: Allow for cloning or owning of major data types
Hello,

This is more of an API design question, though I couldn't find any contact information for a discussion. Thus I'm writing this as an issue.

I'm currently evaluating libbpf-rs for production usage. I've previously looked at [Aya](https://github.com/aya-rs/aya), but it's lacking BTF support - which is increasingly becoming more and more important.

While trying out a prototype conversion from Aya to libbpf-rs, the most pain I have is that libbpf-rs uses a lot of references. For many structs, it appears impossible to obtain ownership of them.

For example, if I want to work with a map, I can call `libbpf_rs::Object::map_mut()`, which gives me a mutable reference to that map. I can then add, remove or lookup elements. This is fine for small projects, i.e. if all of my program logic is in `main()`.

However, on a larger codebase, I'm working with a number of maps and I possibly want references to multiple maps at the same time. Due to Rusts borrow checker, this is impossible with libbpf-rs: I cannot call `map_mut()` twice (without dropping the map reference first), since that would require two mutable borrows of the Object.

This makes working with maps really painful, as I'm pretty much forced to pass around an `Arc<Mutex<Object>>`-style object to have multiple ownership of the Object and then get temporary references to the maps and drop them immediately afterwards.

I found a small hack to circumvent this: I can effectively clone a map by doing something like this: `libbpf_rs::Map::from_map_id(object.info()?.info.id)?` (i.e. creating a new Map from an existing map ID). This feels like a hack around the underlying problem though - if this were intended API usage, Map should implement `Clone` by itself. The omission of Clone on pretty much everything feels like it was intentional. But I'm having trouble on designing a clean API around libbpf-rs, since it's heavy usage of references makes it difficult to juggle multiple programs, maps and stuff around.

I've seen that libbpf-rs currently has this statement in its code for the Bpf object:

```
/// An `Object` is logically in charge of all the contained [`Program`]s and [`Map`]s as well as
/// the associated metadata and runtime state that underpins the userspace portions of BPF program
/// execution. As a libbpf-rs user, you must keep the `Object` alive during the entire lifetime
/// of your interaction with anything inside the `Object`.
///
/// Note that this is an explanation of the motivation -- Rust's lifetime system should already be
/// enforcing this invariant.
```

This makes complete sense. By only allowing references we can ensure that i.e. maps do not outlive the BPF object. However, it also disallows a lot of API designs and removes flexibility.

Aya for example also uses a similar approach by default. However, Aya has functions like `take_map` that return owned maps. Logically this transfers ownership from the Bpf Object to the caller, who is now responsible for managing the map's lifetime: Once the map struct is dropped, the map is gone forever. This allows for complex programs, where different modules can be in charge of a given map, instead of centralizing everything around a single Bpf Object.

GermanCoding:
So, this is a first idea on how these handles could look like.

As it turns out, the vast majority of the current `Map` logic can be put into a `MapHandle`. In this first proposal I've done exactly that: Move most things from `Map` to `MapHandle`. This results in a rather large diff, but I've tried to keep commits small(er).

These handles are currently marked as `Send + Sync` (auto-derived by the compiler), which should be fine since all libbpf functions called are (effectively) threadsafe small syscall wrappers. Since their only reference is an FD, they can also be (safely) cloned (though the kernel may return I/O errors when DUP'ing FDs). There is no new unsafe code in this PR (as of now).

In order to avoid code duplication and to allow generic code, I've added a trait and moved code so that `Map` is now a relatively thin wrapper around `MapHandle`. This results in the `MapHandle`s being less lightweight than I had originally imagined. Let me know what you think about this.

This is still WIP and a few things are missing (new tests perhaps, changelog & version bump at least). The new handles do result in slight breaking changes: Most usages of `Map` now require the new trait `BpfMap` to be imported, and you can only create `MapHandle`s from userspace, not `Map`s.

This PR fixes issue #448.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -836,12 +836,38 @@ impl MapHandle {
         })
     }
 
+    /// Fetch extra map information
+    #[inline]
+    pub fn info(&self) -> Result<MapInfo> {
+        MapInfo::new(self.fd.as_fd())
+    }
+
+    /// Retrieve the `Map`'s name.
+    #[inline]
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    /// Returns a file descriptor to the underlying map.
+    #[inline]
+    pub fn fd(&self) -> BorrowedFd<'_> {
+        self.fd.as_fd()
+    }
+
+    /// Retrieve type of the map.
+    #[inline]
     pub fn map_type(&self) -> MapType {
         self.ty
     }
+
+    /// Key size in bytes
+    #[inline]
     pub fn key_size(&self) -> u32 {
         self.key_size
     }
+
+    /// Value size in bytes
+    #[inline]
     pub fn value_size(&self) -> u32 {
         self.value_size
     }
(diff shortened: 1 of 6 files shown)


danielocfb (commented):
The changes seem fine to me. We can probably move this out of Draft stage.

I left a few minor comments. Meta-comment: Can you please make sure that all commits build successfully? We shouldn't have intermittent breakages. Also see https://github.com/libbpf/libbpf-rs/blob/master/CONTRIBUTING.md#submitting-a-pull-request

Given your current implementation, it appears we could also get away without the trait and just add a `Deref` impl that, well, dereferences `Map` to `MapHandle`. Just bringing this up for potential discussion, though, I am not convinced think it's the superior design (as it kind of would leak the implementation detail, if you will). It would, however, make for slightly easier usage (no import of the `BpfMap` trait necessary, though we may need `std::ops::Deref` instead) and lower code duplication & boilerplate. Out of curiosity, have you considered doing that?

GermanCoding:
> I left a few minor comments. Meta-comment: Can you please make sure that all commits build successfully? We shouldn't have intermittent breakages. Also see https://github.com/libbpf/libbpf-rs/blob/master/CONTRIBUTING.md#submitting-a-pull-request

This was my intention, but I will double check this when I get to it (however, I will likely be unable to avoid compiler warnings, as I had to move code in stages to keep commits reasonably sized so there's always unused functions somewhere). There's also some weird `cargo fmt` issue I have to troubleshoot (linter is unhappy but my local cargo fmt is convinced the commits are formatted).

> Given your current implementation, it appears we could also get away without the trait and just add a Deref impl that, well, dereferences Map to MapHandle. Just bringing this up for potential discussion, though, I am not convinced think it's the superior design (as it kind of would leak the implementation detail, if you will). It would, however, make for slightly easier usage (no import of the BpfMap trait necessary, though we may need std::ops::Deref instead) and lower code duplication & boilerplate. Out of curiosity, have you considered doing that?

I hadn't really thought about `Deref`, but in hindsight it makes sense as `Map` has essentially become a wrapper around `MapHandle`. This wasn't was I was originally aiming for, but since I didn't want to duplicate code either that's what I ended up with.

I'm not entirely happy with the trait either, as it decouples the declaration and defintion of a fairly large interface that only has one real implementation (and will probably never have multiple). So getting rid of it may be the better option. The handles are fully public types anyway, so I don't think its an implementation detail we're leaking.

I plan on continuing this on Friday. so I'm going to address the other comments then.

danielocfb:
> (however, I will likely be unable to avoid compiler warnings, as I had to move code in stages to keep commits reasonably sized so there's always unused functions somewhere).

Warnings should not be an issue.

> There's also some weird `cargo fmt` issue I have to troubleshoot (linter is unhappy but my local cargo fmt is convinced the commits are formatted).

I think you will need nightly `rustfmt`. That's the only thing I can think of.

libbpf-rs/src/map.rs
```diff
@@ -345,66 +464,205 @@ impl Map {
...
+    /// Value size in bytes
+    fn value_size(&self) -> u32;
+
+    /// Fetch extra map information
+    fn info(&self) -> Result<MapInfo>;
+
+    /// Returns a file descriptor to the underlying map.
+    fn fd(&self) -> BorrowedFd<'_> {
+        self.as_fd()
+    }
```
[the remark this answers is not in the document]
GermanCoding: I originally tried to leave it out, but it's used frequently all over the place so this breaks a bunch of code. The fix is trivial of course.

In the end I left this in for compatibility, but if you're not worried about that I can remove it and fix the usages.

libbpf-rs/src/object.rs
```diff
@@ -18,6 +17,7 @@ use crate::OpenProgram;
 use crate::PrintLevel;
 use crate::Program;
 use crate::Result;
+use crate::{libbpf_sys, BpfMap};
```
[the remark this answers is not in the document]
GermanCoding: Ah, now I get it. The existing import style is defined in .rustfmt.toml:

`imports_granularity = "item"`

but this is only supported by unstable. Using unstable rustfmt indeed resolves this.

GermanCoding:
* All commits should now build
* Formatting issues fixed
* Feedback addressed (except for `Deref`)
* Added changelog
* Added test

I've played around with the `Deref` idea and I think this can look nicely. I was worried that there would be ambiguity with `pin` and `unpin` (because those have different implementations and should not be dereferenced), but from my testing this doesn't appear to be an issue at all.

Still, this is not in this PR yet because it requires a major rebasing. Do you prefer the `Deref` approach over the trait? I personally don't mind both, but I'm slightly leaning towards `Deref`.

GermanCoding:
I've added the `Deref` approach. However, I can revert the force push if we decide that we prefer the trait approach.

libbpf-rs/src/map.rs
```diff
@@ -241,12 +238,180 @@ impl Map {
...
+
+    /// Retrieve the underlying [`libbpf_sys::bpf_map`].
+    pub fn as_libbpf_bpf_map_ptr(&self) -> NonNull<libbpf_sys::bpf_map> {
+        self.ptr
+    }
+
+    /// Create a new [`MapHandle`] from this map.
+    pub fn to_handle(&self) -> Result<MapHandle> {
+        self.handle.try_clone()
+    }
```
danielocfb, line 322: Let's remove this method? Now that we deref into a `MapHandle` this is seems like unnecessary duplication. It's arguably a bit confusing that `<Map>.try_clone()` produces a `MapHandle`, at least from a naming perspective, but logically it does make sense.

libbpf-rs/src/map.rs
```diff
@@ -691,118 +824,30 @@ impl Map {
...
-    fn from(map: Map) -> Self {
-        match map.fd {
-            MapFd::Owned(o) => o,
-            MapFd::Borrowed(_) => unreachable!(
-                "it shouldn't be possible to have an owned map that doesn't own its fd"
-            ),
+    /// [Unpin](https://facebookmicrosites.github.io/bpf/blog/2018/08/31/object-lifetime.html#bpffs)
+    /// this map from bpffs.
+    pub fn unpin<P: AsRef<Path>>(&mut self, path: P) -> Result<()> {
+        match std::fs::remove_file(path) {
```
danielocfb, line 840: Nit: Not your change, but let's use an import at the top please?

danielocfb (commented):
Awesome! I think it looks good, but can you please address the CI issue?

libbpf-rs/src/map.rs
GermanCoding: Oh yeah right, now you can call `try_clone` directly. How about renaming `try_clone` to `new_handle`/`clone_handle`/`to_new_handle`/`to_handle` or something like that? Then it would make sense in both call sites.

GermanCoding:
Ran out of time for today, fixing the remaining issues on Monday.

libbpf-rs/src/map.rs
danielocfb: I actually kind of like `try_clone`. `to_*` methods probably should not be fallible. I suppose `try_clone_handle` would be okay -- the way I see it, we either end up with duplicate information when called on a `MapHandle` or insufficient context when called on ` Map`.

The one thing we could do is change the `MapHandle::try_clone` signature to:
```rust
pub fn try_clone(this: &MapHandle) -> Result<Self>
```

That's similar to what various pointer-style objects in the standard library do (e.g., https://doc.rust-lang.org/std/rc/struct.Rc.html#method.try_unwrap). It would mean that callers have to use `MapHandle::try_clone(&<map-handle>)`. Perhaps that would be the least confusing? I don't have a strong preference either way.

GermanCoding: I guess this is one of the reasons why the docs advise to be careful with `Deref`-implementations: They can be confusing.

I personally like Rust's `&self` syntax more, but I see why the stdlib does it like this. It's definitely the most explicit way, so probably the least confusing.

GermanCoding:
I've resolved the linting issues and changed `try_clone` to `&this`-syntax (or whatever its called).

However, I realized that it is now not obvious how to actually obtain a handle, because you have to call `try_clone` yourself, but pass a map instead of a handle (which then gets auto dereferenced to a handle). Perhaps we should add a usage example somewhere (there's one in the tests, but it's not really visible there).

danielocfb:
> Perhaps we should add a usage example somewhere (there's one in the tests, but it's not really visible there).

Yeah, that would be great. Perhaps as a doc comment?
```diff
--- libbpf-rs/src/map.rs
+++ libbpf-rs/src/map.rs
@@ -350,6 +350,16 @@ impl AsFd for Map {

 /// A handle to a map. Handles can be duplicated and dropped.
 ///
+/// While possible to [created directly][MapHandle::create], in many cases it is
+/// useful to create such a handle from an existing [`Map`]:
+/// ```no_run
+/// # use libbpf_rs::Map;
+/// # use libbpf_rs::MapHandle;
+/// # let get_map = || -> &Map { todo!() };
+/// let map: &Map = get_map();
+/// let map_handle = MapHandle::try_clone(map).unwrap();
+/// ```
+///
 /// Some methods require working with raw bytes. You may find libraries such as
 /// [`plain`](https://crates.io/crates/plain) helpful.
 #[derive(Debug)]
```

danielocfb (approved):
Looks good to me other than the aforementioned comment. Please let me know how you want to document or if you agree with the proposal.

GermanCoding:
Yeah I was looking at something similar. I actually wanted to add a similar comment already yesterday, but then I got annoyed by a huge rebase conflict that appeared while adding it and got sidetracked with other things. Do you mind if I add the patch as an extra commit on top (as opposed to rebasing it on the first one)? That would be quicker right now, otherwise I need a moment to reorder my commits.

danielocfb:
> Yeah I was looking at something similar. I actually wanted to add a similar comment already yesterday, but then I got annoyed by a huge rebase conflict that appeared while adding it and got sidetracked with other things. Do you mind if I add the patch as an extra commit on top (as opposed to rebasing it on the first one)? That would be quicker right now, otherwise I need a moment to reorder my commits.

That's fine.

danielocfb:
I merged the pull request with a few minor changes:
- change log additions were slightly reworded and adjusted in style
- removed `MapHandle::fd()` method (there exists `AsFd` impl)

Thanks again for getting this done, @GermanCoding !

PR #482, opened 2023-06-27
Remove Program::fd() method

danielocfb:
In recent times we started using file descriptor abstractions over "raw" integers. As part of that we also relied on the `AsFd` trait and provided implementations for the same.
Given that we implement this trait, it no longer makes sense to have a dedicated `fd()` method. With this change we remove it from the `Program` type.


anakryiko (approved):
LGTM

PR #486, opened 2023-06-28
Loosen RingBuffer lifetime requirements

danielocfb:
Please see individual commits for description.

### DIFF
--- a/libbpf-rs/tests/test.rs
+++ b/libbpf-rs/tests/test.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::let_unit_value)]
+
 use std::collections::HashSet;
 use std::ffi::c_void;
 use std::fs;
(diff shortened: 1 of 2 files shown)


danielocfb:
Also FYI @mendess in case you can help with review.

libbpf-rs/tests/test.rs
```diff
@@ -891,6 +891,48 @@ fn test_sudo_object_ringbuf_closure() {
...
+        let mut builder = libbpf_rs::RingBufferBuilder::new();
+        builder.add(map, callback).expect("Failed to add ringbuf");
+        let ringbuf = builder.build().expect("Failed to build");
+
+        drop(obj);
+
+        // Trigger the tracepoint. At this point `map` along with the containing
+        // `obj` have been destroyed.
+        let _pid = unsafe { libc::getpid() };
+        let () = ringbuf.consume().expect("Failed to consume ringbuf");
```
anakryiko, line 929: can you use ringbuf.poll() instead of consume, to validate that epoll actually works properly? .consume() will just go and read mmap'ed memory, but won't excercise epoll_wait()

anakryiko (approved):
API changes make sense, let's improve the test a bit to increase test coverage to epoll_wait stuff, but assume this still all works fine, ship it

libbpf-rs/tests/test.rs
danielocfb: Will test with that as well, thanks for the suggestion.

danielocfb:
Adjusted the test as per suggestion. Nightly seems busted but there is nothing we can do. Merging.

Issue #421, opened 2023-04-15
Private padding fields prevent construction of objects with struct initialization syntax


tgnottingham:
The change #297 caused a private padding field to be added to one of my generated structs, which prevented the ability to construct the object on the Rust side with struct initialization syntax (e.g. `Foo { x: 0, y: 1 }`) (just in a test, for what it's worth). I'm wondering if these private padding fields are necessary. For #297, the padding was added to ensure proper alignment, but could something like `#[repr(C, align(16)]` have been used instead? If so, are there other cases where explicit private padding fields are added, and could/should those be eliminated as well? Or perhaps be made public?

No is a fine answer. Perhaps I shouldn't be depending on being able to construct objects like this. I see that there is a `Default` implementation that could be used to first construct the object, and then modify the fields. Or I could use a struct which mirrors the generated type, and convert from the generated type to the manually created type where necessary.

danielocfb:
> For #297, the padding was added to ensure proper alignment, but could something like `#[repr(C, align(16)]` have been used instead?

From what I recall (I don't have all the details paged in anymore), we can't really infer what `align(xx)` attribute to use, because that's not directly contained in the BTF. I believe C code does a similar thing to what we do.

I don't see anything obvious wrong with making the fields public. Though arguably if private fields may come and go, perhaps the structs should have an empty private member to prevent accidental usage as in your case from the start. Right now I don't feel either way, but I haven't thought about it too hard either.

> Perhaps I shouldn't be depending on being able to construct objects like this.

Yeah, I think you probably shouldn't. The alternatives you mention seem fine.

PR #499, opened 2023-07-05
Make padding fields of generated types `pub`

danielocfb:
It arguably makes very little sense for all "regular" fields in libbpf-cargo generated types to be publicly accessible, while padding fields are not.
Mark them as public as well. That's similar to what cbindgen does, with its named padding members.

Closes: #421


danielocfb:
Going ahead with the merge here.

PR #512, opened 2023-07-11
map: move keys() helper from Map to MapHandle

atenart:
Before the introduction of MapHandle it was possible to create a Map using libbpf_rs and to iterate over its keys (with care). But now the helper is only available for Map as it was not moved alongside the other public functions. Fix this by moving the keys() helper to MapHandle, still allowing Map objects to use it thanks to the Deref impl.

For the above to work, make MapKeyIter reference a MapHandle.

[My understanding looking at the history is this move was simply forgotten, as `Map::ptr` isn't used in `fn keys` and I don't think there's an issue using `fn keys` on `MapHandle` objects specifically]

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -285,15 +285,6 @@ impl Map {
         util::parse_ret(ret)
     }
 
-    /// Returns an iterator over keys in this map
-    ///
-    /// Note that if the map is not stable (stable meaning no updates or deletes) during iteration,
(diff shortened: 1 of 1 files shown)


danielocfb (approved):
Looks great to me! Thanks! Will add a changelog entry and push.

danielocfb:
Merged as https://github.com/libbpf/libbpf-rs/commit/be363a496232f3458c2695f46fb0c99e8e984b15

PR #533, opened 2023-07-28
Implement Default for struct containing pointers

danielocfb:
Rust does not implement `Default` for pointers, no matter if the pointee implements the trait. As such, a bunch of our generated code does not compile. Specifically, bindings for struct types with a `Default` auto-derive and which contain pointers do not compile. Adjust the generation logic to emit a `Default` impl when a struct contains a pointer.


danielocfb:
The first change has the potential to unintentionally break some generated code, if there are other bugs that cause us to not have `Default` impls for pointee types. I don't have a crystal ball, so we have to see.

PR #541, opened 2023-08-07
libbpf-rs: Change ObjectBuilder opts semantics

alexgartrell:
By stashing an opts struct internally and converting name from a String to a CString, we can return a reference to an opt struct. This allows us to stack the lifetimes so that we don't need to copy-paste the name_ptr logic everywhere we call it.

The motivation for doing this was so that I could do the same for pin_root_path without adding another 3 copies of the name implementation.

### DIFF
--- a/libbpf-rs/src/object.rs
+++ b/libbpf-rs/src/object.rs
@@ -1,8 +1,8 @@
 use core::ffi::c_void;
 use std::collections::HashMap;
 use std::ffi::CStr;
+use std::ffi::CString;
 use std::mem;
-use std::os::raw::c_char;
 use std::path::Path;
 use std::ptr::NonNull;
 use std::ptr::{self};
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/object.rs
```diff
@@ -77,43 +80,23 @@ impl ObjectBuilder {
         let path_c = util::str_to_cstring(path_str)?;
         let path_ptr = path_c.as_ptr();
 
-        // Convert name to a C style pointer
-        //
-        // NB: we must hold onto a CString otherwise our pointer dangles
-        let name = util::str_to_cstring(&self.name)?;
-        let name_ptr = if !self.name.is_empty() {
```
danielocfb, line 84: I am wondering why we had this case in there. Does `libbpf` not like an empty string as name? It seems your logic changed semantics, where we now set `name` to `NULL` only if not set at all, but not when it's an empty string. Should we add a test for that (e.g., in `test_object_name`)?

danielocfb (commented):
> This allows us to stack the lifetimes so that we don't need to copy-paste the name_ptr logic everywhere we call it.

What do you mean by "stack the lifetimes"? I don't believe I fully understand the reasoning. Wouldn't just introducing a function allow you to accomplish pretty much the same goal of reducing code duplication?

I think the `Option<CString>` part of the change is fine and I like that we get rid of the pointer input from the public API. What I am somewhat struggling with is the embedding of `bpf_object_open_opts`. The main problem I see is that unless we go through `ObjectBuilder::opts()`, we end up with a half initialized object internally, as we are now free to reference the `ObjectBuilder::opts` attribute directly.
Would it be better to keep the `opts` attribute up-to-date on each setter (name + relaxed_maps currently) and make `opts()` strictly a true getter (on a shared receiver (`&self`))?

danielocfb:
> Would it be better to keep the `opts` attribute up-to-date on each setter (name + relaxed_maps currently) and make `opts()` strictly a true getter (on a shared receiver (`&self`))?

Then we could also make `relaxed_maps` work directly on `ops`, as opposed to requiring its own attribute. We only really need the `name` attribute because it keeps the allocated memory alive (that may justify adding a comment).

alexgartrell:
Yeah I think that's cleaner

alexgartrell:
By stacked lifetimes, I just meant that there wasn't a way to get the opts function to pull from its internal state without accidentally hiding the lifetime constraint. By returning a reference to opts, we can do that (unless someone copies the struct, ofc).

danielocfb (approved):
Seems fine to me. Thanks!

Issue #546, opened 2023-08-08
Can we make libbpf_rs::Link impl Send trait?


iamwwc:
I'm using `libbpf_rs::Link` on tokio context. But it's not `Send`

```
error: future cannot be sent between threads safely
    |
152 |     let h = tokio::task::spawn(task).map(|f| match f {
    |                                ^^^^ future created by async block is not `Send`
    |
    = help: within `[async block@src/:69:16: 151:6]`, the trait `std::marker::Send` is not implemented for `NonNull<bpf_link>`
note: future is not `Send` as this value is used across an await
    |
76  |         let link = xdp_pass_prog.attach_xdp(ifidx as i32)?;
    |             ---- has type `Link` which is not `Send`
...
119 |         let conn = c_c.new().await?;
    |                                           ^^^^^ await occurs here, with `link` maybe used later
...
151 |     };
    |     - `link` is later dropped here
note: required by a bound in `tokio::spawn`
   --> /root/.cargo/registry/src/rsproxy.cn-8f6827c7555bfaf8/tokio-1.29.1/src/task/spawn.rs:166:21
    |
164 |     pub fn spawn<T>(future: T) -> JoinHandle<T::Output>
    |            ----- required by a bound in this function
165 |     where
166 |         T: Future + Send + 'static,
    |                     ^^^^ required by this bound in `spawn`


```

expect it to be `Send`

```rs
unsafe impl Send for libbpf_rs::Link{}
```

danielocfb:
I think it should be fine to add this `impl`. Would you mind creating a pull request, @iamwwc?

PR #549, opened 2023-08-09
Add send constraint for Link

iamwwc:
fix https://github.com/libbpf/libbpf-rs/issues/546

### DIFF
--- a/libbpf-rs/src/link.rs
+++ b/libbpf-rs/src/link.rs
@@ -121,3 +121,5 @@ impl Drop for Link {
         let _ = unsafe { libbpf_sys::bpf_link__destroy(self.ptr.as_ptr()) };
     }
 }
+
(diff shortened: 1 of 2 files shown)


iamwwc:
I decide to close this PR as there are many struct  need `Send` such as `OpenProgram, Link, OpenMap`, I'm not sure whether this data structure are thread-safe, we would better re-investigate it

danielocfb:
> I decide to close this PR as there are many struct need `Send` such as `OpenProgram, Link, OpenMap`, I'm not sure whether those data structure are thread-safe, we would better re-investigate it

Yeah, we should take it one step at a time. `Link` should be safe to `Send` as per my understanding. There has been a bunch of discussions around similar or related issues [here](https://github.com/libbpf/libbpf-rs/issues/245) (and somewhere else I can't find off hand).

PR #564, opened 2023-08-22
Implement Send for Link

danielocfb:
It seems as if `Link` objects (`bpf_link`) are fine to send across thread boundaries. Mark the type as `Send` accordingly.

Fixes: #546


anakryiko (approved):
I'm actually unaware of anything in libbpf that might not be Send. Libbpf doesn't use thread-local storage or anything else that might have a thread-specific internal dependency. And I don't see why that would ever change, so I think it should be fine to mark all libbpf-mapped structs as Send.

PR #567, opened 2023-08-28
Support parsing btf files in raw format and add load_vmlinux_btf function 

chengshuyi:
Because most of the btf files that come with the kernel are in raw format, the btf from_path function needs to support parsing btf files in raw format. In addition, the load_vmlinux_btf function is added, which is very convenient and does not need to manually specify the btf path.

Thanks!

### DIFF
--- a/libbpf-rs/src/btf/mod.rs
+++ b/libbpf-rs/src/btf/mod.rs
@@ -149,6 +149,17 @@ impl Btf<'static> {
         inner(path.as_ref())
     }
 
+    /// Load the vmlinux btf information from few well-known locations.
+    pub fn from_vmlinux() -> Result<Self> {
+        let ptr = create_bpf_entity_checked(|| unsafe { libbpf_sys::btf__load_vmlinux_btf() })?;
+
+        Ok(Btf {
+            ptr,
+            drop_policy: DropPolicy::SelfPtrOnly,
+            _marker: PhantomData,
+        })
+    }
+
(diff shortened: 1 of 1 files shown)


libbpf-rs/src/btf/mod.rs
```diff
@@ -149,6 +149,17 @@ impl Btf<'static> {
         inner(path.as_ref())
     }
 
+    /// Load the vmlinux btf information from few well-known locations.
+    pub fn load_vmlinux_btf() -> Result<Self> {
```
danielocfb, line 153: I understand why you opted for this name, but it seems inconsistent with respect to `from_path` in the grand scheme of things. What do you think about `from_vmlinux` or `from_system_vmlinux` instead?

Also, should we add a test for this functionality?

danielocfb (approved):
Looks good to me. Left one suggestion & ask for test. Let me know what you think.

libbpf-rs/src/btf/mod.rs
chengshuyi: > I understand why you opted for this name, but it seems inconsistent with respect to `from_path` in the grand scheme of things. What do you think about `from_vmlinux` or `from_system_vmlinux` instead?
>

`from_vmlinux` is great. 

> Also, should we add a test for this functionality?

will do.

danielocfb (approved):
Looks good to me, thanks!

PR #581, opened 2023-09-06
Introduce AsRawLibbpf trait

danielocfb:
`libbpf-rs` aims to be a convenient but thin wrapper around `libbpf` functionality. Not all functionality provided of `libbpf` is exposed through its API. This can quickly become a blocker if, say, some getter is not hooked up, but the value required to perform an operation. In the spirit of being a thin wrapper, it makes sense for us to provide access to the underlying `libbpf` objects. When nothing else is available, these can be used directly in conjunction with `libbpf-sys` to, effectively, write Rust-syntax C code.
So far we have provided very ad-hoc accessor methods for retrieving pointers to the underlying `libbpf` objects. With this change we formalize this approach some more, by introducing the `AsLibbpfObject` trait and implementing it for major `libbpf-rs` types.
Using a trait is not the only possible way. We could also continue that path we were on earlier where relevant types had custom named accessors without a common trait (e.g., `Map::as_libbpf_bpf_map_ptr`). The benefits a trait brings to the table is that it unifies naming and makes it reasonably easy to get an overview of types that have `libbpf` counterparts.
The risk is that we may have to adjust the trait signature in one way or another if a newly added type has different constraints (e.g., if a `libbpf-rs` object has a potentially-`NULL` libbpf object).

### DIFF
--- a/libbpf-rs/CHANGELOG.md
+++ b/libbpf-rs/CHANGELOG.md
@@ -9,6 +9,8 @@ Unreleased
   - Adjusted `opts` to return a reference to `libbpf_sys::bpf_object_open_opts`
   - Removed object name argument from `open_memory` constructor
   - Added `pin_root_path` setter
+- Added `AsRawLibbpf` trait as a unified way to retrieve `libbpf` equivalents
+  for `libbpf-rs` objects
 - Implemented `Send` for `Link`
 - Updated `bitflags` dependency to `2.0`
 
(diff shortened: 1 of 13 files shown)


danobi:
Maybe "AsRawLibbpf"? "AsLibbpfObject" reads amibuous - those words have been used a lot in libbpf-rs as well

danielocfb:
> Maybe "AsRawLibbpf"? "AsLibbpfObject" reads amibuous - those words have been used a lot in libbpf-rs as well

Fine with me, will adjust. Thanks for the feedback.

PR #583, opened 2023-09-11
Rework Error type

danielocfb:
Our existing Error type is sub-optimal in many ways:
1) using an exhaustive enum we can't easily extend it without breaking
   backwards compatibility
2) the provided variants are rather subjective and not useful: what is a
   system error? Why is it relevant that something is a system error vs.
   something else? What could be more relevant to users, would be the
   kind of error encountered, e.g., permission denied or object not
   found.
   Similarly, what is an internal error? If an internal invariant is
   violated that's a bug, but that's not how most sites use this
   variant.
3) it is impossible to chain errors, reporting lower level errors (such
   as file not found) along with higher level context (e.g., file name).

This change proposes to rework the Error type completely. The type is very close to io::Error, just by virtue of us doing a lot of IO and interfacing with system-level APIs (most of which effectively report io::Error objects) and so it is a natural fit. We certainly mirror several of the std::io::ErrorKind variants in our own version of ErrorKind. On top of the error itself we also add a bunch of convenience conversion functions, mostly to provide additional context. These are heavily inspired by anyhow's Error type. That crate also acted as role model for the way we format our errors.
This change only introduces the error type in a new module, error2. A subsequent one will adjust all internal consumers to use it and remove the existing Error enum.


anakryiko (approved):
LGTM, the justification make sense

PR #610, opened 2023-11-20
Update nix dependency to 0.27

danielocfb:
Update the nix dependency that we depend on to 0.27, to prevent copies of multiple versions being pulled in during the build.


danielocfb:
See https://github.com/danielocfb/libbpf-rs/actions/runs/6924376450/job/18833538780#step:4:54

PR #611, opened 2023-11-20
Fix up some mutability warts on datasec skeleton accessors

danielocfb:
Please see individual commits for descriptions

### DIFF
--- a/libbpf-cargo/src/gen/mod.rs
+++ b/libbpf-cargo/src/gen/mod.rs
@@ -530,24 +530,33 @@ fn gen_skel_datasec_getters(
             None => continue,
         };
         let struct_name = format!("{obj_name}_{name}_types::{name}");
-        let mutability = if loaded && map_is_readonly(map) {
-            ""
+        let immutable = loaded && map_is_readonly(map);
+        let mutabilities = if immutable {
+            [false].as_ref()
         } else {
-            "mut"
+            [false, true].as_ref()
         };
 
-        write!(
-            skel,
-            r#"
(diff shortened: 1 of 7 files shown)


libbpf-cargo/src/gen/mod.rs
```diff
@@ -524,9 +524,9 @@ fn gen_skel_datasec_getters(
         write!(
             skel,
             r#"
-            pub fn {name}(&mut self) -> &'_ {mutability} {struct_name} {{
+            pub fn {name}(&{mutability} self) -> &{mutability} {struct_name} {{
                 unsafe {{
-                    std::mem::transmute::<*mut std::ffi::c_void, &'_ {mutability} {struct_name}>(
+                    std::mem::transmute::<*mut std::ffi::c_void, &{mutability} {struct_name}>(
```
anakryiko: is it intentional that we transmute to *mut c_void for both mutabilities?

anakryiko (approved):

libbpf-cargo/src/gen/mod.rs
danielocfb: Not sure which commit you looked at/commented on, but in the end `mutability` would be just be the empty string for the "shared" accessor. So yeah, it's intentional.

anakryiko: ah, I commented on commit #2 ("Use non-exclusive receiver for immutable datasec getters"), and it will later be replaced with {ptr_suffix}, never mind.

danielocfb:
Uhm...I can't rebase and resolve the conflict for <reasons>, so I am going to recreate the pull request with updated changes and then merge. Thanks for the review.

PR #612, opened 2023-11-20
Bump minimum supported Rust version to 1.65

danielocfb:
Bump the minimum supported Rust version to 1.65. This version is what nix 0.27 requires, and we wish to update to it.


danielocfb:
Included in https://github.com/libbpf/libbpf-rs/pull/610

PR #613, opened 2023-11-21
libbpf-cargo: Support on-the-fly vmlinux.h extraction

ueno:
This adds the `extract_vmlinux_header` method to `SkelBuilder`, which lets libbpf-cargo auto-generate the "vmlinux.h" header from the running kernel using bpftool.  This would be particularly useful when packaging a crate with "cargo package", as it doesn't need the header file to be added to the repository.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -279,6 +279,7 @@ pub fn build(
 
 // Only used in libbpf-cargo library
 #[allow(dead_code)]
+#[allow(clippy::too_many_arguments)]
 pub fn build_single(
     debug: bool,
     source: &Path,
@@ -286,6 +287,8 @@ pub fn build_single(
     clang: Option<&PathBuf>,
     skip_clang_version_checks: bool,
     options: &str,
+    bpftool: Option<&Path>,
+    vmlinux_h: Option<&Path>,
 ) -> Result<()> {
     let clang = extract_clang_or_default(clang);
     check_clang(debug, &clang, skip_clang_version_checks)?;
@@ -296,6 +299,24 @@ pub fn build_single(
     } else {
         options.to_string()
     };
+
+    if let Some(vmlinux_h) = vmlinux_h {
+        let file = fs::File::create(vmlinux_h)?;
+        let bpftool = bpftool.unwrap_or_else(|| Path::new("bpftool"));
+        Command::new(bpftool)
+            .arg("btf")
+            .arg("dump")
+            .arg("file")
+            .arg("/sys/kernel/btf/vmlinux")
+            .arg("format")
+            .arg("c")
+            .stdout(file)
+            .status()?;
+        compiler_options += &vmlinux_h
+            .parent()
+            .map(|dir| format!(" -I{}", dir.display()))
+            .ok_or_else(|| anyhow!("Failed to find include directory for vmlinux.h"))?;
+    }
 
     // Explicitly disable stack protector logic, which doesn't work with
     // BPF. See https://lkml.org/lkml/2020/2/21/1000.
--- a/libbpf-cargo/src/lib.rs
+++ b/libbpf-cargo/src/lib.rs
@@ -120,6 +120,9 @@ pub struct SkeletonBuilder {
     clang: Option<PathBuf>,
     clang_args: String,
     skip_clang_version_check: bool,
+    bpftool: Option<PathBuf>,
+    extract_vmlinux_header: bool,
+    vmlinux_h: Option<PathBuf>,
     rustfmt: PathBuf,
     dir: Option<TempDir>,
 }
@@ -139,6 +142,9 @@ impl SkeletonBuilder {
             clang: None,
             clang_args: String::new(),
             skip_clang_version_check: false,
+            bpftool: None,
+            extract_vmlinux_header: false,
+            vmlinux_h: None,
             rustfmt: "rustfmt".into(),
             dir: None,
         }
@@ -202,6 +208,22 @@ impl SkeletonBuilder {
         self
     }
 
+    /// Specify which `bpftool` binary to use
+    ///
+    /// Default searchs `$PATH` for `bpftool`
+    pub fn bpftool<P: AsRef<Path>>(&mut self, bpftool: P) -> &mut SkeletonBuilder {
+        self.bpftool = Some(bpftool.as_ref().to_path_buf());
+        self
+    }
+
+    /// Tell whether to extract vmlinux.h from the running kernel
+    ///
+    /// Default is off
+    pub fn extract_vmlinux_header(&mut self, extract: bool) -> &mut SkeletonBuilder {
+        self.extract_vmlinux_header = extract;
+        self
+    }
+
     /// Specify which `rustfmt` binary to use
     ///
     /// Default searches `$PATH` for `rustfmt`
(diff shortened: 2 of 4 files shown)


danielocfb:
Thanks for the pull request! Will take a closer look this week. One preliminary question: why do we need to introduce a feature? Can't this functionality be provided unconditionally? If the issue is the test then please make it `ignored` by default or something like that. Can you please also resolve the conflicts?

libbpf-cargo/src/build.rs
```diff
@@ -242,11 +242,11 @@ fn compile(debug: bool, objs: &[UnprocessedObj], clang: &Path, target_dir: &Path
...
-fn extract_clang_or_default(clang: Option<&PathBuf>) -> PathBuf {
-    match clang {
+fn extract_command_or_default(command: Option<&PathBuf>, default: &str) -> PathBuf {
+    match command {
         Some(c) => c.into(),
         // Searches $PATH
-        None => "clang".into(),
+        None => default.into(),
     }
 }
```
danielocfb, line 251: Please don't adjust this function like this. It's already questionable at best: the name is misleading and it basically is an `Option::unwrap_or_else` in disguise and with confusing and borderline nonsensical details. Proper way is:

```patch
--- src/build.rs
+++ src/build.rs
@@ -286,7 +286,7 @@ pub fn build_single(
     clang: Option<&PathBuf>,
     skip_clang_version_checks: bool,
     options: &str,
-    bpftool: Option<&PathBuf>,
+    bpftool: Option<&Path>,
     vmlinux_h: Option<&PathBuf>,
 ) -> Result<()> {
     let clang = extract_command_or_default(clang, "clang");
@@ -301,7 +301,7 @@ pub fn build_single(

     if let Some(vmlinux_h) = vmlinux_h {
         let file = fs::File::create(vmlinux_h)?;
-        let bpftool = extract_command_or_default(bpftool, "bpftool");
+        let bpftool = bpftool.unwrap_or_else(|| Path::new("bpftool"));
         Command::new(bpftool)
             .arg("btf")
             .arg("dump")
diff --git src/lib.rs src/lib.rs
index 73e145..dce9a5 100644
--- src/lib.rs
+++ src/lib.rs
@@ -289,7 +289,7 @@ impl SkeletonBuilder {
             self.clang.as_ref(),
             self.skip_clang_version_check,
             &self.clang_args,
-            self.bpftool.as_ref(),
+            self.bpftool.as_deref(),
             self.vmlinux_h.as_ref(),
         )
         .map_err(Error::Build)?;
```

libbpf-cargo/src/build.rs
```diff
@@ -286,8 +286,10 @@ pub fn build_single(
     clang: Option<&PathBuf>,
     skip_clang_version_checks: bool,
     options: &str,
+    bpftool: Option<&PathBuf>,
+    vmlinux_h: Option<&PathBuf>,
```
danielocfb, line 290: Please use `Option<&Path>` as suggested above.

libbpf-cargo/src/build.rs
```diff
@@ -297,6 +299,21 @@ pub fn build_single(
...
+        Command::new(bpftool)
+            .arg("btf")
+            .arg("dump")
+            .arg("file")
+            .arg("/sys/kernel/btf/vmlinux")
+            .arg("format")
+            .arg("c")
+            .stdout(file)
+            .status()?;
+        compiler_options += &format!(" -I{}", vmlinux_h.parent().unwrap().display());
```
danielocfb, line 314: Let's not panic here and instead handle the potential error gracefully, please.

libbpf-cargo/src/lib.rs
```diff
@@ -240,15 +262,25 @@ impl SkeletonBuilder {
...
-            let objfile = dir.path().join(format!("{name}.o"));
-            self.obj = Some(objfile);
             // Hold onto tempdir so that it doesn't get deleted early
             self.dir = Some(dir);
         }
 
+        if self.obj.is_none() {
+            let name = filename.split('.').next().unwrap();
+            // Unwrap is safe here since we guarantee that dir.is_some() above
+            let objfile = self.dir.as_ref().unwrap().path().join(format!("{name}.o"));
```
danielocfb, line 274: Let's use `SANITY` style comment, similar to here:
https://github.com/libbpf/libbpf-rs/blob/855eb8422e98e4632d433d5fbf065ab703674fbf/libbpf-rs/src/query.rs#L291-L292

```suggestion
            // SANITY: `dir` is guaranteed to be `Some` at this point.
            let objfile = self.dir.as_ref().unwrap().path().join(format!("{name}.o"));
```

libbpf-cargo/src/lib.rs
```diff
@@ -240,15 +262,25 @@ impl SkeletonBuilder {
...
+            self.obj = Some(objfile);
+        }
+
+        if self.extract_vmlinux_header {
+            // Unwrap is safe here since we guarantee that dir.is_some() above
```
danielocfb, line 279:
```suggestion
            // // SANITY: `dir` is guaranteed to be `Some` at this point.
            let vmlinux_h = self.dir.as_ref().unwrap().path().join("vmlinux.h");
```

danielocfb (changes requested):
Left a few comments for improvements, on top of what I already mentioned.

Regarding the larger picture: it seems to be common practice to just carry around `vmlinux.h` files from everywhere, so I guess that conceptually this functionality is fine to add. However, the question is why would we want to have a dependency on and shell out to `bpftool`: from what I understand `libbpf` provides the means for dumping BTF as C code. `bpftool`'s logic just seems to be a thin wrapper around that:

https://github.com/libbpf/bpftool/blob/784216403e89e9767833be3705bb8be0873eb52a/src/btf.c#L463-L505

In my opinion we should interface with `libbpf` here directly. Well, indirectly through `libbpf-rs` or potentially `libbpf-sys`. Let me know what you think.

ueno:
> In my opinion we should interface with libbpf here directly. Well, indirectly through libbpf-rs or potentially libbpf-sys. Let me know what you think.

Thank you for the suggestion; I will try that in this week. So far, the review comments should be addressed.

> One preliminary question: why do we need to introduce a feature? Can't this functionality be provided unconditionally?

Let me explain the idea behind this: I thought that it is a good practice to "pin" a certain version of the kernel header when distributing a package alongside its binary (.deb, .rpm, etc.), while it is generally not when packaging a source crate, including auto-generated files. So the expected workflow is: enable the `bpftool` feature when building a crate, while making it still possible to swap the vmlinux.h when building a binary package. This might not be fully supported by this PR yet, but the idea is to make `extract_vmlinux_header` a no-op for the latter case.

libbpf-cargo/src/build.rs
ueno: Thanks; applied

libbpf-cargo/src/build.rs
ueno: Done

danielocfb:
> > One preliminary question: why do we need to introduce a feature? Can't this functionality be provided unconditionally?
> 
> Let me explain the idea behind this: I thought that it is a good practice to "pin" a certain version of the kernel header when distributing a package alongside its binary (.deb, .rpm, etc.), while it is generally not when packaging a source crate, including auto-generated files. So the expected workflow is: enable the `bpftool` feature when building a crate, while making it still possible to swap the vmlinux.h when building a binary package.

Thanks for the background. This sounds like a lot of policy that I'd say does not belong in this crate. There is no concept of binary package from `libbpf-rs`'s point of view, is there? I would expect that if clients need this kind of logic, they can build it themselves inside the build script. We only provide the building blocks/mechanisms.

Building this in the client has the added benefit that `libbpf-cargo` does not need to be rebuilt with a different feature.

hzhuang1:
@ueno By default, vmlinux_515.h is only for x86. And it results in building error on aarch64 (#615). The workaround is using bpftool to generate a new vmlinux header instead.

I want to know how to generate a new vmlinux header with your patch. While I run "cargo build", I only find that "examples/capable/src/bpf/vmlinux.h" still links to "vmlinux_515.h". So I still have the build error. Do I miss any step to build libbpf-rs?

ueno:
@hzhuang1 I'm still thinking about what UI would be appropriate, but with the current code, the following change might help:
```patch
diff --git a/examples/capable/build.rs b/examples/capable/build.rs
index 3758e12..580acc1 100644
--- a/examples/capable/build.rs
+++ b/examples/capable/build.rs
@@ -10,6 +10,7 @@ fn main() {
     out.push("capable.skel.rs");
     SkeletonBuilder::new()
         .source(SRC)
+        .extract_vmlinux_header(true)
         .build_and_generate(&out)
         .unwrap();
     println!("cargo:rerun-if-changed={SRC}");
```

hzhuang1:
> @hzhuang1 I'm still thinking about what UI would be appropriate, but with the current code, the following change might help:
> 
> ```diff
> diff --git a/examples/capable/build.rs b/examples/capable/build.rs
> index 3758e12..580acc1 100644
> --- a/examples/capable/build.rs
> +++ b/examples/capable/build.rs
> @@ -10,6 +10,7 @@ fn main() {
>      out.push("capable.skel.rs");
>      SkeletonBuilder::new()
>          .source(SRC)
> +        .extract_vmlinux_header(true)
>          .build_and_generate(&out)
>          .unwrap();
>      println!("cargo:rerun-if-changed={SRC}");
> ```

[s1.log](https://github.com/libbpf/libbpf-rs/files/13703916/s1.log)

This patch doesn't help. Log is attached.

ueno:
Sorry for the inactivity; I'll try to get back to it soon.

Issue #618, opened 2023-12-03
Separate mod for Xdp


syogaraj:
Hi,

Can we have a separate mod for xdp (say `xdp.rs`) similar to what we have for `tc`? Below are the functional specs that I've worked out for this. This will also help catch up with the [feature parity with libbpf-1.0](https://github.com/libbpf/libbpf-rs/milestone/1). I do have some doubts about the API in libbpf and kindly suggest improvements if any.

Struct:
```
struct Xdp {
      fd: BorrowedFd,
      attach_opts: libbpf_sys::bpf_xdp_attach_opts,
      query_opts: libbpf_sys::bpf_xdp_query_opts,
}
```

Methods:
| Method Name | Args | libbpf-API |
|--------|--------|--------|
| attach | ifindex, flags, attach_opts |  `bpf_xdp_attach` |
| detach | ifindex, flags, attach_opts | `bpf_xdp_detach` |
| query | ifindex, flags, query_opts | `bpf_xdp_query` |
| query_id | ifindex, flags, prog_id | `bpf_xdp_query_id` | 

Need clarifications for the following:
1. Would it make sense to move the ifindex to the struct itself? 
2. Can we have a `XDPBuilder` similar to `TcHookBuilder`? I don't see any requirements for this. Just asking though.
3. `bpf_xdp_attach_opts`: attribute `old_prog_fd` - I guess this is used in combination with `libbpf_sys::XDP_FLAGS_REPLACE`. If this is correct, we should have a method for `replace` as well.
4. `bpf_xdp_query_opts`: How to expose an API for this to use the `query` method?

danielocfb:
Sounds reasonable to me.

> Need clarifications for the following:
> 
> 1. Would it make sense to move the ifindex to the struct itself?

If it's required for all operations then this would make sense, I think.

> 2. Can we have a `XDPBuilder` similar to `TcHookBuilder`? I don't see any requirements for this. Just asking though.

We should probably have a use case for it to motivate the introduction.

> 3. `bpf_xdp_attach_opts`: attribute `old_prog_fd` - I guess this is used in combination with `libbpf_sys::XDP_FLAGS_REPLACE`. If this is correct, we should have a method for `replace` as well.

I am not familiar with the attribute. Sounds like a `libbpf` question, so perhaps @anakryiko would know?

> 4. `bpf_xdp_query_opts`: How to expose an API for this to use the `query` method?

[...]

Issue #263, opened 2022-09-02
Attach mode parameter to function attach_xdp


Rei-Tw:
Hi,

Is there any way to specify the attach mode for XDP ? The current function to attach an XDP program only has the interface index as parameter (https://docs.rs/libbpf-rs/latest/libbpf_rs/struct.Program.html#method.attach_xdp).

Available attach modes are :
  - XDP_FLAGS_SKB_MODE
  - XDP_FLAGS_DRV_MODE
  - XDP_FLAGS_HW_MODE

chenhengqi:
You can use `libbpf_sys::bpf_xdp_attach`.

insearchoflosttime:
[bpf_xdp_attach()](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/tree/tools/lib/bpf/netlink.c#n277) this but we haven't yet added bpf_xdp_attach() support for libbpf-rs.  It's on[ the list of features missing for feature-parity with libbpf 1.0](https://github.com/libbpf/libbpf-rs/wiki/Feature-parity-with-libbpf-1.0).

mdaverde:
In general I think we lack support for attachment flags? 

The use case I ran into was trying to attach (directly, not through a link) a bpf program to a cgroup but needing to set `ALLOW_MULTI` to the attachment flags.

danielocfb:
@Rei-Tw @mdaverde If this is still an issue for you, please feel free to introduce an `attach_xdp_*` variant that accepts flags in a pull request. We have done something similar for uprobe attachment with de00fbe4097ebcf1f53a9f6867695093f6327e92, for example.

Noah-Kennedy:
It's an issue for me. I'll give this a go either tomorrow or next week.

danielocfb:
Fixed by https://github.com/libbpf/libbpf-rs/pull/625

Issue #206, opened 2022-03-16
Add wrappers for bpf_xdp_query and bpf_xdp_query_id


insearchoflosttime:
```
LIBBPF_API int bpf_xdp_query(int ifindex, int flags, struct bpf_xdp_query_opts *opts);
LIBBPF_API int bpf_xdp_query_id(int ifindex, int flags, __u32 *prog_id);
```

https://github.com/libbpf/libbpf/blob/3591deb9bc6b0848721f831790b78ae0a593b4c5/src/libbpf.h#L902-L903

danielocfb:
Fixed by https://github.com/libbpf/libbpf-rs/pull/625

Issue #205, opened 2022-03-16
Support bpf_xdp_attach and bpf_xdp_detach


insearchoflosttime:
Defined in libbpf as:
```
LIBBPF_API int bpf_xdp_attach(int ifindex, int prog_fd, __u32 flags,
			      const struct bpf_xdp_attach_opts *opts);
LIBBPF_API int bpf_xdp_detach(int ifindex, __u32 flags,
			      const struct bpf_xdp_attach_opts *opts);
```
https://github.com/libbpf/libbpf/blob/3591deb9bc6b0848721f831790b78ae0a593b4c5/src/libbpf.h#L898-L901

Noah-Kennedy:
Is this blocked on anything? If not, I'd be willing to contribute this.

danielocfb:
> Is this blocked on anything? If not, I'd be willing to contribute this.

I am not aware of any blockers. Please, feel free to go ahead and open a pull request @Noah-Kennedy.

danielocfb:
Fixed by https://github.com/libbpf/libbpf-rs/pull/625

PR #625, opened 2023-12-09
add: separate mod for xdp

syogaraj:
XDP-related operations are now added as a separate mod as discussed in #618 

Closes:
#618 
#205 
#206 
#263


danielocfb:
Will take a look by Wednesday the latest.

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,108 @@
...
+        };
+        util::parse_ret(ret)
+    }
+
+    /// Detach the XDP program from the interface
+    pub fn detach(&mut self, ifindex: i32) -> Result<()> {
+        let ret = unsafe {
+            libbpf_sys::bpf_xdp_detach(
+                ifindex,
+                XdpFlags::UPDATE_IF_NOEXIST.bits(),
```
danielocfb, line 76: Why are we hardcoding the flags here instead of passing them in?

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,108 @@
...
+            )
+        };
+        util::parse_ret(ret)
+    }
+
+    /// Query to inspect the program
+    pub fn query(&mut self, ifindex: i32, flags: XdpFlags) -> Result<bpf_xdp_query_opts> {
+        let mut opts = self.query_opts;
+        let err = unsafe {
+            libbpf_sys::bpf_xdp_query(ifindex, flags.bits() as i32, (&mut opts) as *mut _)
```
danielocfb, line 87: Why is this cast necessary?
```suggestion
            libbpf_sys::bpf_xdp_query(ifindex, flags.bits() as i32, &mut opts)
```

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub fn query_id(&mut self, ifindex: i32, flags: XdpFlags) -> Result<u32> {
+        let mut prog_id: u32 = 0;
+        let err = unsafe {
+            libbpf_sys::bpf_xdp_query_id(ifindex, flags.bits() as i32, (&mut prog_id) as *mut u32)
+        };
+        if err != 0 {
+            Err(Error::from_raw_os_error(errno::errno()))
+        } else {
+            Ok(prog_id)
+        }
```
danielocfb, line 106: Why are we looking at `errno` here instead of parsing the error?

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// Attach the XDP program to the given interface to start processing the packets
+    pub fn attach(&mut self, ifindex: i32, flags: XdpFlags) -> Result<()> {
+        let ret = unsafe {
+            libbpf_sys::bpf_xdp_attach(
+                ifindex,
+                self.fd.as_raw_fd(),
+                flags.bits(),
+                &self.attach_opts as *const _,
```
danielocfb, line 65: Cast seems unnecessary, no?

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,108 @@
...
+        /// See [`libbpf_sys::XDP_FLAGS_MASK`].
+        const MASK              = libbpf_sys::XDP_FLAGS_MASK as _;
+    }
+
+}
+
+/// Represents a XDP program.
+///
+/// This struct exposes operations to attach, detach and query a XDP program
+#[derive(Clone, Copy, Debug)]
```
danielocfb, line 38: I am not convinced that this type should be cloneable, let alone copyable. What is the reasoning?

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,108 @@
...
+            attach_opts: libbpf_sys::bpf_xdp_attach_opts::default(),
+            query_opts: libbpf_sys::bpf_xdp_query_opts::default(),
+        };
+        xdp.attach_opts.sz = size_of::<libbpf_sys::bpf_xdp_attach_opts>() as libbpf_sys::size_t;
+        xdp.query_opts.sz = size_of::<libbpf_sys::bpf_xdp_query_opts>() as libbpf_sys::size_t;
+        xdp
+    }
+
+    /// Attach the XDP program to the given interface to start processing the packets
+    pub fn attach(&mut self, ifindex: i32, flags: XdpFlags) -> Result<()> {
```
danielocfb, line 59: Why do all methods have `&mut self` receivers?

danielocfb (commented):
Thanks for your pull request! Your change seems mostly fine. I left a few comments. I think we should think about adding more tests. Specifically, for
- query()
- forced errors for attach(), detach(), query() and query_id()
- double attach() & detach() (second one should fail, I would expect)?

What do you think?

libbpf-rs/src/xdp.rs
syogaraj: I took references from `tc.rs` and thought that it was the preferred style of passing opts.

syogaraj:
@danielocfb, 
1. test cases for `query` and `query_id` will not return any error even if the prog is not loaded to the interface. If that's the case, the returned `opts` struct will have the `prog_id` as 0.
2. Added a test case for forced error in the `attach` API. i.e., attach twice and expect an error. 
3. Double detach doesn't seem to raise an error. 
4. Some of the requested changes were made since I took references from `tc.rs` (ex: unnecessary casts and usage of errno). Have updated with the required changes..

danielocfb (commented):
Thanks for updating! I left a few more comments regarding the API. Looking closer at it I think there are some cases that we want to support but do not currently. Let me know what you think.

danielocfb:
> 1. test cases for `query` and `query_id` will not return any error even if the prog is not loaded to the interface. If that's the case, the returned `opts` struct will have the `prog_id` as 0.

I think error handling for these functions is broken currently. Don't know if that will change much, though.

libbpf-rs/tests/test_xdp.rs
```diff
@@ -0,0 +1,40 @@
...
+        .query_id(LO_IFINDEX, XdpFlags::UPDATE_IF_NOEXIST)
+        .is_ok());
+
+    assert!(xdp_prog
+        .query(LO_IFINDEX, XdpFlags::UPDATE_IF_NOEXIST)
+        .is_ok());
+
+    assert!(xdp_prog
+        .detach(LO_IFINDEX, XdpFlags::UPDATE_IF_NOEXIST)
+        .is_ok());
```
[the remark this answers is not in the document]
syogaraj: Is there any teardown-like functionality that we can make use of for this? 
Another way I can think of is, having the two tests run in serial while the latter will always be a dummy to call the detach API. What do you think of this?

libbpf-rs/src/xdp.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// This struct exposes operations to attach, detach and query a XDP program
+#[derive(Debug)]
+pub struct Xdp<'fd> {
+    fd: BorrowedFd<'fd>,
+    attach_opts: libbpf_sys::bpf_xdp_attach_opts,
+    query_opts: libbpf_sys::bpf_xdp_query_opts,
+}
+
+impl<'fd> Xdp<'fd> {
```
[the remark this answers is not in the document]
syogaraj: Should we really support flags other than the defined ones? If its user exposed, the flags usage may not be defined. If it's really needed, I can make this change. Please let me know about this.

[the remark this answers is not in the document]
danielocfb: We should not support arbitrary flags. But we should support no flags if that's valid usage from the `libbpf` point of view. `MapFlags` [supports this use case just fine](https://github.com/libbpf/libbpf-rs/blob/9a96b1ef3a98232e15bb7b5d96ed0fd7b37e8a49/libbpf-rs/src/map.rs#L913-L914), for example.

libbpf-rs/tests/test_xdp.rs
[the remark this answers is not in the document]
danielocfb: > Another way I can think of is, having the two tests run in serial while the latter will always be a dummy to call the detach API. What do you think of this?

I don't think there are any ordering guarantees. Other than mutual exclusion all best are off.

> Is there any teardown-like functionality that we can make use of for this?

I'd create a guard object that detaches on `drop`. Alternatively we could use [`defer`](https://crates.io/crates/defer) for that.

libbpf-rs/src/xdp.rs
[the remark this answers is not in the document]
syogaraj: Got it. I've added a custom flag `ANY` with the value as `0` to support this and linked this discussion in the comment for future reference.

libbpf-rs/tests/test_xdp.rs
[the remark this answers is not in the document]
syogaraj: `scopeguard` does provide `defer` macro for this. Have added it in the `test_xdp.rs`

danielocfb (approved):
Seems okay. Thanks.

syogaraj:
@danielocfb Can you please release a build for this?

danielocfb:
Hi @syogaraj , we just released `0.22.0` and haven't accumulated much since then. Could it wait another week or so?

syogaraj:
We really needs this support in the libbpf-rs for our use case. Can we expect the release in a week? Asking to avoid manually building it from source and use them.

danielocfb:
We can publish a new release this week.

danielocfb:
FYI, I just pushed 0.22.1.

PR #647, opened 2024-01-07
add: replace support for xdp

### LINKED ISSUE #646: Support XDP replace
Currently, libbpf-rs has supported xdp attach and detach, but not for xdp replace, this is necessary for xdp program upgrade case.
For xdp replace functionality, an old program fd need to be filled in bpf_xdp_attach_opts struct for attach API。

syogaraj:
Closes #646

### DIFF
--- a/libbpf-rs/src/xdp.rs
+++ b/libbpf-rs/src/xdp.rs
@@ -91,4 +91,19 @@ impl<'fd> Xdp<'fd> {
             unsafe { libbpf_sys::bpf_xdp_query_id(ifindex, flags.bits() as i32, &mut prog_id) };
         util::parse_ret(err).map(|()| prog_id)
     }
+
+    /// Replace an existing xdp program (identified by old_prog_fd) with this xdp program
+    pub fn replace(&self, ifindex: i32, old_prog_fd: BorrowedFd<'_>) -> Result<()> {
+        let mut opts = self.attach_opts;
+        opts.old_prog_fd = old_prog_fd.as_raw_fd();
+        let ret = unsafe {
+            libbpf_sys::bpf_xdp_attach(
+                ifindex,
+                self.fd.as_raw_fd(),
+                XdpFlags::REPLACE.bits(),
+                &opts,
+            )
+        };
+        util::parse_ret(ret)
+    }
 }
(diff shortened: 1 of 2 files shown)


libbpf-rs/tests/test_xdp.rs
```diff
@@ -41,6 +46,10 @@ fn test_sudo_xdp() {
         .query(LO_IFINDEX, XdpFlags::UPDATE_IF_NOEXIST)
         .is_ok());
 
+    assert!(xdp_prog1
+        .replace(LO_IFINDEX, fd.as_raw_fd(), XdpFlags::REPLACE)
```
danielocfb, line 50: It doesn't really make sense to have a user require specifying the flag `REPLACE` on a replace operation, does it? I presume it would fail or not do the right thing without that?

Also, can you please add a proper function test and not just a mere check for a return code? This test would pass if all the function did was return `Ok(())`.

libbpf-rs/src/xdp.rs
```diff
@@ -91,4 +92,14 @@ impl<'fd> Xdp<'fd> {
             unsafe { libbpf_sys::bpf_xdp_query_id(ifindex, flags.bits() as i32, &mut prog_id) };
         util::parse_ret(err).map(|()| prog_id)
     }
+
+    /// Replace an existing xdp program (identified by old_prog_fd) with this xdp program
+    pub fn replace(&self, ifindex: i32, old_prog_fd: RawFd, flags: XdpFlags) -> Result<()> {
```
danielocfb, line 97: We shouldn't be working with a `RawFd` here. Can you please use a primitive with proper lifetime management?

danielocfb:
Can you please also make sure that CI passes? Thanks.

libbpf-rs/tests/test_xdp.rs
syogaraj: > It doesn't really make sense to have a user require specifying the flag REPLACE on a replace operation, does it? I presume it would fail or not do the right thing without that?

Got it. Should've noticed it. Removed the `flag` arg now.

> Also, can you please add a proper function test and not just a mere check for a return code? This test would pass if all the function did was return Ok(()).

Added the test to check for the old and new prog id. This way we can make sure that the replace had happened.

danielocfb (approved):
Looks good to me, thanks.

syogaraj:
@danielocfb please close all xdp related. Just FYI.

Issue #615, opened 2023-11-23
build issue with incomplete definition


hzhuang1:
I tried to build libbpf-rs. But I got this error. I could reproduce it on both arm64 and x86. The test environment was ubuntu 20.04 and ubuntu 22.04. I tried both official rust version and ubuntu rust version.

How to fix this build issue?

:~/libbpf-rs$ cargo build
   Compiling phf v0.11.2
   Compiling capable v0.1.0 (/home/hzhuang1/libbpf-rs/examples/capable)
error: failed to run custom build command for `capable v0.1.0 (/home/hzhuang1/libbpf-rs/examples/capable)`

Caused by:
  process didn't exit successfully: `/home/hzhuang1/libbpf-rs/target/debug/build/capable-2a443e9fcbfb8a89/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at examples/capable/build.rs:14:10:
  called `Result::unwrap()` on an `Err` value: Build(Failed to compile /tmp/.tmpvx8MpJ/capable.o from src/bpf/capable.bpf.c

  Caused by:
      0: Command `clang -I/tmp/.tmpshftmq/bpf/src -fno-stack-protector -D__TARGET_ARCH_arm64 -g -O2 -target bpf -c src/bpf/capable.bpf.c -o /tmp/.tmpvx8MpJ/capable.o` failed (exit status: 1)
      1: src/bpf/capable.bpf.c:102:5: error: incomplete definition of type 'struct user_pt_regs'
         int BPF_KPROBE(kprobe__cap_capable, const struct cred *cred,
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         /tmp/.tmpshftmq/bpf/src/bpf/bpf_tracing.h:818:20: note: expanded from macro 'BPF_KPROBE'
                 return ____##name(___bpf_kprobe_args(args));                        \
                                   ^~~~~~~~~~~~~~~~~~~~~~~~
         /tmp/.tmpshftmq/bpf/src/bpf/bpf_tracing.h:798:41: note: expanded from macro '___bpf_kprobe_args'
         #define ___bpf_kprobe_args(args...)     ___bpf_apply(___bpf_kprobe_args, ___bpf_narg(args))(args)
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #650, opened 2024-01-12
examples: Use vmlinux.h for current architecture

danielocfb:
In order to compile the BPF byte code the examples need vmlinux.h to be available. This header is vendored alongside the examples, but it is specific to x86. That causes issues once said examples are compiled for any other architecture.
This change fixes this problem by vendoring vmlinux.h for more architectures (copied from libbpf-bootstrap, where we maintain such a set of header files already). It also adjusts all examples to pick the appropriate header file for the architecture currently being targeted.

Fixes: #615


anakryiko:
This is much cleaner than what we had to do in libbpf-bootstrap. We probably can create a few symlinks to accommodate Rust's version of ARCH names to make all that cleaner.

PR #660, opened 2024-01-29
Overhaul feature set

danielocfb:
This change overhauls the crate's feature set. The 'novendor' feature violates Cargo's "features should be additive" rule [0]. It has been made a no-op in libbpf-sys starting with 1.3.0. With this change we remove it from this crate altogether. We introduce the 'vendored' feature, which causes all library dependencies to be built from vendored source code. We preserve the current default, which is using a vendored copy of libbpf but relying on system libelf and zlib, because while it would be great to be self-contained, the additional build tools required to build said libraries makes this a no-go.

[0] https://doc.rust-lang.org/cargo/reference/features.html

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,7 @@ jobs:
         with:
           # Please adjust README and rust-version field in Cargo.toml files when
           # bumping version.
-          toolchain: 1.65.0
+          toolchain: 1.66.0
           components: rustfmt
       - uses: Swatinem/rust-cache@v2
       - name: Build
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -61,6 +61,39 @@ jobs:
...
+    name: Build [${{ matrix.args }}]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - args: "features = ['vendored']"
+          - args: "features = ['static']"
+          # TODO: Should build without features, but that requires system
+          #       libbpf and ubuntu 22.04 only has 0.5 (..?)
```
anakryiko, line 74: yeah, I know :( next LTS is 24.04, right? That will have at least v1.3, though, so just a little more to wait for it...

anakryiko (approved):
LGTM

Issue #589, opened 2023-09-28
Potential unsoundness in libbpf-cargo generated code


danielocfb:
I think there is possibility of unsoundness in the generated code, depending on what C code does. E.g., in the `capable` example we have the `tool_config` struct, which contains the `uniqueness` enumeration:

https://github.com/libbpf/libbpf-rs/blob/cb4186daf9436d7c6e4c481a81aeec7bf2af1d42/examples/capable/src/bpf/capable.bpf.c#L13-L18

https://github.com/libbpf/libbpf-rs/blob/cb4186daf9436d7c6e4c481a81aeec7bf2af1d42/examples/capable/src/bpf/capable.h#L21-L23

Once we generate the skeleton we end up with:

```rust
pub mod capable_rodata_types {
    #[derive(Debug, Copy, Clone)]
    #[repr(C)]
    pub struct rodata {
        pub tool_config: __anon_1,
    }
    #[derive(Debug, Default, Copy, Clone)]
    #[repr(C)]
    pub struct __anon_1 {
        pub tgid: u32,
        pub verbose: bool,
        pub unique_type: uniqueness,
    }
    #[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
    #[repr(u32)]
    pub enum uniqueness {
        #[default]
        UNQ_OFF = 0,
        UNQ_PID = 1,
        UNQ_CGROUP = 2,
    }
}
```
and
```rust
pub fn rodata(&mut self) -> &'_ mut capable_rodata_types::rodata {
    unsafe {
        std::mem::transmute::<*mut std::ffi::c_void, &'_ mut capable_rodata_types::rodata>(
            self.skel_config.map_mmap_ptr(2).unwrap(),
        )
    }
}
```

That cast, however, is anything but safe. Rust's `uniqueness` requires that it has *only* any of the variant values 0, 1, or 2. But in C code it is fine to assign, say, `0xff` to the `enum` object (without triggering UB, I believe). The result would be UB in Rust.

A similar thing can be observed for `bool`, though it may be that both languages have the same `bool` representation (questionable, because C leaves binary representation mostly undefined, I believe, but if it's all LLVM we probably luck out).

My reading is that this is basically a variant of https://github.com/rust-lang/rust/issues/36927 (though I haven't gone through the issue in its entirety).

[...]

PR #672, opened 2024-02-09
libbpf-cargo: Wrap generated "unsafe" type in MaybeUninit

d-e-s-o:
Certain Rust types that we use in our generated skeleton are not valid for all bit patterns. Bools, for example, are defined to be one byte in size, but their representation allows only for 0 and 1 values to be used in said byte -- everything else is undefined behavior. Enums have similar problems, in that not the entire space of the backing type is necessarily used by variants.
The result is an unsoundness issue in our generated skeletons, because C code could accidentally set an enum to an invalid value and Rust code would exhibit undefined behavior.
This change fixes this problem by wrapping the corresponding attributes in MaybeUninit. In so doing users have to be explicit when accessing them and make sure that the current representation is indeed valid.

Closes: #589

### DIFF
--- a/examples/capable/src/main.rs
+++ b/examples/capable/src/main.rs
@@ -187,8 +187,16 @@ fn main() -> Result<()> {
     let mut open_skel = skel_builder.open()?;
     //Pass configuration to BPF
     open_skel.rodata_mut().tool_config.tgid = opts.pid; //tgid in kernel is pid in userland
-    open_skel.rodata_mut().tool_config.verbose = opts.verbose;
-    open_skel.rodata_mut().tool_config.unique_type = opts.unique_type;
+    open_skel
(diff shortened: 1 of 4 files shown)


anakryiko (commented):
still draft state, so you plan to work on it some more?

can we get a small example/test with enum/bool global variables so it would be easy to see how it is meant to be used in Rust code?

d-e-s-o:
> still draft state, so you plan to work on it some more?

Yeah, will need to fix the tests.

> can we get a small example/test with enum/bool global variables so it would be easy to see how it is meant to be used in Rust code?

Will do.

d-e-s-o:
This pull request should be ready for review now, @anakryiko.

anakryiko (approved):

PR #676, opened 2024-02-13
Stop generation of duplicate types

danielocfb:
When a type is present in more than one section in the ELF binary (e.g., in .bss as well as .rodata), the generation code ends up generating multiple Rust correspondents as well. That is horrible for interoperability, because there is no language-level connection between these types and so the corresponding objects are not interchangeable in any way.

This change fixes this problem by putting all generated types in a single Rust module, '<project>_types', as opposed to having them be emitted into modules for each of the "data sections".


anakryiko (approved):
very straightforward, nice!

PR #680, opened 2024-02-19
RingBuffer: introduce raw_poll() and raw_consume()

arighi:
When consuming items from a ring buffer we may want to get the exact amount of items consumed, instead of returning a simple success/failure result.

For this reason introduce new methods: raw_poll() and raw_consume() that will return the same return code (as i32) from libbpf.

This can be used either to check the amount of items consumed from the ring buffer but also to check for specific error numbers.

Thank you for considering a contribution!

In order to streamline review experience for contributors and reviewers, please
be sure to *read* and *follow* the [Contributor's Guide][cont-guide]. It
lays out basic best practices, which, if followed will reduce unnecessary back
and forth and, ultimately, minimize the time it takes to get your change into
the library.

[cont-guide]: https://github.com/libbpf/libbpf-rs/blob/master/CONTRIBUTING.md

### DIFF
--- a/libbpf-rs/src/ringbuf.rs
+++ b/libbpf-rs/src/ringbuf.rs
@@ -156,13 +156,23 @@ impl RingBuffer<'_> {
     /// each one. Polls continually until we either run out of events to consume
     /// or `timeout` is reached. If `timeout` is Duration::MAX, this will block
     /// indefinitely until an event occurs.
-    pub fn poll(&self, timeout: Duration) -> Result<()> {
+    ///
+    /// Return the amount of events consumed, or a negative value in case of error.
+    pub fn poll_raw(&self, timeout: Duration) -> i32 {
         let mut timeout_ms = -1;
         if timeout != Duration::MAX {
             timeout_ms = timeout.as_millis() as i32;
         }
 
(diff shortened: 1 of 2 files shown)


danielocfb (commented):
Makes sense to me, thanks for the addition! Two requests: could we change the names from `raw_XXX` to `XXX_raw`, please? And then would you mind adding a test that relies on the error reporting behavior that you mentioned to quickly exit? This way we'd be sure to not break or remove this functionality. You should be able to copy and amend one of the existing ring buffer tests easily.

danielocfb:
Also, feel free to rebase on top of `master` to get rid of the CI failure.

arighi:
Added a test for the new `consume_raw()` and `poll_raw()` primitives and also another test to check the errcode returned by a callback.

danielocfb (approved):
Looks great, thanks!

danielocfb:
Note that it will likely take a while for a new release to be cut, as we are in the process of addressing a bunch of short comings that are breaking changes for the most part. If you urgently need a new release I may be able to cut a one-off v0.22.2 or so, though we generally only work at top of tree and don't backport.

arighi:
@danielocfb it's ok, I don't need this change urgently, but it's good to know that it's merged. Once the new libbpf-rs will be out I'll update my code, thanks!

PR #687, opened 2024-02-24
Add recursion_misses field to ProgramInfo struct

jfernandez:
The `recursion_misses` statistic is incremented under two scenarios when BPF stats are enabled:

1. When recursion is detected within a BPF program execution. In such a case, the recursive execution is prevented, thus incrementing the recursion_misses counter.
2. When a kprobe handler tries to execute a BPF program and the kernel finds that another program is already running on the CPU. 

These increments help track instances where BPF program execution is skipped due to recursion or concurrent execution prevention mechanisms.

This PR exposes this counter in the ProgramInfo struct. I'd like to add it to [bpftop](https://github.com/Netflix/bpftop)

### DIFF
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -153,6 +153,7 @@ pub struct ProgramInfo {
     pub prog_tags: Vec<Tag>,
     pub run_time_ns: u64,
     pub run_cnt: u64,
+    pub recursion_misses: u64,
 }
 
 /// An iterator for the information of loaded bpf programs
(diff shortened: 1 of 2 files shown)


libbpf-rs/src/query.rs
```diff
@@ -153,6 +153,7 @@ pub struct ProgramInfo {
     pub prog_tags: Vec<Tag>,
     pub run_time_ns: u64,
     pub run_cnt: u64,
+    pub recursion_misses: u64,
```
danielocfb, line 156: Can you please add a doc comment for the new member?

danielocfb (commented):
The change seems fine, thanks, I just have one request inline.

danielocfb (approved):
Looks good to me, thanks!

PR #692, opened 2024-03-02
Add support for attach_ksyscall()

syogaraj:


danielocfb:
Hi @syogaraj, I suppose you are still working on this PR, as it's in draft state. Do let me know when you want me to take a look.

syogaraj:
sure @danielocfb. 
The test case for this PR succeeds in my machine where as failing in the runner action. I'm still checking on this. Do you have any idea about this?
Note: I couldn't find any test cases for `attach_kprobe` in `test.rs`.

libbpf-rs/tests/test.rs
```diff
@@ -1793,3 +1793,28 @@ fn test_sudo_map_autocreate_disable() {
...
+        .expect("Failed to find program");
+
+    let _link = prog
+        .attach_ksyscall(false, "kill")
+        .expect("Failed to attach prog");
+
+    let map = obj.map("ringbuf").expect("Failed to get ringbuf map");
+    let action = || {
+        // This is fine as we won't be having a negative pid
+        let _ = std::process::Command::new("kill -9 -1000").spawn();
```
chantra, line 1814: Mind that negative PIDs have special meaning with kill.
man 2 kill
> If pid is less than -1, then sig is sent to every process in the process group whose ID is -pid

`kill`, the command line, seems to only care about -1 though.

danielocfb:
Perhaps the following patch helps?
```patch
--- tests/test.rs
+++ tests/test.rs
@@ -3,6 +3,7 @@
 use std::collections::HashSet;
 use std::ffi::c_void;
 use std::fs;
+use std::io;
 use std::hint;
 use std::io::Read;
 use std::mem::size_of;
@@ -1810,9 +1811,11 @@ fn test_sudo_attach_ksyscall() {

     let map = obj.map("ringbuf").expect("Failed to get ringbuf map");
     let action = || {
-        // This is fine as we won't be having a negative pid
-        let _ = std::process::Command::new("kill -9 -1000").spawn();
-        std::thread::sleep(Duration::from_secs(1)); // sleep neeeded, else the test fails.
+        // Send `SIGCHLD`, which is ignored by default, to our process.
+        let ret = unsafe { libc::kill(libc::getpid(), libc::SIGCHLD) };
+        if ret < 0 {
+            panic!("kill failed: {}", io::Error::last_os_error());
+        }
     };
     let result = with_ringbuffer(map, action);

```

syogaraj:
Thanks @danielocfb 
The suggested changes worked and all checks have passed. I've updated this PR for review.

PR #693, opened 2024-03-05
fix: pass negative error code to Error::from_raw_os_error

honsunrise:
Since `libbpf_get_error` always returns negative error code, it should be **inverted** before passing in `Error::from_raw_os_error`.


danielocfb:
What's the call you made that failed with a negative error code in this way @honsunrise ?

danielocfb (approved):
Thanks!

Issue #606, opened 2023-11-07
Mutability conflicts between various skeleton-based accessors


danielocfb:
The skeleton-based accessors (`bss`, `rodata`, ...) tend to conflict with other reasonable-to-call methods, such as `maps()`.

For example, from `capable`:
```rust
    impl OpenCapableSkel<'_> {
        pub fn progs(&self) -> OpenCapableProgs<'_> {
            OpenCapableProgs { inner: &self.obj }
        }

        pub fn progs_mut(&mut self) -> OpenCapableProgsMut<'_> {
            OpenCapableProgsMut {
                inner: &mut self.obj,
            }
        }

        pub fn maps(&self) -> OpenCapableMaps<'_> {
            OpenCapableMaps { inner: &self.obj }
        }

        pub fn maps_mut(&mut self) -> OpenCapableMapsMut<'_> {
            OpenCapableMapsMut {
                inner: &mut self.obj,
            }
        }

        pub fn rodata(&mut self) -> &'_ mut capable_rodata_types::rodata {
            unsafe {
                std::mem::transmute::<*mut std::ffi::c_void, &'_ mut capable_rodata_types::rodata>(
                    self.skel_config.map_mmap_ptr(2).unwrap(),
                )
            }
        }

        pub fn bss(&mut self) -> &'_ mut capable_bss_types::bss {
            unsafe {
                std::mem::transmute::<*mut std::ffi::c_void, &'_ mut capable_bss_types::bss>(
                    self.skel_config.map_mmap_ptr(3).unwrap(),
                )
            }
        }

        pub fn kconfig(&mut self) -> &'_ mut capable_kconfig_types::kconfig {
            unsafe {
                std::mem::transmute::<*mut std::ffi::c_void, &'_ mut capable_kconfig_types::kconfig>(
                    self.skel_config.map_mmap_ptr(4).unwrap(),
                )
            }
        }
    }
```

If you have a `bss` reference you can't have `rodata` as well. As a first step, we may want to add `*_mut` variants, similar to what we do for `maps`. But really this is all rather...painful to use by clients. We should see if we can improve usability in general. Conceptually there is no connection between those calls, and yet they can conflict.

danielocfb:
[...]

PR #700, opened 2024-03-06
libbpf-cargo: Add "raw" datasec accessors

danielocfb:
It's possible to run into borrow checker conflicts with the generated datasec accessor functions. The reason for that is that they all borrow from the very same skeleton object and so once one of the borrows is mutable, no other borrows may exist.
Yet, in reality each datasec object is backed by strictly distinct regions of memory and so there is no actual conflict. However, we haven't found a way to convey this property to the Rust borrow checker given current language features, other than falling back to runtime checking (there are proposals floating around for annotating method signatures with the members they borrow, which would allow the borrow checker to reason correctly, but they are not stable by any means).

With this change we take a way out in the form of exposing the "raw" pointers, effectively opting out of borrow checking altogether and pushing the burden of "proof" to the user. E.g., for the capable example:

```rust
let rodata = open_skel.rodata_raw();
let rodata_mut = open_skel.rodata_mut();
//Pass configuration to BPF
rodata_mut.tool_config.tgid = opts.pid; //tgid in kernel is pid in userland
rodata_mut.tool_config.verbose.write(opts.verbose);
rodata_mut.tool_config.unique_type.write(opts.unique_type);

println!("{:?}", unsafe { (*rodata).tool_config });
```

Closes: #606

### DIFF
--- a/libbpf-cargo/CHANGELOG.md
+++ b/libbpf-cargo/CHANGELOG.md
@@ -9,6 +9,7 @@ Unreleased
   having multiple modules for various sections (`.bss`, `.rodata`, etc.)
 - Fixed potential unsoundness issues in generated skeletons by wrapping "unsafe"
   type in `MaybeUninit`
+- Added pointer based ("raw") access to datasec type to generated skeletons
 - Updated `libbpf-sys` dependency to `1.3.0`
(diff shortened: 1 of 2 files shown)


anakryiko (approved):
it's a good escape hatch to have, LGTM

PR #706, opened 2024-03-11
Use `CARGO_CFG_TARGET_ARCH` environment variable in example build scripts

danielocfb:
It appears as if `std::env::consts::ARCH` does not get updated to the target architecture in a cross-compilation context. Switch to using the `CARGO_CFG_TARGET_ARCH` environment variable in the hopes that we get the desired behavior.


anakryiko (approved):
LGTM

PR #707, opened 2024-03-11
libbpf-cargo: Use "contains" for checking existence of __TARGET_ARCH_…

danielocfb:
…<xxx>

When we added logic for passing through clang arguments in a more structured fashion, we switched to using an equality check for the presence of the __TARGET_ARCH_<xxx> define. However, that does not work, because the define contains the architecture as a suffix. Switch to just checking for a substring.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -179,7 +179,7 @@ fn compile_one(
 
     if !clang_args
         .iter()
-        .any(|arg| arg.as_os_str() == OsStr::new("-D__TARGET_ARCH_"))
+        .any(|arg| arg.to_string_lossy().contains("__TARGET_ARCH_"))
     {
         let arch = match ARCH {
             "x86_64" => "x86",


libbpf-cargo/src/build.rs
```diff
@@ -179,7 +179,7 @@ fn compile_one(
 
     if !clang_args
         .iter()
-        .any(|arg| arg.as_os_str() == OsStr::new("-D__TARGET_ARCH_"))
+        .any(|arg| arg.to_string_lossy().contains("__TARGET_ARCH_"))
```
anakryiko, line 182: you actually need to check if it's a prefix of a string, not really "contains"

danielocfb: Both `-DXXX` and `-D XXX` are valid argument specifications, from what I understand. So I don't think we can assume that it's a strict prefix.

anakryiko: ah, the original sting contains `-D`. It's probably fine in practice to just go with contains, unlikely someone will do `-Dhaha__TARGET_ARCH_blah`, but technically speaking you are matching more stuff than you actually intend to.

PR #708, opened 2024-03-11
libbpf-cargo: Set __TARGET_ARCH_<xxx> define based on CARGO_CFG_TARGE…

danielocfb:
…T_ARCH

Similar to what we did in commit 1b08a4441404 ("Use `CARGO_CFG_TARGET_ARCH` environment variable in example build scripts"), switch to using the `CARGO_CFG_TARGET_ARCH` environment variable for inferring what value to set the __TARGET_ARCH_<xxx> to. This should make the logic work properly even in cross compilation contexts.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -181,12 +181,15 @@ fn compile_one(
         .iter()
         .any(|arg| arg.to_string_lossy().contains("__TARGET_ARCH_"))
     {
-        let arch = match ARCH {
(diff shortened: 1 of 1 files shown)


danielocfb:
Ugh, that may not work that easily, because the code may be used from a build script OR the library, from what I understand.

danielocfb:
Okay, I think we have it now. `CARGO_CFG_TARGET_ARCH` is only available in a build script context, but the code in question may be invoked either from a build script or as "regular" library code.

PR #712, opened 2024-03-13
Remove lazy_static dependency

yan-ace62:
std::sync::OnceLock has been stabilized in rust version 1.70.  [std::sync::LazyLock](https://github.com/rust-lang/rust/issues/109736) is not stable yet. Implement LazyLock with std::sync::OnceLock as an alternative to std::sync::LazyLock.


danielocfb (approved):
Very nice, thanks!

PR #715, opened 2024-03-13
Introduce tcp_ca example

danielocfb:
Introduce the tcp_ca example, showing how to use BPF struct_ops to inject a dummy TCP congestion algorithm into the kernel and use it during a loopback based data exchange.


iamkafai:
tcp_ca.bpf.c looks fine. It is only sending a few Mbytes in/out of lo. Speed does not matter, so no need to be a real CC to adjust cwnd or things in .cong_control/.ssthresh/.undo_cwnd.

danielocfb:
> tcp_ca.bpf.c looks fine. It is only sending a few Mbytes in/out of lo. Speed does not matter, so no need to be a real CC to adjust cwnd or things in .cong_control/.ssthresh/.undo_cwnd.

Thanks for taking a look. Let's get this thing in then.

PR #726, opened 2024-03-19
Remove nix dependency in libbpf-rs

### LINKED ISSUE #714: Eliminate `nix` usage
We should probably trim down our dependencies a bit. `nix` provides very little value while requiring repeated upkeep. It seems like a sensible candidate to drop. Could venture into `libbpf-sys` as well and eliminate it from there.

yan-ace62:
close #714

### DIFF
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -14,11 +14,14 @@ use std::ffi::c_void;
 use std::ffi::CString;
 use std::io;
 use std::mem::size_of_val;
+use std::os::fd::AsFd;
+use std::os::fd::AsRawFd;
+use std::os::fd::BorrowedFd;
+use std::os::fd::FromRawFd;
+use std::os::fd::OwnedFd;
 use std::os::raw::c_char;
 use std::ptr;
 use std::time::Duration;
-
-use nix::unistd::close;
 
 use crate::util;
 use crate::MapType;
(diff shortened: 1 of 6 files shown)


libbpf-rs/src/query.rs
```diff
@@ -419,14 +421,26 @@ impl Iterator for ProgInfoIter {
...
-                Err(_err) => (),
-            }
+impl Iterator for ProgInfoIter {
+    type Item = ProgramInfo;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let fd = match self.next_valid_fd() {
+            Some(fd) => fd,
+            None => return None,
+        };
```
danielocfb, line 436: Should be the same as:
```suggestion
        let fd = self.next_valid_fd()?;
```
Right?

libbpf-rs/src/query.rs
```diff
@@ -562,14 +577,26 @@ impl Iterator for BtfInfoIter {
...
+impl Iterator for BtfInfoIter {
+    type Item = BtfInfo;
 
-            match info {
-                Ok(i) => return Some(i),
-                // TODO: We should consider bubbling up errors properly.
-                Err(_err) => (),
-            }
+    fn next(&mut self) -> Option<Self::Item> {
+        let fd = match self.next_valid_fd() {
```
danielocfb, line 589: Same question as above.

danielocfb (approved):
Looks great, thanks again for working on this issue. Overall this seems like a solid improvement. I left two comments regarding minor nitpicks. Let me know if I missed something. Could you please format the code so that CI is green? Then we can go ahead and merge.

PR #739, opened 2024-03-21
Introduce and use vmlinux crate

danielocfb:


vmlinux/src/lib.rs
```diff
@@ -0,0 +1,161 @@
+#![allow(non_upper_case_globals)]
+
+use std::env::consts::ARCH;
+use std::path::Path;
+use std::path::PathBuf;
+
+// TODO: Ideally we'd deduplicate contents at this level (as well; not just on
```
anakryiko, line 7: those are also huge text files which would compress very nicely, but I assume it's impossible in const context?

anakryiko (approved):
LGTM. Do you plan to convert libbpf-bootstrap rust examples to this as well?

vmlinux/src/lib.rs
d-e-s-o: Likely not in a `const` context, yeah. We could probably have a function based interface if need be and load and cache stuff on demand. I mostly got nerd-sniped here and wanted to make it work as I had originally envisioned...

The other avenue would to exclude unnecessary headers based on features. Either way, it's probably not urgent unless/until we use this functionality in anything but example & test code.

d-e-s-o:
> LGTM. Do you plan to convert libbpf-bootstrap rust examples to this as well?

Not immediately, but if we decide to publish this crate eventually then yes, happy to adjust the code there.

PR #741, opened 2024-03-26
Avoid using generated va_list type

EHfive:
When using libbpf-sys' bindgen feature, the generated va_list type varies on different platforms.

Use void pointer as va_list pointer type and transmute pointer of type use va_list pointer instead for cross platform support.

See https://github.com/rust-lang/rust-bindgen/issues/2631


danielocfb (approved):
This seems fine, but can you please add a comment as to why that is being done? Otherwise it will get removed eventually, I am sure.

danielocfb:
Thanks again!

PR #742, opened 2024-03-26
Init padding fields in libbpf_sys types

EHfive:
libbpf-sys structs might have additional padding fields on some 32-bit platform(like arm32, mips32) when using bindgen, initialize all fileds of libbpf structs with `...Default.default()`.


danielocfb (approved):
Thank you!

PR #749, opened 2024-03-29
libbpf-cargo: Add support for custom .data sections

danielocfb:
This change adds support for handling of custom .data sections, declared via SEC(".data.xxxx") in the BPF C code. It makes sure that we generate the corresponding Rust types as well as the proper maps getters.

### DIFF
--- a/libbpf-cargo/src/gen/mod.rs
+++ b/libbpf-cargo/src/gen/mod.rs
@@ -6,6 +6,9 @@ use std::collections::HashSet;
 use std::ffi::c_void;
 use std::ffi::CStr;
 use std::ffi::CString;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Result as FmtResult;
 use std::fmt::Write as fmt_write;
 use std::fs::File;
 use std::io::stdout;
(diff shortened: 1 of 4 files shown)


danielocfb:
Unsure if we need the same dance for `.bss.XXX` and `.rodata.XXX`? Are these possible combinations?

libbpf-cargo/src/gen/mod.rs
```diff
@@ -208,6 +212,9 @@ pub(crate) fn canonicalize_internal_map_name(s: &str) -> Option<InternalMapType>
         // The `*.link` extension really only sets an additional flag in lower
         // layers. For our intents and purposes both can be treated similarly.
         Some(InternalMapType::StructOps)
+    } else if let Some(idx) = s.find(".data.") {
```
anakryiko, line 215: can you please also add ".rodata.*" and ".bss.*" for completeness?

also see https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L1789-L1794, I don't think you need this complicated logic, just check for ".data." prefix.

anakryiko:
> Unsure if we need the same dance for `.bss.XXX` and `.rodata.XXX`? Are these possible combinations?

sorry, didn't see this question when I wrote previous comment. Yes, all of these are possible.

libbpf-cargo/src/gen/mod.rs
danielocfb: Thanks for the pointer. All done now.

anakryiko (approved):
Looks good, thanks!

PR #750, opened 2024-04-01
libbpf-cargo: escape map names

eliad-wiz:
commit 51ffd0f added support for custom sections.

this broke skeleton generation when bpf_printk() was used along with BPF_NO_GLOBAL_DATA, as e.g. the map '.rodata.str1.1' contains invalid chars. add simple escaping to the map name as well (similar to the one used for the section name)


danielocfb (approved):
Yeah, good catch, thanks! The `replace` in the one path was really only intended for the first separator, which is always present. I figured we would not allow additional special characters in the rest. Alas...

PR #757, opened 2024-04-02
libbpf-cargo: Add support for "import injection" to generated skeletons

danielocfb:
Our generated skeletons hardcode the libbpf-rs dependency to use. That is fine in principle, but it can be limiting for users. For example, a user may legitimately want to import libbpf-rs under a different name, in which case skeletons simply won't work. Or, instead of having a direct dependency on libbpf-rs they may want to rely on a re-export through a different dependency.
Currently there is no way to cater to these cases. With this change we introduce an escape hatch that allows for such usage. Specifically, we generate a super::* import that optionally pulls in items defined in the outer module. It should be extremely unlikely that this pulls in any conflicting items, because this should only honor 'pub'. In any event, conflicts can always be mitigated by introducing an intermediary module that has a clean slate environment. What this hatch provides users with, is the ability to inject a "prelude" of sorts into the generated skeleton. E.g., in the case of a re-export as mentioned before, that could be a 'use <renamed-libbpf-rs> as libbpf_rs;' statement.

### DIFF
--- a/examples/tcp_ca/Cargo.toml
+++ b/examples/tcp_ca/Cargo.toml
@@ -11,5 +11,7 @@ vmlinux = { path = "../../vmlinux" }
 
 [dependencies]
 clap = { version = "4.0.32", features = ["derive"] }
-libbpf-rs = { path = "../../libbpf-rs" }
+# We rename the `libbpf-rs` dependency here to illustrate how skeleton creation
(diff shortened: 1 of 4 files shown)


anakryiko (approved):
dark Rust magic, but if it helps - it helps :)

Issue #777, opened 2024-05-07
NULL key when using bpf_update_elem with BLOOM_FILTER


barthr:
Hi there,

Thanks for the great library. It has been a pleasure to use in writing my ebpf programs. Recently I did encounter an issue when using a BLOOM_FILTER. The docs state the following:

> A userspace program can add a value to a bloom filter using libbpf’s bpf_map_update_elem function. The key parameter must be set to NULL and flags must be set to BPF_ANY. Returns 0 on success, or negative error in case of failure.

However, it is not possible to pass NULL to this function with the current signature:
https://github.com/libbpf/libbpf-rs/blob/c10fffad7807ac0a03a8cf4fe9b76a4993b358af/libbpf-rs/src/map.rs#L777

Key has to be a slice to an u8 array. When passing in &[] it won't be able to add the value to the BLOOM_FILTER map. 

I did find the following code for generating the key:
https://github.com/libbpf/libbpf-rs/blob/c10fffad7807ac0a03a8cf4fe9b76a4993b358af/libbpf-rs/src/map.rs#L561-L567

I think this needs to be extended to also include BLOOM_FILTER. I can create a PR for that if you like.

danielocfb:
Closed via 946feb5d841f734c86426a576bf0bc12417cdb85

PR #778, opened 2024-05-07
 Add crud support for BPF_MAP_TYPE_BLOOM_FILTER 

barthr:
This PR fixes issue: #777


danielocfb:
Seems fine to me, thanks! Would you be able to add a test for this, by any chance? Also, can you please squash commits into a single one?

barthr:
Yes will do!

barthr:
So couple of things, when I started working on the tests I noticed that there is no way to do a proper lookup using the current lookup functionality. So I've added another function which should be used when u are using a bloom filter. I made it conform the spec as per (https://docs.kernel.org/bpf/map_bloom_filter.html). I also added some small refactoring. The PR turned out a little bit bigger than expected. Let me know what you think, I've only recently (this week) started to learn rust so it's all quite new to me.

libbpf-rs/src/map.rs
```diff
@@ -631,7 +632,14 @@ impl MapHandle {
     ///
     /// If the map is one of the per-cpu data structures, the function [`MapHandle::lookup_percpu()`]
     /// must be used.
+    /// If the map is of type bloom_filter the function [`MapHandle::lookup_bloom_filter()`] must be used
     pub fn lookup(&self, key: &[u8], flags: MapFlags) -> Result<Option<Vec<u8>>> {
+        if self.map_type().is_bloom_filter() {
+            return Err(Error::with_invalid_data(format!(
+                "lookup_bloom_filter() must be used for bloom_filter maps (type of the map is {})",
+                self.map_type(),
```
danielocfb, line 640:
```suggestion
            return Err(Error::with_invalid_data("lookup_bloom_filter() must be used for bloom filter maps"
```

danielocfb (approved):
I am unsure how exactly I feel about `lookup_bloom_filter`, but I suppose there is precedent, I don't see a better way, and the alternative of using `lookup` seems too....dodgy. So why not. Thanks!
Can you please take a look at the one suggestion I left?

barthr:
I get your view. I tried looking into staying as close as possible to existing calls and existing API's however it would make the lookup code a bit ugly and violating the principle of least suprises. 

So that is why I opted for the specific method for the bloom filter map. However, there might also be solutions which use a more polymorphism approach. Like a trait with a default lookup method with custom overrides for certain map types. I think for this PR this is out of scope.

Unfortunately, the libbpf api is a bit funky since the same methods mean different things on different types, that is the downside of making everything a map I suppose.

I will look at your comments tomorrow.

PR #786, opened 2024-05-12
Remove dependency on `num_enum`.

rbartlensky:
Hello! 

I am using `libbpf-rs` in one of my personal projects and I noticed that I was pulling in `toml-edit` via `libbpf-rs` which accounted for 1.3 seconds of my total build time. With the help of `cargo tree` I realized that `num_enum` is pulling in that dependency (and many others!) and I decided to try to remove it from `libbpf-rs` completely. Let me know what you think!

`cargo udeps` also noticed some extra things that can be removed, so I went ahead and removed those.

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -23,8 +23,6 @@ use std::slice::from_raw_parts;
 use bitflags::bitflags;
 use libbpf_sys::bpf_map_info;
 use libbpf_sys::bpf_obj_get_info_by_fd;
-use num_enum::IntoPrimitive;
-use num_enum::TryFromPrimitive;
 use strum_macros::Display;
 
 use crate::util;
@@ -67,10 +65,7 @@ impl OpenMap {
     /// Retrieve type of the map.
     pub fn map_type(&self) -> MapType {
         let ty = unsafe { libbpf_sys::bpf_map__type(self.ptr.as_ptr()) };
-        match MapType::try_from(ty) {
-            Ok(t) => t,
-            Err(_) => MapType::Unknown,
-        }
+        MapType::from(ty)
     }
 
     fn initial_value_raw(&self) -> (*mut u8, usize) {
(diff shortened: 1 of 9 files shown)


danielocfb (commented):
Thanks for the proposal! I am in favor of removing these dependencies, I am just not sure about the current implementation hardcoding the values. Can we instead rely on the symbolic values, ala:
```patch
--- libbpf-rs/src/btf/mod.rs
+++ libbpf-rs/src/btf/mod.rs
@@ -96,7 +96,7 @@ impl TryFrom<u32> for BtfKind {
         use BtfKind::*;

         Ok(match value {
-            0 => Void,
+            x if x == Void as u32 => Void,
```

libbpf-rs/src/map.rs
```diff
@@ -1066,6 +1060,55 @@ impl MapType {
...
+            23 => Stack,
+            24 => SkStorage,
+            25 => DevmapHash,
+            26 => StructOps,
+            27 => RingBuf,
+            28 => InodeStorage,
+            29 => TaskStorage,
+            30 => BloomFilter,
+            31 => UserRingBuf,
+            u32::MAX => Unknown,
```
danielocfb, line 1100: Can we remove this case? What purpose does it serve?

rbartlensky:
> Thanks for the proposal! I am in favor of removing these dependencies, I am just not sure about the current implementation hardcoding the values. Can we instead rely on the symbolic values, ala:
> 
> ```diff
> --- libbpf-rs/src/btf/mod.rs
> +++ libbpf-rs/src/btf/mod.rs
> @@ -96,7 +96,7 @@ impl TryFrom<u32> for BtfKind {
>          use BtfKind::*;
> 
>          Ok(match value {
> -            0 => Void,
> +            x if x == Void as u32 => Void,
> ```

I think that's even better! I made the changes suggested.

libbpf-rs/src/map.rs
rbartlensky: consistency with the "old" implementation, but to be fair it's useless, so I removed it

rbartlensky:
@danielocfb Thank you for taking the time to review! When you're happy with the PR I will squash the fixups away.

danielocfb (approved):
The changes look great now! Feel free to squash and we can go ahead with the merge. Thanks!

PR #789, opened 2024-05-17
Remove unsafe and platform dependent errno retrieval

amirbou:
The recent addition of user ring buffers included calls to `*libc::__errno_location()`.

Apart from requiring an unsafe block, this function might not be available on some targets, such as Android, causing build failures.

Use the portable implementation provided by `std::io::Error::last_os_error` instead.


danielocfb (approved):
Yeah, I was worried about this case, but failed to recollect in which context I had seen such issues. Looks good to me, thanks!

PR #799, opened 2024-06-05
Use test-tag crate for tagging root tests

d-e-s-o:
The way we specify the tests to run as root is non-obvious and error-prone: we just include "sudo" inside a tests name and then filter for that when running.
The test_tag crate provides first class support for tagging of tests, which makes what is going on a bit more obvious. Switch over to using it.


danielocfb:
I compared the numbers of tests run with the most recent `master` CI run. They are the same.

Issue #602, opened 2023-11-03
No bindings generated for types used in bpf maps


danielocfb:
We need to generate Rust types for types used in bpf maps. Currently we lack any such generation and that's causing folks to have to resort to using `bindgen` out-of-band, which is its own can of worms.

PR #809, opened 2024-06-20
Generate Rust types for types used in BPF maps

danielocfb:
Generate Rust correspondents for the types used in BPF maps. Having these types generated will eliminate the need for miscellaneous hacks to be employed to get the corresponding Rust types by other means.

Closes: #602

### DIFF
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -771,12 +771,25 @@ fn test_skeleton_builder_basic() {
         #include "vmlinux.h"
         #include <bpf/bpf_helpers.h>
 
+        struct unique_key {{
+            int cap;
+            u32 tgid;
+            u64 cgroupid;
+        }};
+
         struct {{
                 __uint(type, BPF_MAP_TYPE_HASH);
                 __uint(max_entries, 1024);
-                __type(key, u32);
+                __type(key, struct unique_key);
                 __type(value, u64);
         }} mymap SEC(".maps");
+
+        struct {{
+                __uint(type, BPF_MAP_TYPE_HASH);
+                __uint(max_entries, 1024);
+                __uint(key_size, 8);
+                __type(value, u64);
+        }} mymap2 SEC(".maps");
 
         SEC("kprobe/foo")
         int this_is_my_prog(u64 *ctx)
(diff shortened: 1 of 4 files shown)


anakryiko (commented):
Do you have an example skeleton.rs code generated? I'm failing to see how this all just auto-magically works for key/value types of map?...

danielocfb:
> Do you have an example skeleton.rs code generated?

I tried it with the `capable` skeleton.
https://gist.github.com/danielocfb/c73a4e12161ce2a888b410f78864fd3c
https://gist.github.com/danielocfb/046d0e8f3529f78a9e718ff462028b5d

anakryiko:
Ah, ok, so I see that we get a bunch of "garbage types" like `__anon_3`, which then transitively force libbpf-cargo to emit useful types, like `unique_key`. I think it would be best to avoid those anonymous types (which are just a detail on how we capture map definition). 

See https://github.com/libbpf/libbpf/blob/master/src/libbpf.h#L1024-L1025, you can directly ask for types that represent key/value of a map (they should be zero if there is no type association).

danielocfb:
Thanks for the suggestion. Adjusted the code accordingly. Please take another look when you get a chance, @anakryiko.

libbpf-cargo/src/test.rs
```diff
@@ -771,10 +771,16 @@ fn test_skeleton_builder_basic() {
...
+            int cap;
+            u32 tgid;
+            u64 cgroupid;
+        }};
+
         struct {{
                 __uint(type, BPF_MAP_TYPE_HASH);
                 __uint(max_entries, 1024);
-                __type(key, u32);
+                __type(key, struct unique_key);
```
anakryiko, line 783: as I mentioned above, it's legal to have `__uint(key_size, 8);`, maybe let's add another test for this to make sure this is handled gracefully?

anakryiko (changes requested):
This looks like a more cleaner approach, let's just make sure to not error out unnecessarily in some corner cases.

danielocfb:
> This looks like a more cleaner approach, let's just make sure to not error out unnecessarily in some corner cases.

Doesn't really seem necessary. ID 0 is [`void`](https://www.kernel.org/doc/html/next/bpf/btf.html#type-encoding). So it's a terminal type and we just ignore it. There is no error. Anyway, checking `== 0` now.

anakryiko (approved):
`__type(key, 8);` needs adjustment as discussed offline, but proactively approving as it overall looks good.

PR #812, opened 2024-06-24
Use different target names in USDT tests

danielocfb:
We probably shouldn't be sharing the same target name for the static probes created in the USDT tests, if for no other reason than it being stupidly confusing. Rename one of them.


danielocfb:
CI failures seem to be due to https://github.com/rust-lang/rustfmt/issues/6210

PR #813, opened 2024-06-24
Adjust gen_bpf_object_iter to use const* bpf_object

danielocfb:
There is seemingly no need for the gen_bpf_object_iter macro to work with a mutable bpf_object pointer. Use a const one instead and simplify a bunch of code in the process.


danielocfb:
CI failures seem to be due to https://github.com/rust-lang/rustfmt/issues/6210

PR #825, opened 2024-06-27
Introduce MapCore trait

danielocfb:
This change reworks our BPF map abstractions somewhat. Specifically, we remove the MapHandle part from a Map. These two are more or less separate concepts and having one Deref into the other, while perhaps convenient, is somewhat confusing. Perhaps more importantly, it leads to a bunch of shenanigans, such as:
- Map creation being a fallible operation, despite a properly created bpf_map being provided already
- Map objects effectively carrying heaps of duplicated information: the file descriptor, map type, key size and value size are all already contained in the additionally stored bpf_map object
- the existence of the conceptually unnecessary MapFd enum
- the arbitrary difference in implementation between OpenMap (only bpf_map pointer) and Map (bpf_map pointer + handle) versus OpenProgram (only bpf_program pointer) and Program (only bpf_program pointer)

As a result of this decoupling, we make the Map::new constructor infallible. To keep existing ergonomics, we unite common map related functionality in the newly introduced MapCore trait.

### DIFF
--- a/examples/tc_port_whitelist/src/main.rs
+++ b/examples/tc_port_whitelist/src/main.rs
@@ -10,6 +10,7 @@ use clap::Parser;
 
 use libbpf_rs::skel::OpenSkel;
 use libbpf_rs::skel::SkelBuilder;
+use libbpf_rs::MapCore as _;
 use libbpf_rs::MapFlags;
 use libbpf_rs::TcHookBuilder;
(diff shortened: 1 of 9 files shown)


danielocfb:
If there are any comments please leave them here and we can address them before the next release. Also, open to suggestions on naming of `MapCore` trait.

Anyway, going ahead with the merge to unblock follow on work.

PR #838, opened 2024-07-03
Remove a bunch of unnecessary (??) boxes

d-e-s-o:
Unclear why we need a bunch of boxes everywhere. It appears all we want is storage for a pointer. Gotta love the indirection.


d-e-s-o:
#YOLO

Issue #603, opened 2023-11-03
Include enums in generated types?


danielocfb:
Sort of orthogonal to https://github.com/libbpf/libbpf-rs/issues/602, we also got a generic request to include `enum` types in the set of types generated.

Reasoning:
> they aren't often referenced from types explicitly used in .bss/.data/maps

PR #867, opened 2024-07-24
libbpf-cargo: Generate Rust types for all BTF types

danielocfb:
Adjust the skeleton code to just generate Rust types for types present in the BTF of the object file being used. There is no real reason to artificially restrict types, as the BTF itself is already heavily trimmed down by the generation infrastructure. Yet, making all types present in it available to Rust users can reduce friction in usage, as everything that is available in C will be available in Rust as well.

Closes: #603

### DIFF
--- a/libbpf-cargo/CHANGELOG.md
+++ b/libbpf-cargo/CHANGELOG.md
@@ -2,8 +2,8 @@ Unreleased
 ----------
 - Reworked generated skeletons to contain publicly accessible maps and
   program members, no longer requiring method calls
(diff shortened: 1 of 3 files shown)


danielocfb:
There is a clash between our generated `struct_ops` type and the in-kernel one. Will have to see how to resolve that...

danielocfb:
Arm failures are b/s and have nothing to do with the change at hand from what I can tell...

danielocfb:
Anyway, these changes are probably ready for review now.

anakryiko (approved):
SGTM, sorry for delay!

PR #890, opened 2024-08-06
Make mutable ProgramInput.context_in

### LINKED ISSUE #888: Possible mutation of `context_in` constant when using `test_run` in `syscall` methods
The field `context_in` in `ProgramInput` structure now looks like this::
```rust
struct ProgramInput<'dat> {
  context_in: Option<&'dat [u8]>,
  // ...
}
```

But this field can be overwritten by kernel when calling `syscall` method from bpf. [File `net/bpf/test_run.c` at line 1567](https://github.com/torvalds/linux/blob/defaf1a2113a22b00dfa1abc0fd2014820eaf065/net/bpf/test_run.c#L1567): kernel is copy back `context_in` to user space after calling `syscall`. And this is the only way to return some data from `syscall` (something bigger than `i32`).

Perhaps the `context_in` field should have the signature `Option<&'dat mut [u8]>` or special attention should be added to the comments of `test_run` method and `context_in` field.

Example of valid bpf code:
```c++
SEC("syscall")
int mySyscall(int *ctx) {
  *ctx += *ctx;
  return 0;
}
```
And rust with `libbpf-rs`:
```rust
let value: i32 = 21;
let mut value = value.to_ne_bytes();

let mut input = ProgramInput::default();
input.context_in = Some(&mut value);

mySyscall.test_run(input).unwrap();

let value = i32::from_ne_bytes(value);
assert_eq!(value, 42);
```

x1b6e6:
Content of this field may change by using some programs (like `syscall`)

Fix #888

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -104,7 +104,6 @@ impl From<TracepointOpts> for libbpf_sys::bpf_tracepoint_opts {
     }
 }
 
-
 /// An immutable parsed but not yet loaded BPF program.
 pub type OpenProgram<'obj> = OpenProgramImpl<'obj>;
 /// A mutable parsed but not yet loaded BPF program.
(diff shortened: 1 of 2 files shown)


danielocfb (approved):
Looks good to me. Thanks

PR #899, opened 2024-08-08
Run tcp_ca example with leak sanitizer enabled

danielocfb:
Make sure to run our tcp_ca example with leak sanitizer enabled, so as to catch any potential memory leaks in the skeleton generation logic.


danielocfb:
Here is a run without https://github.com/libbpf/libbpf-rs/commit/c18616517a204b65d8e54e6bfb31c8a4ffd1030a that catches the leak that that commit addresses: https://github.com/danielocfb/libbpf-rs/actions/runs/10305416690/job/28526196182

Issue #909, opened 2024-08-14
BPF map `kconfig` does not have mmap pointer


GerardGarcia:
Apparently, when there is a reference to a kconfig variable and it is automatically mapped into a .kconfig map, the resulting skeleton builder cannot be opened and shows this error:

```
BPF map `kconfig` does not have mmap pointer
```

It seems it started happening in v0.24.0. It is easy reproducible because the `capable` example fails when it is executed:

```
❯ sudo ./target/debug/capable
thread 'main' panicked at /tmp/libbpf-rs/examples/capable/src/bpf/capable.skel.rs:110:26:
BPF map `kconfig` does not have mmap pointer
```

Some of my system info:
```
Linux testvm 6.9.0-1-aarch64-bpf-ARCH #2 SMP PREEMPT_DYNAMIC Mon Jul  8 09:47:30 CEST 2024 aarch64 GNU/Linux
rustc 1.80.1 (3f5fd8dd4 2024-08-06)
```

d-e-s-o:
Thanks for the report. I can reproduce.

danielocfb:
Comparing the skeletons we create we have always emitted accessors for `kconfig`:

With `0.23`:
```
        pub fn kconfig_raw(&self) -> *const capable_types::kconfig {
            self.skel_config
                .map_mmap_ptr(4)
                .unwrap()
                .cast::<capable_types::kconfig>()
        }
```

So we would have panicked there as well, but the access was optional, meaning that nobody hit the problem. Now with the member always being inlined we see the issue pop up at initialization time. Be that as it may, that suggests that our logic for identifying whether data is present is incorrect.

danielocfb:
This is the logic that we use for determining whether something is a "datasec" and, by implication, can be mmaped:

https://github.com/libbpf/libbpf-rs/blob/17d6cb8dfc9bba914c84c8d0e620e251fb8cacf2/libbpf-cargo/src/gen/mod.rs#L306-L316

In `bpftool` it seems a bit more...complex:
```c
static bool get_map_ident(const struct bpf_map *map, char *buf, size_t buf_sz)
{
	static const char *sfxs[] = { ".data", ".rodata", ".bss", ".kconfig" };
	const char *name = bpf_map__name(map);
	int i, n;

	if (!bpf_map__is_internal(map)) {
		snprintf(buf, buf_sz, "%s", name);
		return true;
	}

[...]

PR #911, opened 2024-08-14
libbpf-cargo: Don't populate datasec contents for kconfig map

danielocfb:
The .kconfig section is special in that libbpf sets its values as part of the load procedure. That means that until then, i.e., during open, contents are invalid. To that end the library doesn't even set the mmap pointer to anything but NULL, despite all heuristics pointing at it being mmapable and whatnot.
That's a problem for us, because we generate section contents and attempt to have them reference this nonexistent mmap area, which will fail.
Work around the issue by special casing this section and not permitting direct datasec access. If users require this access they have to manually iterate maps, retrieve initial value, and cast it to types::kconfig.

Closes: #909

### DIFF
--- a/libbpf-cargo/CHANGELOG.md
+++ b/libbpf-cargo/CHANGELOG.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+- Fixed panic on "open" of skeleton with `kconfig` map
+
+
 0.24.1
 ------
 - Fixed missing BPF object cleanup after skeleton destruction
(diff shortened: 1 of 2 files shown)


danielocfb:
Updated this PR to be a bit more lenient and provide `kconfig` data access after load.

danielocfb:
Going ahead with the merge to get this into a release. @anakryiko if you have any post-land comments do let me know.

PR #919, opened 2024-08-22
program: add get_fd_by_pin() api

vankoven:
After an eBPF program is loaded to the kernel and the loader process 
exits, a user may want to attach this program to PROG_ARRAY
of another one. This pull request adds a function to get a file
descriptor based on a pin file and guarantees that it represents a
program. Later, this descriptor can be used to be added to the
`PROG_ARRAY` or for creating an `XDP` program handle.

### DIFF
--- a/libbpf-rs/CHANGELOG.md
+++ b/libbpf-rs/CHANGELOG.md
@@ -1,3 +1,9 @@
+Unreleased
+----------
+- Added `Program::fd_from_pinned_path` method for restoring program descriptor
+  from a pinned path
+
+
 0.24.0
 ------
 - Split `{Open,}{Map,Program}` into `{Open,}{Map,Program}` (for shared
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -26,6 +26,7 @@ use libbpf_sys::bpf_func_id;
 
 use crate::util;
 use crate::util::validate_bpf_ret;
+use crate::util::BpfObjectType;
 use crate::AsRawLibbpf;
 use crate::Error;
 use crate::ErrorExt as _;
@@ -622,6 +623,35 @@ impl<'obj> Program<'obj> {
         };
         util::parse_ret(ret)?;
         Ok(prog_info.id)
+    }
+
+    /// Returns fd of a previously pinned program
+    ///
+    /// Returns error, if the pinned path doesn't represent an eBPF program.
+    pub fn fd_from_pinned_path<P: AsRef<Path>>(path: P) -> Result<OwnedFd> {
+        let path_c = util::path_to_cstring(&path)?;
+        let path_ptr = path_c.as_ptr();
+
+        let fd = unsafe { libbpf_sys::bpf_obj_get(path_ptr) };
+        let fd = util::parse_ret_i32(fd).with_context(|| {
+            format!(
+                "failed to retrieve BPF object from pinned path `{}`",
+                path.as_ref().display()
+            )
+        })?;
+        let fd = unsafe { OwnedFd::from_raw_fd(fd) };
+
+        // A pinned path may represent an object of any kind, including map
+        // and link. This may cause unexpected behaviour for following functions,
+        // like bpf_*_get_info_by_fd(), which allow objects of any type.
+        let fd_type = util::object_type_from_fd(fd.as_fd())?;
+        match fd_type {
+            BpfObjectType::Program => Ok(fd),
+            other => Err(Error::with_invalid_data(format!(
+                "retrieved BPF fd is not a program fd: {:#?}",
+                other
+            ))),
+        }
     }
 
     /// Returns flags that have been set for the program.
--- a/libbpf-rs/src/util.rs
+++ b/libbpf-rs/src/util.rs
@@ -1,12 +1,16 @@
 use std::ffi::CStr;
 use std::ffi::CString;
+use std::fs;
 use std::mem::transmute;
 use std::ops::Deref;
+use std::os::fd::AsRawFd;
+use std::os::fd::BorrowedFd;
 use std::os::raw::c_char;
 use std::path::Path;
 use std::ptr::NonNull;
 use std::sync::OnceLock;
 
+use crate::error::IntoError;
 use crate::Error;
 use crate::Result;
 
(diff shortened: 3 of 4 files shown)


libbpf-rs/src/program.rs
```diff
@@ -624,6 +624,30 @@ impl<'obj> Program<'obj> {
         Ok(prog_info.id)
     }
 
+    /// Returns program fd by pin
+    pub fn get_fd_by_pin<P: AsRef<Path>>(path: P) -> Result<OwnedFd> {
+        let path_c = util::path_to_cstring(path)?;
+        let path_ptr = path_c.as_ptr();
+        let fd = util::parse_ret_i32(unsafe { libbpf_sys::bpf_obj_get(path_ptr) })?;
```
danielocfb, line 631: Can you please add context to fallible calls (not just this one):
```suggestion
        let fd = util::parse_ret_i32(unsafe { libbpf_sys::bpf_obj_get(path_ptr) })
            .with_context(|| format!("failed to retrieve BPF object from path `{}`", path.as_ref().display()))?;
```
(may not compile as-is but you get the idea)

libbpf-rs/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Unreleased
+----------
+- Added `Program::get_fd_by_pin` method for restoring program descriptor
+  from pin path
```
danielocfb, line 4:
```suggestion
  from pinned path
```

libbpf-rs/src/program.rs
```diff
@@ -624,6 +624,30 @@ impl<'obj> Program<'obj> {
         Ok(prog_info.id)
     }
 
+    /// Returns program fd by pin
+    pub fn get_fd_by_pin<P: AsRef<Path>>(path: P) -> Result<OwnedFd> {
```
danielocfb, line 628: I understand why you named the method this way but the existing `..._by_...` naming is horrible, in my opinion. What do you think about `fd_from_pinned_path`, analogous to https://docs.rs/libbpf-rs/0.24.2/libbpf_rs/struct.MapHandle.html#method.from_pinned_path?

libbpf-rs/src/program.rs
```diff
@@ -624,6 +624,30 @@ impl<'obj> Program<'obj> {
...
+        let mut info: libbpf_sys::bpf_prog_info = unsafe { mem::zeroed() };
+        let mut size = mem::size_of_val(&info) as u32;
+        // SAFETY: All pointers are derived from references and hence valid.
+        util::parse_ret(unsafe {
+            libbpf_sys::bpf_prog_get_info_by_fd(
+                fd.as_raw_fd(),
+                &mut info as *mut libbpf_sys::bpf_prog_info,
+                &mut size as *mut u32,
+            )
+        })?;
```
danielocfb, line 646: Nit: Can we use an intermediate variable binding for the return value of `bpf_prog_get_info_by_fd`, please?

danielocfb (commented):
I think this looks great for the most part, thanks! Long term we may actually want to introduce a `ProgramHandle` type to mirror the existing `MapHandle` (and perhaps same for `ProgramInfo`), but it's clearly out of scope here. Left a few comments. Let me know what you think.

libbpf-rs/src/program.rs
vankoven: Added.

libbpf-rs/CHANGELOG.md
vankoven: Fixed.

libbpf-rs/src/program.rs
vankoven: Renamed as proposed. I think `..._by_...` was somehow influenced by `bpf.h`, and I agree, `from` is more natural here.

libbpf-rs/src/program.rs
vankoven: Added, and there were a couple of cases like that. I also changed them the same way for better readability.

vankoven:
Thank you for the review!

I have to rework this PR significantly. To my surprise, `bpf_prog_get_info_by_fd()` is happy to eat descriptors of maps or links and fill the info structure based on the object's actual type. It's totally okay to push a `map` fd into `bpf_PROG_info()` and get the map information inside the prod data structure. So, the caller is responsible for knowing what exactly was pinned before and using the correct data structures.

I tried to dig into [how bpftool works with that case](https://github.com/libbpf/bpftool/blob/64402b8ea43b2ef62ee8ba71d9b50a5e32603fd2/src/common.c#L451) and found no alternatives. I decided the' query' module was the best place for that function.

> Long term we may actually want to introduce a `ProgramHandle`

Oh, I agree here. There are so many possibilities that you can still do with only the program descriptor, and recovering the initial `bpf_program` structure is not possible.
The `Xdp` structure already implements a minimum number of actions for XDP programs, but it's limited in scope. I will keep in mind that implementation of the `ProgramHandle` is welcomed.

libbpf-rs/src/program.rs
```diff
@@ -624,6 +626,35 @@ impl<'obj> Program<'obj> {
...
+        let fd = unsafe { OwnedFd::from_raw_fd(fd) };
+
+        // A pinned path may represent an object of any kind, including map
+        // and link. This may cause unexpected behaviour for following functions,
+        // like bpf_*_get_info_by_fd(), which allow objects of any type.
+        let fd_type = query::object_type_from_fd(fd.as_fd())?;
+        match fd_type {
+            BpfObjectType::Program => Ok(fd),
+            other => Err(Error::with_invalid_data(format!(
+                "Retrived BPF fd is not a program fd: {:#?}",
```
d-e-s-o, line 652:
```suggestion
                "Retrieved BPF fd is not a program fd: {:#?}",
```

libbpf-rs/src/query.rs
```diff
@@ -703,3 +706,39 @@ gen_info_impl!(
...
+        .to_str()
+        .ok_or_invalid_data(|| "Can't convert PathBuf to str")?;
+
+    match link_type {
+        "anon_inode:bpf-link" => Ok(BpfObjectType::Link),
+        "anon_inode:bpf-map" => Ok(BpfObjectType::Map),
+        "anon_inode:bpf-prog" => Ok(BpfObjectType::Program),
+        other => Err(Error::with_invalid_data(format!(
+            "Unknown type of BPF fd: {}",
+            other
```
d-e-s-o, line 741:
```suggestion
            "Unknown type of BPF fd: {other}",
```

libbpf-rs/src/query.rs
```diff
@@ -703,3 +706,39 @@ gen_info_impl!(
...
+    /// The object is a BPF link.
+    Link,
+}
+
+/// Get type of BPF object by fd.
+///
+/// This information is not exported directly by bpf_*_get_info_by_fd() functions,
+/// as kernel relies on the userspace code to know what kind of object it
+/// queries. The type of object can be recovered by fd only from the proc
+/// filesystem.The same approach is used in bpftool.
```
d-e-s-o, line 726:
```
/// filesystem. The same approach is used in bpftool.
```

libbpf-rs/src/query.rs
```diff
@@ -703,3 +706,39 @@ gen_info_impl!(
...
+pub fn object_type_from_fd(fd: BorrowedFd<'_>) -> Result<BpfObjectType> {
```
d-e-s-o, line 727: Could you also add a test making sure that this function specifically fails as expected when used on fd derived from a file system `File`?

d-e-s-o, line 727: Let's maybe not make this function public? We can use it as implementation detail for higher level APIs targetting the individual objects, but I don't think we should expose it. It just needlessly adds to the API surface, in my opinion.

d-e-s-o (approved):
Nice! Thank you! Left a few more comments. The only additional meta-comment I have is that I'd prefer we lowercase all error messages. Right now you introduce a mixture of cases:
- `"failed to retrieve BPF...`
- `"Unknown type of BPF...`

I am trying to make everything uniformly start lowercased, which I believe to be more universally accepted practice.

Please take another look and we can merge once addressed.

libbpf-rs/src/program.rs
vankoven: Fixed.

libbpf-rs/src/query.rs
vankoven: fixed.

libbpf-rs/src/query.rs
vankoven: Added a unit test for fd from a usual file or stdout.

vankoven: I moved it to the utils module, which is internal and doesn't expose the function publically (and updated the changelog to remove announce of that new public function).

vankoven:
> I'd prefer we lowercase all error messages

Also fixed that.

d-e-s-o (approved):
Looks great to me, thanks again!

d-e-s-o:
Let me know in case you need a new version to be released with this functionality.

PR #939, opened 2024-09-09
feat(perf_buffer): perf_buffer relies on traits rather than specific type

CoderPoet:
#### Enhancement
This PR introduces a flexible and generic implementation for `PerfBufferBuilder` to support any type that implements the `MapCore` and `AsFd` traits, rather than being restricted to a specific `Map` type. 

#### Why This Change is Needed:
Currently, `PerfBufferBuilder` is limited to a specific `Map` type, which restricts its usability with other types that implement the required traits. This update increases the extensibility of the `PerfBufferBuilder` and allows developers to use it with a wider variety of map-like structures. For example, a scenario that uses a pinning map.


d-e-s-o (approved):
Seems fine to me, thanks!

PR #943, opened 2024-09-13
libbpf-rs: program: Show AsFd in documentation

danobi:
Previously, AsFd trait was implemented on ProgramImpl. Since ProgramImpl is not exported, the trait implementation did not show up on docs.rs. You could still use it - but it wasn't obvious that it was available.

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -564,6 +564,13 @@ pub struct ProgramImpl<'obj, T = ()> {
     _phantom: PhantomData<&'obj T>,
 }
 
+impl<T> ProgramImpl<'_, T> {
+    fn as_fd(&self) -> BorrowedFd<'_> {
+        let fd = unsafe { libbpf_sys::bpf_program__fd(self.ptr.as_ptr()) };
+        unsafe { BorrowedFd::borrow_raw(fd) }
+    }
+}
+
 impl<'obj> Program<'obj> {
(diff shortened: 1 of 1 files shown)


d-e-s-o (commented):
I think there are a bunch of cases where stuff doesn't show up at all or doesn't show up correctly in the documentation, especially after the recent introduction of the `Mut` variants. Problem is, I think that more than anything these are issues with `rustdoc`. I opened one or two issues in the past about problems in that realm, but since they haven't gotten any traction I stopped filing.
The change is fine in principle, but I am somewhat reluctant to merge it: we are introducing unnecessary complexity/duplication just to work around what I'd think is a bug (though I'd be happy to be convince of the contrary). If we merge that, we should merge every other similar fix up. And I am not sure where to draw the line once more complex/ugly/<...> work arounds are necessary.
If you feel strongly about fixing this and you agree it's a bug with the doc rendering, here would be my suggestion: open an issue and then reference this issue in a TODO at the two `AsFd` impl sites as a reminder to dedup things once fixed upstream. Then I'd be much less reluctatnt to merge the change. What do you think?

danobi:
Yeah fair enough. Let’s put a pin in this for now. I’ll see if I can hack rustdoc to make this properly work

danobi:
Can you share some of the issues you opened?

d-e-s-o:
The only one in that realm that I found with a quick search is https://github.com/rust-lang/rust/issues/116854

danobi:
Alright, I think I found the "official" solution

PR #944, opened 2024-09-13
Gate out test functions that use probe! on ppc64le and s390x

michel-slm:
These require `feature(asm_experimental_arch)` which only works on nightly compilers.

See https://github.com/rust-lang/rust/issues/93335

Also loosen version requirement on `probe`; on Fedora we have been building with `probe` 0.5.1 with no issue.

Tested by doing a scratch build on Fedora's Koji build system:

https://koji.fedoraproject.org/koji/taskinfo?taskID=123361398

### DIFF
--- a/libbpf-rs/Cargo.toml
+++ b/libbpf-rs/Cargo.toml
@@ -48,7 +48,7 @@ libbpf-rs = {path = ".", features = ["generate-test-files"]}
 log = "0.4.4"
 memmem = "0.1.1"
 plain = "0.2.3"
-probe = "0.3"
+probe = ">=0.3, <0.6"
 scopeguard = "1.1"
 serial_test = { version = "3.0", default-features = false }
 tempfile = "3.3"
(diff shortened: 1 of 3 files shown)


libbpf-rs/Cargo.toml
```diff
@@ -48,7 +48,7 @@ libbpf-rs = {path = ".", features = ["generate-test-files"]}
 log = "0.4.4"
 memmem = "0.1.1"
 plain = "0.2.3"
-probe = "0.3"
+probe = ">=0.3, <0.6"
```
d-e-s-o, line 51: Let's just bump to `0.5` then? Nobody is really testing any version ranges, so this is hard to test for correctness. It's also very atypical for Rust projects from what I understand.

libbpf-rs/tests/common/mod.rs
```diff
@@ -9,7 +9,6 @@ use libbpf_rs::ObjectBuilder;
 use libbpf_rs::OpenObject;
 use libbpf_rs::ProgramMut;
 
-
```
d-e-s-o, line 12: Would you mind reverting the spurious formatting change?

d-e-s-o (commented):
This change seems fine, thanks. Left two comments; please let me know what you think.

PR #947, opened 2024-09-17
Fully generate docs

danobi:
Follow-on from https://github.com/libbpf/libbpf-rs/pull/943.

See the second commit for how/why this works.


d-e-s-o (approved):
Awesome, thanks! One question: would it still work if we were to tag it `#[doc(hidden)]` (I believe that's what it's called). I feel like that would be even better.

Issue #940, opened 2024-09-11
Mark Link with must_use


lbrndnr:
Hi,

I ran into an issue where my sockops program is immediately detached because I didn't actively ignore its link, as the following code shows:
```
skel.progs_mut()
    .sock_ops()
    .attach_cgroup(cgroup_fd)?;
```

Given that this problem is super easy to miss, would it make sense to mark `Link` with `must_use`? What do you think? :)

danielocfb:
Yes, that's probably a good idea. Feel free to open a pull request. Thanks!

d-e-s-o:
Closed by https://github.com/libbpf/libbpf-rs/commit/d97de7646b76bdd00216a28a3da6c2d7268e9261.

PR #951, opened 2024-09-24
Mark link with must_use

lbrndnr:
This is a small fix that marks Link with `must_use`, as discussed in #940. This makes it clear right away that the program will detach immediately if not used.


d-e-s-o:
Thank you!

PR #958, opened 2024-10-07
Fix __TARGET_ARCH_xx define for riscv64,loongarch64,sparc64,mips64

kxxt:
This commit updates the __TARGET_ARCH_xx define to match what is used in bpf_tracing.h.

Reference https://github.com/torvalds/linux/blob/8cf0b93919e13d1e8d4466eb4080a4c4d9d66d7b/tools/lib/bpf/bpf_tracing.h#L8-L38

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -190,6 +190,10 @@ fn compile_one(
             "aarch64" => "arm64",
             "powerpc64" => "powerpc",
             "s390x" => "s390",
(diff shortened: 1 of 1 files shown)


d-e-s-o (approved):
This seems fine, but I think we should also update https://github.com/libbpf/libbpf-rs/blob/704543e256db470796f028b9f5f61ca3e8389afe/libbpf-rs/build.rs#L189-L193

Will take care of it.

PR #959, opened 2024-10-07
Use runtime CARGO_CFG_TARGET_ARCH in build-script

kxxt:
option_env retrieves the CARGO_CFG_TARGET_ARCH at compile-time, which is not correct for the usage here. Instead, we should retrieve the variable using env::var at runtime in the build-script.

Fix https://github.com/kxxt/tracexec/actions/runs/11213953908/job/31168026074?pr=41

See also #958


d-e-s-o (approved):
Looks good to me, thanks!

PR #964, opened 2024-10-09
Use vmlinux.h repository

d-e-s-o:
Integration with https://github.com/libbpf/vmlinux.h/pull/1. Please see individual commits for descriptions.


libbpf-rs/Cargo.toml
```diff
@@ -42,6 +46,7 @@ vsprintf = "2.0"
 [build-dependencies]
 libbpf-sys = { version = "1.4.1", default-features = false, optional = true }
 tempfile = { version = "3.3", optional = true }
+vmlinux = { git = "https://github.com/d-e-s-o/vmlinux.h.git", branch = "topic/init", optional = true }
```
danielocfb, line 49: It seems that this may actually prevent publishing the crate. Sigh. Will have to check if there is a way around that limitation...

kxxt: Probably by moving generate-test-files out of `build.rs` into a binary crate and then we can move vmlinux to `dev-dependencies`.

danielocfb: I think we need it to run as part of the build script, though, otherwise when do we actually generate the necessary files?

danielocfb: Found a way.

danielocfb: After sleeping over it, I believe I understand now what you were getting at. Yeah, that should hopefully make things a bit simpler.

PR #967, opened 2024-10-11
Disable LeakSanitizer tests

danielocfb:
Leak sanitizer infrastructure in Rust is badly broken [0]. Disable it until it resembles something of a working piece of software again.

[0] https://github.com/rust-lang/rust/issues/111073


d-e-s-o:
Seems to be no longer necessary now...

PR #968, opened 2024-10-16
libbpf-rs: Remove unused `log` dependency

javierhonduco:
libbpf-rs: Remove unused log dependency


javierhonduco:
Turns out this was used in doctests.

PR #975, opened 2024-10-22
libbpf-cargo: Handle empty unions in BTF types

### LINKED ISSUE #974: libbpf-cargo: Build fails when a C struct contains an empty union
I am using `libbpf-cargo` version `0.24.6`. I'm trying to use `struct bio` like this in my `.bpf.c`:

```c
SEC("fentry/submit_bio")
int BPF_PROG(submit_bio, struct bio *bio)
{
...
	vcnt = bio->bi_vcnt;
...
}
```

However `cargo build` fails with:

```
  thread 'main' panicked at /home/jrc/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libbpf-cargo-0.24.6/src/gen/btf.rs:769:49:
  index out of bounds: the len is 0 but the index is 0
```

The error refers to [this line](https://github.com/libbpf/libbpf-rs/blob/f73d75c02e4f8fc33dc4809b3b69f4a660c0f90a/libbpf-cargo/src/gen/btf.rs#L769), which is part of the code to "write a Default implementation for a union".

At [vmlinux_601.h#L7735](https://github.com/libbpf/vmlinux.h/blob/172793d6a409d98d1cfb843c80df73733e9f832f/include/x86/vmlinux_601.h#L7735), we find this:

```c
struct bio {
...
	union {};
...
};
```

It looks like empty unions cause problems for `libbpf-cargo`. Arguably this is an issue of `vmlinux.h`, and I am aware that the `vmlinux.h` repo is advertised as "Development-only". This issue could be fixed by removing the empty union in `vmlinux.h`. However, it seems more convenient to me to add support for empty unions to `libbpf-cargo`. Is that reasonable?

JoelColledge:
This was noticed because a "vmlinux.h" contained:

```
struct bio {
...
	union {};
...
};
```

Which caused the following error when trying to use "struct bio" in a .bpf.c file:

```
  thread 'main' panicked at /home/jrc/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libbpf-cargo-0.24.6/src/gen/btf.rs:769:49:
  index out of bounds: the len is 0 but the index is 0
```

Fix this by hiding empty unions from the generated types. Rust does not allow union types with zero fields.

Fixes #974.

### DIFF
--- a/libbpf-cargo/CHANGELOG.md
+++ b/libbpf-cargo/CHANGELOG.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+- Fixed handling of empty unions in BPF types
+
+
 0.24.6
 ------
 - Fixed incorrect Cargo environment variable query when used in build
--- a/libbpf-cargo/src/gen/btf.rs
+++ b/libbpf-cargo/src/gen/btf.rs
@@ -578,6 +578,11 @@ impl<'s> GenBtf<'s> {
         t: types::Composite<'_>,
         opts: &TypeDeclOpts,
     ) -> Result<()> {
+        if t.is_empty_union() {
+            // Ignore empty union; Rust does not allow unions with no fields.
+            return Ok(());
+        }
+
         let packed = is_struct_packed(&t, &self.btf)?;
 
         // fields in the aggregate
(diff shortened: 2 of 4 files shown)


danielocfb (approved):
Looks good to me, thanks!

PR #988, opened 2024-11-03
Update goblin dependency to 0.9

musicinmybrain:
Update the dependency on `goblin` from 0.6 to 0.9.

The trivial use of this crate in a test for `libbpf-cargo` is not affected by any of the [documented breaking changes](https://github.com/m4b/goblin/blob/master/CHANGELOG.md) between these releases.

Furthermore, running `cargo test` in `libbpf-cargo/` still works.


danielocfb:
Looks good to me, thanks! In the future, please change related functionality in a single commit. There is no reason to have two commits here and it can cause issue when bisecting, for example, because the `Cargo.lock` file will be regenerated and have untracked changes.

danielocfb (approved):

PR #989, opened 2024-11-04
libbpf-cargo: Generate C enums as custom types with const fields

### LINKED ISSUE #982: libbpf-cargo: enums with duplicated values cause compilation issues
In https://github.com/libbpf/libbpf-rs/commit/4e62c38322cced90ad673d25e8253f47ad37c566, Rust type generation changed from the used types to all types. This is mostly harmless, except in the case were various enum entries share the same value. This is perfectly valid in C and[ this happens in the `bpf_map_type` enum](https://github.com/torvalds/linux/blob/98f7e32f20d28ec452afb208f9cffc08448a2652/include/uapi/linux/bpf.h#L984). The error currently get:

```
error[E0081]: discriminant value `19` assigned more than once
   --> lightswitch-capabilities/src/bpf/features_skel.rs:439:9
    |
439 |         pub enum bpf_map_type {
    |         ^^^^^^^^^^^^^^^^^^^^^
...
460 |             BPF_MAP_TYPE_CGROUP_STORAGE_DEPRECATED = 19,
    |                                                      -- `19` assigned here
461 |             BPF_MAP_TYPE_CGROUP_STORAGE = 19,
    |                                           -- `19` assigned here
```

I've verified that the parent commit https://github.com/libbpf/libbpf-rs/commit/65094bc5d40186cc83d7017ad1b77feed47f1927 does not show this issue. Perhaps rather than [emitting Rust enums](https://github.com/libbpf/libbpf-rs/blob/a0577750e54710b8949c65bff1321ede5f7ea2ed/libbpf-cargo/src/gen/btf.rs#L815) we could declare them as constants? Opening this issues mostly so I don't forget. I'll be happy to submit a PR if this approach sounds reasonable. 

cc @d-e-s-o

javierhonduco:
Rather than generating Rust enums from C enums, which might not be correct, as for example, C allows enum variants to share a value while Rust does not.

This fixes #982.

Test Plan
=========

Verified that tests pass in CI in my fork and that my project compiles fine and works will with this change.

### DIFF
--- a/libbpf-cargo/src/gen/btf.rs
+++ b/libbpf-cargo/src/gen/btf.rs
@@ -797,6 +797,8 @@ impl<'s> GenBtf<'s> {
             _ => bail!("Invalid enum size: {}", t.size()),
         };
 
+        let enum_name = self.anon_types.type_name_or_anon(&t);
+
         let mut signed = "u";
         for value in t.iter() {
             if value.value < 0 {
@@ -805,30 +807,37 @@ impl<'s> GenBtf<'s> {
             }
         }
 
-        writeln!(
-            def,
-            r#"#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]"#
-        )?;
-        writeln!(def, r#"#[repr({signed}{repr_size})]"#)?;
-        writeln!(
-            def,
-            r#"pub enum {name} {{"#,
-            name = self.anon_types.type_name_or_anon(&t),
-        )?;
+        let mut first_field = None;
 
-        for (i, value) in t.iter().enumerate() {
-            if i == 0 {
-                writeln!(def, r#"    #[default]"#)?;
-            }
+        writeln!(def, r#"#[derive(Debug, Copy, Clone)]"#)?;
+        writeln!(def, r#"#[repr(transparent)]"#)?;
+        writeln!(def, r#"pub struct {enum_name}({signed}{repr_size});"#)?;
+        writeln!(def, "#[allow(non_upper_case_globals)]")?;
+        writeln!(def, r#"impl {enum_name} {{"#,)?;
+
+        for value in t.iter() {
+            first_field = first_field.or(Some(value));
+
             writeln!(
                 def,
-                r#"    {name} = {value},"#,
+                r#"    pub const {name}: {enum_name} = {enum_name}({value});"#,
                 name = value.name.unwrap().to_string_lossy(),
                 value = value.value,
             )?;
         }
 
-        writeln!(def, "}}")?;
+        writeln!(def, r#"}}"#)?;
+
+        if let Some(first_field) = first_field {
+            writeln!(def, r#"impl Default for {enum_name} {{"#)?;
+            writeln!(
+                def,
+                r#"    fn default() -> Self {{ {enum_name}::{name} }}"#,
+                name = first_field.name.unwrap().to_string_lossy()
+            )?;
+            writeln!(def, r#"}}"#)?;
+        }
+
         Ok(())
     }
 
--- a/libbpf-cargo/CHANGELOG.md
+++ b/libbpf-cargo/CHANGELOG.md
@@ -1,6 +1,7 @@
 0.24.7
 ------
 - Fixed handling of empty unions in BPF types
+- Represent C enums with custom types and const fields
 
 
 0.24.6
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -1817,20 +1817,26 @@ fn test_btf_dump_definition_enum() {
 enum Foo {
     Zero = 0,
     One,
-    seven = 7,
+    Seven = 7,
+    ZeroDup = Zero,
 };
 
 enum Foo foo;
 "#;
 
     let expected_output = r#"
-#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
-#[repr(u32)]
-pub enum Foo {
-    #[default]
-    Zero = 0,
-    One = 1,
-    seven = 7,
+#[derive(Debug, Copy, Clone)]
+#[repr(transparent)]
+pub struct Foo(u32);
+#[allow(non_upper_case_globals)]
+impl Foo {
+    pub const Zero: Foo = Foo(0);
+    pub const One: Foo = Foo(1);
+    pub const Seven: Foo = Foo(7);
+    pub const ZeroDup: Foo = Foo(0);
+}
+impl Default for Foo {
+    fn default() -> Self { Foo::Zero }
 }
 "#;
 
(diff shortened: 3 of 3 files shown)


libbpf-cargo/src/gen/btf.rs
```diff
@@ -805,30 +807,22 @@ impl<'s> GenBtf<'s> {
...
-            r#"#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]"#
-        )?;
-        writeln!(def, r#"#[repr({signed}{repr_size})]"#)?;
-        writeln!(
-            def,
-            r#"pub enum {name} {{"#,
-            name = self.anon_types.type_name_or_anon(&t),
-        )?;
+        writeln!(def, r#"#[derive(Debug, Copy, Clone, Default)]"#)?;
+        writeln!(def, r#"pub struct {enum_name}({signed}{repr_size});"#,)?;
```
danielocfb, line 811: Let's tag this type as transparent, please:
```suggestion
        writeln!(def, r#"#[derive(Debug, Copy, Clone, Default)]"#)?;
        writeln!(def, r#"#[repr(transparent)]"#)?;
        writeln!(def, r#"pub struct {enum_name}({signed}{repr_size});"#)?;
```
(or so)

libbpf-cargo/src/gen/btf.rs
```diff
@@ -805,30 +807,22 @@ impl<'s> GenBtf<'s> {
...
 
-        for (i, value) in t.iter().enumerate() {
-            if i == 0 {
-                writeln!(def, r#"    #[default]"#)?;
```
danielocfb, line 821: I think you are inadvertently changing the default to something that is not necessarily a valid constant and I believe we should preserve the existing logic instead. You will probably have to remember the name of the first constant and then generate a `Default` impl that returns it.

danielocfb (commented):
Thanks for working on this fix! Overall this looks great, but I think I spotted an issue with the handling of default values. Please take another look. Also, can you please add a test for the generation of a Rust enum for a C correspondent with two variants of the same value?

libbpf-cargo/src/gen/btf.rs
```diff
@@ -805,30 +807,22 @@ impl<'s> GenBtf<'s> {
...
-            }
+        writeln!(def, r#"impl {enum_name} {{"#,)?;
+
+        for value in t.iter() {
+            writeln!(def, "    #[allow(non_upper_case_globals)]")?;
```
danielocfb, line 816: Consider moving this hint to the overall type level instead of emitting it for each variant/constant.

javierhonduco:
Apologies for the delay and thanks for the feedback. Applied all the changes and added an integration tests to ensure compilation succeeds. I've also re-tested this branch with my project and it compiles fine. Just for completeness, the type that was causing issues before:

<details>

```rust
        #[derive(Debug, Copy, Clone)]
        #[repr(transparent)]
        pub struct bpf_map_type(u32);
        #[allow(non_upper_case_globals)]
        impl bpf_map_type {
            pub const BPF_MAP_TYPE_UNSPEC: bpf_map_type = bpf_map_type(0);
            pub const BPF_MAP_TYPE_HASH: bpf_map_type = bpf_map_type(1);
            pub const BPF_MAP_TYPE_ARRAY: bpf_map_type = bpf_map_type(2);
            pub const BPF_MAP_TYPE_PROG_ARRAY: bpf_map_type = bpf_map_type(3);
            pub const BPF_MAP_TYPE_PERF_EVENT_ARRAY: bpf_map_type = bpf_map_type(4);
            pub const BPF_MAP_TYPE_PERCPU_HASH: bpf_map_type = bpf_map_type(5);
            pub const BPF_MAP_TYPE_PERCPU_ARRAY: bpf_map_type = bpf_map_type(6);
            pub const BPF_MAP_TYPE_STACK_TRACE: bpf_map_type = bpf_map_type(7);
            pub const BPF_MAP_TYPE_CGROUP_ARRAY: bpf_map_type = bpf_map_type(8);
            pub const BPF_MAP_TYPE_LRU_HASH: bpf_map_type = bpf_map_type(9);
            pub const BPF_MAP_TYPE_LRU_PERCPU_HASH: bpf_map_type = bpf_map_type(10);
            pub const BPF_MAP_TYPE_LPM_TRIE: bpf_map_type = bpf_map_type(11);
            pub const BPF_MAP_TYPE_ARRAY_OF_MAPS: bpf_map_type = bpf_map_type(12);
            pub const BPF_MAP_TYPE_HASH_OF_MAPS: bpf_map_type = bpf_map_type(13);
            pub const BPF_MAP_TYPE_DEVMAP: bpf_map_type = bpf_map_type(14);
            pub const BPF_MAP_TYPE_SOCKMAP: bpf_map_type = bpf_map_type(15);
            pub const BPF_MAP_TYPE_CPUMAP: bpf_map_type = bpf_map_type(16);
            pub const BPF_MAP_TYPE_XSKMAP: bpf_map_type = bpf_map_type(17);
            pub const BPF_MAP_TYPE_SOCKHASH: bpf_map_type = bpf_map_type(18);
            pub const BPF_MAP_TYPE_CGROUP_STORAGE_DEPRECATED: bpf_map_type = bpf_map_type(19);
            pub const BPF_MAP_TYPE_CGROUP_STORAGE: bpf_map_type = bpf_map_type(19);
            pub const BPF_MAP_TYPE_REUSEPORT_SOCKARRAY: bpf_map_type = bpf_map_type(20);
            pub const BPF_MAP_TYPE_PERCPU_CGROUP_STORAGE: bpf_map_type = bpf_map_type(21);
            pub const BPF_MAP_TYPE_QUEUE: bpf_map_type = bpf_map_type(22);
            pub const BPF_MAP_TYPE_STACK: bpf_map_type = bpf_map_type(23);
            pub const BPF_MAP_TYPE_SK_STORAGE: bpf_map_type = bpf_map_type(24);
            pub const BPF_MAP_TYPE_DEVMAP_HASH: bpf_map_type = bpf_map_type(25);
            pub const BPF_MAP_TYPE_STRUCT_OPS: bpf_map_type = bpf_map_type(26);
            pub const BPF_MAP_TYPE_RINGBUF: bpf_map_type = bpf_map_type(27);
            pub const BPF_MAP_TYPE_INODE_STORAGE: bpf_map_type = bpf_map_type(28);
            pub const BPF_MAP_TYPE_TASK_STORAGE: bpf_map_type = bpf_map_type(29);
            pub const BPF_MAP_TYPE_BLOOM_FILTER: bpf_map_type = bpf_map_type(30);
            pub const BPF_MAP_TYPE_USER_RINGBUF: bpf_map_type = bpf_map_type(31);
            pub const BPF_MAP_TYPE_CGRP_STORAGE: bpf_map_type = bpf_map_type(32);
        }
        impl Default for bpf_map_type {
            fn default() -> Self {
                bpf_map_type::BPF_MAP_TYPE_UNSPEC
            }
        }
```
</details>

javierhonduco:
(Pushed another update to write Rust compiler generated code to a temp folder and to fix a lint that I didn't see with my local version of Clippy)

danielocfb (approved):
Looks good to me, thanks. I think I'll cut another patch release before merging this change, because I believe we have to treat it as breaking.

javierhonduco:
Thanks! :)

PR #994, opened 2024-11-06
libbpf-cargo: return stderr from clang

JakeHillion:
Update `build_and_generate` to return a `CompilationOutput` owning the stderr
of the compiler process. This gives consumers access to the stderr to output as
they see fit.

Provides a `stderr` method that calls `String::from_utf8_lossy`. The primary
consumer is `sched-ext/scx` which outputs each line with
`println!("cargo:warning={}", l)`, showing the warnings quite nicely in the
Cargo output when there isn't an error.

Test plan:
- Pointed that Cargo.lock at this repo and built.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -19,6 +19,21 @@ use tempfile::tempdir;
 use crate::metadata;
 use crate::metadata::UnprocessedObj;
 
+/// Contains information about a successful compilation.
+#[derive(Debug)]
+pub struct CompilationOutput {
+    stderr: Vec<u8>,
+}
+
+impl CompilationOutput {
+    // Only used in libbpf-cargo library
+    #[allow(dead_code)]
+    /// Read the stderr from the compilation
+    pub fn stderr(&self) -> &[u8] {
+        &self.stderr
+    }
+}
+
 fn check_progs(objs: &[UnprocessedObj]) -> Result<()> {
     let mut set = HashSet::with_capacity(objs.len());
     for obj in objs {
@@ -171,7 +186,7 @@ fn compile_one(
     out: &Path,
     clang: &Path,
     clang_args: &[OsString],
-) -> Result<()> {
+) -> Result<CompilationOutput> {
     if debug {
         println!("Building {}", source.display());
     }
@@ -234,7 +249,12 @@ fn compile_one(
     // system specific and temporary paths. That can render our generated
     // skeletons unstable, potentially rendering them unsuitable for inclusion
     // in version control systems. So strip this information.
-    strip_dwarf_info(out).with_context(|| format!("Failed to strip object file {}", out.display()))
+    strip_dwarf_info(out)
+        .with_context(|| format!("Failed to strip object file {}", out.display()))?;
+
+    Ok(CompilationOutput {
+        stderr: output.stderr,
+    })
 }
 
 fn compile(
@@ -243,31 +263,31 @@ fn compile(
     clang: &Path,
     mut clang_args: Vec<OsString>,
     target_dir: &Path,
-) -> Result<()> {
+) -> Result<Vec<CompilationOutput>> {
     let header_dir = extract_libbpf_headers_to_disk(target_dir)?;
     if let Some(dir) = header_dir {
         clang_args.push(OsString::from("-I"));
         clang_args.push(dir.into_os_string());
     }
 
-    for obj in objs {
-        let stem = obj.path.file_stem().with_context(|| {
-            format!(
-                "Could not calculate destination name for obj={}",
-                obj.path.display()
-            )
-        })?;
+    objs.iter()
+        .map(|obj| -> Result<_> {
+            let stem = obj.path.file_stem().with_context(|| {
+                format!(
+                    "Could not calculate destination name for obj={}",
+                    obj.path.display()
+                )
+            })?;
 
-        let mut dest_name = stem.to_os_string();
-        dest_name.push(".o");
+            let mut dest_name = stem.to_os_string();
+            dest_name.push(".o");
 
-        let mut dest_path = obj.out.to_path_buf();
-        dest_path.push(&dest_name);
-        fs::create_dir_all(&obj.out)?;
-        compile_one(debug, &obj.path, &dest_path, clang, &clang_args)?;
-    }
-
-    Ok(())
+            let mut dest_path = obj.out.to_path_buf();
+            dest_path.push(&dest_name);
+            fs::create_dir_all(&obj.out)?;
+            compile_one(debug, &obj.path, &dest_path, clang, &clang_args)
+        })
+        .collect::<Result<_, _>>()
 }
 
 fn extract_clang_or_default(clang: Option<&PathBuf>) -> PathBuf {
@@ -316,7 +336,7 @@ pub(crate) fn build_single(
     clang: Option<&PathBuf>,
     skip_clang_version_checks: bool,
     mut clang_args: Vec<OsString>,
-) -> Result<()> {
+) -> Result<CompilationOutput> {
     let clang = extract_clang_or_default(clang);
     check_clang(debug, &clang, skip_clang_version_checks)?;
     let header_parent_dir = tempdir()?;
@@ -331,9 +351,7 @@ pub(crate) fn build_single(
     // BPF. See https://lkml.org/lkml/2020/2/21/1000.
     clang_args.push(OsString::from("-fno-stack-protector"));
 
-    compile_one(debug, source, out, &clang, &clang_args)?;
-
-    Ok(())
+    compile_one(debug, source, out, &clang, &clang_args)
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


libbpf-cargo/src/build.rs
```diff
@@ -243,31 +246,31 @@ fn compile(
     clang: &Path,
     mut clang_args: Vec<OsString>,
     target_dir: &Path,
-) -> Result<()> {
+) -> Result<Vec<Vec<u8>>> {
```
danielocfb, line 249: Shouldn't `build` be adjusted to bubble up this value as well?

danielocfb (commented):
Perhaps we could introduce a `CompileReport` type or something along those lines that contains the input file alongside the compiler output to have at least some structure in the result, instead of having some abstruse doubly nested vector of bytes without any documentation as to what it is about? What do you think?

libbpf-cargo/src/lib.rs
```diff
@@ -64,6 +64,8 @@
 )]
 #![deny(unsafe_op_in_unsafe_fn)]
 
+pub use build::CompilationOutput;
```
JakeHillion, line 67: This struct would make more sense in `lib.rs`, but the build process here compiles each of the modules twice - once in the context of `libbpf_cargo` (lib) and once in the context of `libbpf_cargo` (bin). This means the modules can't access structs declared in `lib.rs`, even though declaring them there would make the interface more explicit.

libbpf-cargo/src/build.rs
JakeHillion: I didn't modify `build::build`. It is only used in the binary crate and it's not obvious how to report these errors in the binary crate - the data is available in `build` for each file if someone decides how it should be displayed.

JakeHillion:
> Perhaps we could introduce a `CompileReport` type or something along those lines that contains the input file alongside the compiler output to have at least some structure in the result, instead of having some abstruse doubly nested vector of bytes without any documentation as to what it is about? What do you think?

I added a `CompilationOutput` type (can rename if that helps) holding the `stderr` and an accessor for the `Cow` string, the interface is much nicer to use now and easier to adjust in the future. I think this is enough, the stderr holds the file name where appropriate, and the functions that return more than one borrow an ordered list of inputs anyway so if someone wants that data it's an easy zip. Providing a slice of inputs and getting a matching Vec of outputs seems very natural to me, but I can change it if needed!

libbpf-cargo/src/test.rs
```diff
@@ -785,7 +785,8 @@ fn test_skeleton_builder_clang_opts() {
         .debug(true)
         .clang("clang")
         .build_and_generate(skel.path())
-        .unwrap_err();
```
JakeHillion, line 788: `unwrap_err` requires `Debug` on the Ok side of the result. I didn't want to implement that just for this test on `CompilationOutput`, `.err().unwrap()` achieves the same thing without needing to print the Ok type.

JakeHillion:
> Looks mostly fine now, thanks! Left two more comments, please take another look. Please also rebase to pull in a few changes I landed to clean things up in `libbpf-cargo` (I would think there shouldn't be any conflicts). Lastly, can you please make sure to get CI into a green state?

Should all be sorted. TIL about always implementing Debug on public types, I'll remember that. Rebase was clean & CI should be sorted now (not sure why GitHub was displaying a red signal below the green ones for me).

libbpf-cargo/src/lib.rs
JakeHillion: @danielocfb ah you've fixed this, awesome! Do you prefer the `pub use` or declaring the struct in lib.rs?

danielocfb: I think keeping it in `build.rs` is fine -- after all its pretty closely related to the "build" part.

danielocfb (approved):

d-e-s-o:
@JakeHillion We recently introduced a more formal logging infrastructure (configurable, with optional support for more structured logging; as opposed to just dumping stuff to stdout). I am considering emitting compilation warnings this way. Given that `sched_ext` really only wants to forward these warnings to the user, I'd think we can come up with a `log` based solution for filtering them out, as opposed to requiring the `CompilationOutput` type.

Do you think this be something that could work for your use case? I can help with the implementation, but the workflow I envision is basically something along the lines of: set up some kind of logger that surrounds the `SkeletonBuilder` [here](https://github.com/JakeHillion/scx/commit/ecced704c2eb8c7903f90fd9b66209b73887a5ac) and which extracts only logs for a certain [target](https://docs.rs/log/latest/log/index.html#examples), then just emit the captured string (if any) via `cargo:warning`. I haven't done something like that, so I am not 100% sure it's feasible with the set of APIs available, but I'd be surprised if it wasn't. In any event, I can look into it, I just wanted to check whether that is something you could get on board with first. My goal is to remove `CompilationOutput`, which kind of feels like a kludge, at least for carrying around compilation warnings. The added advantage would be that we wouldn't be restricted to displaying stuff on the happy-path, but could also dump more information in case of error, for example.

d-e-s-o:
Okay, I found the relevant bits, I think. Basically, it would just be a custom logger as implemented [here](https://docs.rs/log/latest/log/#implementing-a-logger), which checks `Metadata` for the expected target and emits those to stdout with the `cargo:warning=` directive prepended. It can discard everything else. The contract between `libbpf-cargo` and user code would be that we fixate the target for compiler output to something you can depend on.

d-e-s-o:
Hmmm...I just noticed that [`set_logger`](https://docs.rs/log/latest/log/fn.set_logger.html) can only set the logger once. I suppose it would still be fine for a build script, but for other use cases one would now need to build some kind of layered logging solution, which isn't particularly great. Perhaps we'd need to switch to `tracing`, which has [`with_default`](https://docs.rs/tracing/0.1.41/tracing/subscriber/fn.with_default.html)...

d-e-s-o:
I implemented the proposal: https://github.com/d-e-s-o/libbpf-rs/tree/topic/compiler-warning-logs

```
$ cargo build -p compiler_warnings
>    Compiling libbpf-cargo v0.25.0-beta.1 (libbpf-rs/libbpf-cargo)
>    Compiling compiler_warnings v0.0.0 (libbpf-rs/examples/compiler_warnings)
> warning: compiler_warnings@0.0.0: src/bpf/compiler_warnings.bpf.c:14:3: warning: 'thefloorislava' is deprecated: can't-touch-this [-Wdeprecated-declarations]
> warning: compiler_warnings@0.0.0:    14 |   thefloorislava();
> warning: compiler_warnings@0.0.0:       |   ^
> warning: compiler_warnings@0.0.0: src/bpf/compiler_warnings.bpf.c:8:16: note: 'thefloorislava' has been explicitly marked deprecated here
> warning: compiler_warnings@0.0.0:     8 | __attribute__((deprecated("can't-touch-this"))) void thefloorislava() {
> warning: compiler_warnings@0.0.0:       |                ^
> warning: compiler_warnings@0.0.0: 1 warning generated.
> warning: compiler_warnings@0.0.0:
>     Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.61s
```

JakeHillion:
@d-e-s-o Thanks for the heads up! Looking at this the output from the end user perspective is okay, I think it's equivalent to what we have now. I don't like that the `set_logger` call is still used. Correct me if I'm wrong but if we ever wanted to see debug logging from `libbpf-cargo` we now couldn't without code changes? This will also break logging for our application if we've already set it up, and seems way more of a kludge than returning a string.

I suppose the question here is what level of abstraction is `libbpf-cargo` providing. `CompilationOutput` made sense to me as it was a light addition on both sides. My theory was that, if this was implemented in a less kludgy way, `libbpf-cargo` would make emitting the warnings prefixed with `cargo:warning` a first class feature on `SkeletonBuilder`. It seems like this change moves the dodgy code into ours. I'm also quite surprised every other project has been happy to lose their warnings up to this point.

We can adapt to this change either way, but there's my 2 cents.

danielocfb:
> @d-e-s-o Thanks for the heads up! Looking at this the output from the end user perspective is okay, I think it's equivalent to what we have now. I don't like that the `set_logger` call is still used. Correct me if I'm wrong but if we ever wanted to see debug logging from `libbpf-cargo` we now couldn't without code changes? This will also break logging for our application if we've already set it up, and seems way more of a kludge than returning a string.

Please refer to what I mentioned earlier https://github.com/libbpf/libbpf-rs/pull/994#issuecomment-2608348441

> I suppose the question here is what level of abstraction is `libbpf-cargo` providing. `CompilationOutput` made sense to me as it was a light addition on both sides. My theory was that, if this was implemented in a less kludgy way, `libbpf-cargo` would make emitting the warnings prefixed with `cargo:warning` a first class feature on `SkeletonBuilder`. It seems like this change moves the dodgy code into ours. I'm also quite surprised every other project has been happy to lose their warnings up to this point.

There is nothing preventing users from using `SkeletonBuilder` in non-build script contexts. Hence, prefixing everything with `cargo:warning` is a no-go.

I think in general it would be possible for `libbpf-cargo` to provide such a logger implementation ready to be plugged in. But the set-once restriction imposed by `log` makes it a tough sell at this point. Will have to ponder options some more.

danielocfb:
FWIW, I have finalized this now over in https://github.com/libbpf/libbpf-rs/pull/1179

PR #1000, opened 2024-11-07
Fix documentation of MapCore::delete_batch()

danielocfb:
Similar to commit 62130a056fae ("Fix documentation of MapCore::update_batch()"), fix the rendering of the documentation of MapCore::delete_batch() method, whose 'keys' description contained wrong markdown.


danielocfb:
1000!!!111one

PR #1019, opened 2024-11-19
libbpf-cargo: Expose inner Rust "enum" values publicly

danielocfb:
With the switch from emitting C enums as Rust enums to mapping them to Rust structs instead, we no longer allow access to the underlying integer. That's a bit of a usability snag, because it forces users to fall back to transmuting constants just to get a hold of the inner value.
Expose the inner value publicly to make that easier.


javierhonduco:
Thanks!! 🙏

PR #1035, opened 2024-12-11
Fix skeleton generation for enum64 types

danielocfb:

### DIFF
--- a/libbpf-cargo/CHANGELOG.md
+++ b/libbpf-cargo/CHANGELOG.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+- Fixed skeleton generation when `enum64` types are present
+
+
 0.25.0-beta.0
 -------------
 - Represent C enums with custom types and const fields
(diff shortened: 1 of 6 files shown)


danielocfb:
Hm, I wonder if some of the test failures are due to low clang/llvm version...everything is working locally.

anakryiko (approved):
LGTM, as discussed, enum64 test problem is probably due to old Clang

danielocfb:
> LGTM, as discussed, enum64 test problem is probably due to old Clang

Thanks! Will see if we can fix that tomorrow. Worst case we have to ignore the test in CI.

etsal:
I can confirm this fixes the compilation problem. Thank you!

danielocfb:
> I can confirm this fixes the compilation problem. Thank you!

Let me know in case you need a new release with this fix.

etsal:
> > I can confirm this fixes the compilation problem. Thank you!
> 
> Let me know in case you need a new release with this fix.

Sorry for the late response. That would be great, thank you.

danielocfb:
Okay, [`v0.25.0-beta.1`](https://crates.io/crates/libbpf-rs/0.25.0-beta.1) is out now.

PR #1040, opened 2024-12-17
Make ARM builds optional in CI

danielocfb:
Ubuntu can't get their act together and provide reliably accessible servers. Make the repeatedly failing jobs optional for the time being.


danielocfb:
No idea if that works properly, but at least on the CI summary page stuff shows up as successful now...

PR #1060, opened 2025-01-16
Downgrade Ubuntu version for Arm testing

danielocfb:
Our Arm jobs have been failing for a while. At first that was sporadic (in retrospect likely because we got runners sporting different images due to staged rollout of a new Ubuntu version) and now its permanent. The cause of the failure seems to be some change in format of their sources.list, causing us to no longer be able to no longer fetch necessary package lists. I haven't found a guide in their documentation how to properly update, so roll back to a known good distro version until somebody has the nerves to deal with this %#&*($&#*($.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,7 +153,7 @@ jobs:
 
   build-aarch64:
     name: Build for aarch64
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


danielocfb:
And actually https://askubuntu.com/questions/1514281/ubuntu-24-04-apt-specify-architecture-using-sources-files is likely of help once we want to upgrade.

PR #1062, opened 2025-01-16
Revert "Make ARM builds optional in CI"

danielocfb:
We just fixed out Arm tests with commit 2ce495438856 ("Downgrade Ubuntu version for Arm testing"). Hence, there is no need for said tests to be optional anymore. Revert the commit making them so.

This reverts commit 271afc86ce55bb51369437b6a00219329a6d70f2.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -154,9 +154,6 @@ jobs:
   build-aarch64:
     name: Build for aarch64
     runs-on: ubuntu-22.04
-    # Make this job optional until Ubuntu gets their act together and
-    # provides usable infrastructure.
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v4
       - name: Add apt sources for arm64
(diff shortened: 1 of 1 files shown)


danielocfb:
> Thanks for re-enabling them, I find myself running more libbpf-rs applications on arm64. Just a drive-by comment, but now that [GitHub added support for native arm64](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) runners perhaps we should use them?
> 
> Just building in this arch won't offer a big benefit I think but if we were to start running tests on arm64 this could be quite of a speed-up compared to running in user space emulation (I don't think we run them yet but I might be wrong).

Oh interesting, I missed this announcement. Last time I [checked](https://github.blog/changelog/2024-09-03-github-actions-arm64-linux-and-windows-runners-are-now-generally-available/), nothing was available to non-paying customers yet. So yes, we can look into switching over.

PR #1063, opened 2025-01-17
Add support for compiling BPF object file from multiple input files

danielocfb:
Please see individual commits for descriptions. This is internal logic only at this point.

### DIFF
--- a/libbpf-cargo/src/build.rs
+++ b/libbpf-cargo/src/build.rs
@@ -267,7 +267,6 @@ fn compile(
     objs: &[UnprocessedObj],
     clang: &Path,
     clang_args: Vec<OsString>,
-    _target_dir: &Path,
 ) -> Result<Vec<CompilationOutput>> {
     objs.iter()
         .map(|obj| -> Result<_> {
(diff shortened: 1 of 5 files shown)


anakryiko (approved):
LGTM, I think there is just one potential issue with the API: compiler args. It seems like currently you only allow to specify compiler args for all .bpf.c files  and there is no way to augment it with some extra stuff for individual files.

This seems like something that can be added in the future, treating current compiler_args as "common compiler args", so doesn't seem blocking. But I wanted to point it out anyways.

d-e-s-o:
Thanks for the review. Fair point, though can you think of any concrete examples where this is a blocker? I mean obviously we can construct something, but I am more asking about partical experience where this would be an unsolvable limitation.

We could rework the API to add individual files (similar to what we do with the [`Linker`](https://docs.rs/libbpf-rs/0.24.8/libbpf_rs/struct.Linker.html#method.add_file)), it just really seems unwieldy and over the top.

anakryiko:
Typically (and by "typically" it's still pretty rare, but still...) it would be the need to do some file-specific `-DSOME_DEFINE` or just muting some known errors with `-Wno-....`, where you wouldn't want to do it across all files. For BPF selftests I see that only two out of many .bpf.c files have extra `-fno-inline` argument (don't ask why, I don't know). So just something that I've seen in practice.

anakryiko:
from API perspective, having common compiler args like you do makes sense and is convenient for 99% of cases. Perhaps the solution here is to have one extra API when adding individual file to also specify compiler args (or just change existing API to access Option[] argument for extra args, which would require all those None`s in common case, so I don't know).

d-e-s-o:
Okay, yeah, I will keep that in mind for when/if we expose this functionality publicly. Until then it is certainly a non-issue.

PR #1064, opened 2025-01-18
map: Add max_entries() for Map objects

Jongy:
My use case is:
1. Read map's max_entries before loading, and resize max_entries according to some logic.
2. After map is loaded, access resized max_entries for logging etc.

For the sake of completeness I added `max_entries` on `MapCore` and implemented on `MapHandle` too, although I don't use this flow (I can add just on `Map` instead, or instead implement it with `bpf_obj_get_info_by_fd` to avoid increasing the struct's size)

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -103,6 +103,11 @@ impl<'obj> OpenMap<'obj> {
             Some(data)
         }
     }
+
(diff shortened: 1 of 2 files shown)


danielocfb (approved):
Seems fine to me, thanks.

Jongy:
Thanks :)

PR #1073, opened 2025-01-21
Use `#[expect(...)]` instead of `#[allow(...)]` in some places

danielocfb:
Use the `#[expect(...)]` attribute instead of `#[allow(...)]` in a bunch of places where we only have the attribute to silence a lint. The former makes sure that the silenced issue is actually flagged, or prints a warning otherwise.


danielocfb:
This requires Rust `1.81` and, hence, will have to wait until `1.86` is released, as per our MSRV policy.

PR #1074, opened 2025-01-22
libbpf-cargo: Handle duplicated types in BTF more gracefully

danielocfb:
Our `SkeletonBuilder` currently only support building a skeleton from a single `.bpf.c` file. However, users may very well compile multiple `.bpf.c` files into `.bpf.o` and link them into a single object. In such a scenario, it is easily possible that we end up with multiple types of the same name in the corresponding BTF. Currently, this results in compilation errors.
With this change we handle this case more gracefully, by enumerating these types, similar to what libbpf does in its dumping logic.

### DIFF
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -1096,6 +1096,98 @@ fn test_skeleton_builder_deterministic() {
     assert_eq!(skel1, skel2);
 }
 
+/// Check that we generate a valid skeleton in the presence of duplicate
+/// structs in the BTF.
+#[test]
+fn test_skeleton_duplicate_struct() {
+    let (_dir, proj_dir, cargo_toml) = setup_temp_project();
+
+    create_dir(proj_dir.join("src/bpf")).expect("failed to create prog dir");
+    let prog1 = proj_dir.join("src/bpf/prog1.bpf.c");
+    write(
+        &prog1,
+        r#"
+            #include "vmlinux.h"
+
+            struct foobar {
+                int bar;
+            } foo;
+        "#,
+    )
+    .expect("failed to write prog.bpf.c");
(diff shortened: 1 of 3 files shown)


anakryiko (approved):
LGTM

libbpf-cargo/src/test.rs
```diff
@@ -1096,6 +1096,98 @@ fn test_skeleton_builder_deterministic() {
...
+    write!(
+        source,
+        r#"
+        #[path = "{skel_path}"]
+        mod skel;
+        use skel::*;
+
+        fn main() {{
+            let _foobar0 = types::foobar::default();
+            let _foobar1 = types::foobar_1::default();
```
anakryiko, line 1173: actually, wouldn't `foobar_2` make more sense (it's a second instance of the same name, we just don't explicitly mark the first one with `_1`)?

d-e-s-o: I am living happily in my counting-starts-at-zero bubble :-)

But happy to change it.

Btw., strictly speaking I think the renaming algorithm is insufficient: we can't just assume that `<type>_<cnt>` is an available name, but should actually check that there is no collision with an existing type of that name.

anakryiko: heh, I see :) I'd switch to two, but ultimately it doesn't matter all that much. If user gets to that point, there is no deterministic order, so whether it's foobar_1 or foobar_2, I wouldn't use any of those in the code directly because their ordering can change on the next compilation

and yeah, you are right about possibility of causing another unintended collision (which, on libbpf side, is mostly mitigated by using triple underscore, which makes it highly unlikely)

d-e-s-o:
@etsal this should fix your compilation error.

PR #1094, opened 2025-02-24
examples/bpf_query: Fix disassemble output

chenhengqi:
The BPF prog disassemble output seems incorrect compared to bpftool. 
Change the bitness to 64 fix this.


d-e-s-o (approved):
Seems sensible to me. Can you please rebase to top of tree and we can get this in. Thanks!

PR #1095, opened 2025-02-24
Fix CI

kxxt:
- Make clippy happy
- Fix JSON escaping in `required-checks`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -261,5 +261,7 @@ jobs:
     ]
     if: always()
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
chenhengqi, line 1: The change here seems unnecessary.

kxxt: It is necessary. It is valid for JSON to contain single quotes and that can break it. It's just that such pitfalls haven't be triggered in this repo for now.

d-e-s-o (approved):
Seems fine to me, thanks!

PR #1096, opened 2025-02-25
libbpf-rs: Make func_name optional for uprobe attach

chenhengqi:
Change UprobeOpts::func_name to Option<String> 
so that user can attach uprobe to specifc offset in binary. 
This would be useful for target that have symbol stripped.

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -4,6 +4,7 @@
 
 use std::ffi::c_void;
 use std::ffi::CStr;
+use std::ffi::CString;
 use std::ffi::OsStr;
 use std::marker::PhantomData;
 use std::mem;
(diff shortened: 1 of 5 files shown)


libbpf-rs/src/program.rs
```diff
@@ -52,7 +53,7 @@ pub struct UprobeOpts {
     /// `func_name` and use `func_offset` argument to specify offset within the
     /// function. Shared library functions must specify the shared library
     /// binary_path.
-    pub func_name: String,
+    pub func_name: Option<String>,
```
d-e-s-o, line 56: It seems the documentation would need to be updated accordingly. Also, `func_offset` appears to be a bit of a misnomer now. Should probably be renamed to `offset`.

d-e-s-o (commented):
Seems fine for the most part, thanks! Would you mind adding a test for this functionality?

libbpf-rs/src/program.rs
chenhengqi: Just added a test. As for the naming and the comments, it seems all derive from libbpf.

danielocfb (approved):
Thanks for the test!

PR #1100, opened 2025-02-26
ringbuf: Fix callback docs

danobi:
It's not non-zero - it's negative error values. The error is also propagated (which is important if translated to Result), so document that as well.


d-e-s-o (approved):
Looks good. For reference, this seems to be the code in question, I think: https://github.com/libbpf/libbpf/blob/42a6ef63161a8dc4288172b27f3870e50b3606f7/src/ringbuf.c#L261-L267

danobi:
Yep, that's what I was referencing. Was in a hurry so didn't link it :P

Issue #1085, opened 2025-02-05
Add `set_autoattach` to the OpenProgramMut


mmullin-halcyon:
There are cases where it is necessary to load but not attach a bpf program.  libbpf-rs exposes the ability to disable autoloading, but it does not expose the ability to disable autoattach

https://github.com/libbpf/libbpf-rs/blob/master/libbpf-rs/src/program.rs#L230C1-L235C6

Similar functionality can be added

```
    /// Set whether a bpf program should be automatically loaded by default
    /// when the bpf object is loaded.
    pub fn set_autoattach(&mut self, autoload: bool) {
        let rc = unsafe { libbpf_sys::bpf_program__set_autoattach(self.ptr.as_ptr(), autoload) };
        debug_assert!(util::parse_ret(rc).is_ok(), "{rc}");
    }
```

danielocfb:
Seems fine to me to add; please feel free to open a pull request.

danielocfb:
Closed via https://github.com/libbpf/libbpf-rs/commit/3ead30dc7736b9ad53bcac4ec5d5b0952f135f1b

PR #1106, opened 2025-03-01
add: binding for `set_autoattach`

syogaraj:
Add binding for `bpf_program__set_autoattach` as discussed in #1085 . 
`bpf_program__set_autoattach` is of type `void`. So, no return checks are needed. 

Thanks to @mmullin-halcyon


chenhengqi (commented):
LGTM but the commit message could be improved though.

danielocfb (approved):
I agree with @chenhengqi, but probably not worth doing another round trip. Please stay consistent with existing commit message styles in the future. Ideally we'd also have a test for this functionality, but given that we don't have one for autoload not going to push for that either. Thanks!

PR #1109, opened 2025-03-04
libbpf-rs: Unify calls to libbpf APIs

chenhengqi:
Previously, we use if/else to call the same libbpf API
with different params (null or non-null). This is a bit
cumbersome. Let's make it more idiomatic and succinct.
No functional changes intended.


danielocfb (approved):
Seems fine to me, thanks!

Issue #1110, opened 2025-03-04
Incorrect offset calculation for uprobe function name


Dredegui:
# Uprobe function name to offset translation not working 

## Description

I am facing an issue with `libbpf-rs` when trying to use its uprobes with opts and having it determine the offset of the function name. The offset doesn't seem correct, and it is causing issues with attaching the uprobe to the correct location.

## My environment and what I executed

1. Lxc containers deployed according to `Jepsen` specifications.
2. Running 5 containers of the latest version of `Redpanda`.
3. Attaching uprobe to cluster::topics_frontend::create_topics() function.
4. Compared it with the same parameters on bpftrace command.

## Results (3 Terminal View)

Both the tracer using libbpf-rs uprobes and bpftrace were executed before the commands on view 2

### Execution of uprobes with libbpf-rs (View 1)
```bash
./run-rosetracer.sh  
...
Trying to attach uprobe for index 0, function _ZN7cluster15topics_frontend13create_topicsE17fragmented_vectorINS_37custom_assignable_topic_configurationELm18446744073709551615EE NSt3_16chrono10time_point IN7seastar12lowres_clockENS5_8durationIxNS4_5ratioILL1ELT1000000000EEEEEEE and pid 60924  
Uprobe result: Ok (Link { ptr: 0x5e3662ddfed0 }) Injected in pos in pid 60924  
Received: n5, 47906, 60950, 0, veth89e4c06a  
Started tracing for pid 60950 with node_name n5  
Binary location: /var/snap/lxd/common/lxd/storage-pools/default/containers/n5/rootfs/opt/redpanda/libexec/redpanda  
Trying to attach uprobe for index 0, function _ZN7cluster15topics_frontend13create_topicsE17fragmented_vectorINS_37custom_assignable_topic_configurationELm18446744073709551615EE NSt3_16chrono10time_point IN7seastar12lowres_clockENS5_8durationIxNS4_5ratioILL1EL11000000000EEEEEEE and pid 60950  
Uprobe result: Ok(Link { ptr: 0x5e3662de0590 }) Injected in pos e in pid 60950  
```
---
### Activation of probed function and print of created uprobes (View 2)
```bash    
~/work/blotds-thesis/env/redpanda-setup$ lxc exec n1 -- rpk topic create test
STATUS  
TOPIC test10 OK  

~/work/blotds-thesis/env/redpanda-setup$ sudo bpftool perf show
[...]

PR #1112, opened 2025-03-06
libbpf-rs: Fix get_symbol_offset() in test code

chenhengqi:
As report in #1110, it seems like the current implementation
of symbol resolve is incorrect. Fix it in BCC way.

### DIFF
--- a/libbpf-rs/tests/common/mod.rs
+++ b/libbpf-rs/tests/common/mod.rs
@@ -3,6 +3,8 @@ use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
 
+use goblin::elf::program_header::PF_X;
+use goblin::elf::program_header::PT_LOAD;
 use goblin::elf::sym;
 use goblin::elf::Elf;
 use libbpf_rs::Map;
@@ -102,9 +104,9 @@ pub fn get_symbol_offset(binary_path: &Path, symbol_name: &str) -> Result<usize,
         }
         if let Some(name) = elf.dynstrtab.get_at(sym.st_name) {
             if name == symbol_name {
-                let sec = &elf.section_headers[sym.st_shndx];
-                let offset = sym.st_value - sec.sh_addr + sec.sh_offset;
-                return Ok(offset as usize);
+                if let Some(offset) = check_symbol_offset(&elf, &sym) {
+                    return Ok(offset);
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


libbpf-rs/tests/common/mod.rs
```diff
@@ -102,9 +103,18 @@ pub fn get_symbol_offset(binary_path: &Path, symbol_name: &str) -> Result<usize,
...
-                let offset = sym.st_value - sec.sh_addr + sec.sh_offset;
-                return Ok(offset as usize);
+                for phdr in elf.program_headers.iter() {
+                    if phdr.p_type != PT_LOAD {
+                        continue;
+                    }
+                    if phdr.p_flags & PF_X == 0 {
+                        continue;
+                    }
+                    if sym.st_value >= phdr.p_vaddr && sym.st_value < phdr.p_vaddr + phdr.p_memsz {
```
danielocfb, line 113:
```suggestion
                    if (phdr.p_vaddr..phdr.p_vaddr + phdr.p_memsz).contains(&sym.st_value) {
```

libbpf-rs/tests/common/mod.rs
```diff
@@ -116,9 +126,18 @@ pub fn get_symbol_offset(binary_path: &Path, symbol_name: &str) -> Result<usize,
         }
         if let Some(name) = elf.strtab.get_at(sym.st_name) {
             if name == symbol_name {
-                let sec = &elf.section_headers[sym.st_shndx];
-                let offset = sym.st_value - sec.sh_addr + sec.sh_offset;
-                return Ok(offset as usize);
+                for phdr in elf.program_headers.iter() {
```
danielocfb, line 129: Can we find a way to eliminate duplication with the above? It seems we should be able to factor out a function?

libbpf-rs/tests/common/mod.rs
```diff
@@ -116,9 +126,18 @@ pub fn get_symbol_offset(binary_path: &Path, symbol_name: &str) -> Result<usize,
...
+                    if phdr.p_type != PT_LOAD {
+                        continue;
+                    }
+                    if phdr.p_flags & PF_X == 0 {
+                        continue;
+                    }
```
danielocfb, line 135: Does this check add any value? Wouldn't things pan out even if the section were not executable? Probably doesn't matter much in this bit of test code, but it seems superfluous nevertheless.

danielocfb (commented):
Seems fine to me. Can you please revert all spurious formatting changes, though? Also left a few comments, please take a look and let me know what you think.

libbpf-rs/tests/common/mod.rs
chenhengqi: I think it is reasonable to be executable. Otherwise the uprobe attach may also fail I guess. Need more investigation.

danielocfb:
Anyway, seems fine either way, given that this is localized test code.

PR #1115, opened 2025-03-06
Add a few program attach variants

atenart:
The following variants are added to attach a program:
- Kprobe multi, with and without providing additional options.
- Kprobe with additional options.
- Raw tracepoint with addition options.

The main motivation is 1) kprobe multi has significant performance benefits and 2) using additional options allows to use BPF cookies in probes.

In addition tests for raw tracepoints are added (for the existing support and for the new variant added), as well as basic tests for all kprobe variants (excluding kretprobes).

A few things to note wrt. to reviews:
- `attach_kprobe_with_opts` kept the `kretprobe` flag as an argument even though it is part of the underlying `libbpf-sys` options struct. It felt more natural IMO to keep this flag visible and in sync with the normal kprobe API, but this can be changed easily.
- The kprobe tests are quite minimal. The difficulty of having kprobe tests is the stability of the probed functions, they could disappear without notice (kernel version, inlined, etc). The approach here was to use the exposed testing kfuncs, which should be a bit more stable and not inlined. There will always be a risk though. Also using this approach prevents from actually doing something in the probe as we can't trigger it. The patch can be removed if not wanted.

### DIFF
--- a/libbpf-rs/CHANGELOG.md
+++ b/libbpf-rs/CHANGELOG.md
@@ -1,5 +1,10 @@
 Unreleased
 ----------
+- Added kprobe multi support for attaching programs, with and without providing
+  additional options
+- Allow to provide additional options when attaching programs to raw
+  tracepoints
+- Allow to provide additional options when attaching programs to kprobes
 - Added `max_entries` getter to various map types
 - Added `OpenProgramMut::set_autoattach`
 - Adjusted `UprobeOpts::func_name` to be an `Option`
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -139,6 +139,32 @@ impl From<RawTracepointOpts> for libbpf_sys::bpf_raw_tracepoint_opts {
     }
 }
 
+/// Options to optionally be provided when attaching to a kprobe.
+#[derive(Clone, Debug, Default)]
+pub struct KprobeOpts {
+    /// Custom user-provided value accessible through `bpf_get_attach_cookie`.
+    pub cookie: u64,
+    #[doc(hidden)]
+    pub _non_exhaustive: (),
+}
+
+impl From<KprobeOpts> for libbpf_sys::bpf_kprobe_opts {
+    fn from(opts: KprobeOpts) -> Self {
+        let KprobeOpts {
+            cookie,
+            _non_exhaustive,
+        } = opts;
+
+        #[allow(clippy::needless_update)]
+        libbpf_sys::bpf_kprobe_opts {
+            sz: size_of::<Self>() as _,
+            bpf_cookie: cookie,
+            // bpf_kprobe_opts might have padding fields on some platform
+            ..Default::default()
+        }
+    }
+}
+
 /// Options to optionally be provided when attaching to multiple kprobes.
 #[derive(Clone, Debug, Default)]
 pub struct KprobeMultiOpts {
@@ -897,6 +923,34 @@ impl<'obj> ProgramMut<'obj> {
         let func_name_ptr = func_name.as_ptr();
         let ptr = unsafe {
             libbpf_sys::bpf_program__attach_kprobe(self.ptr.as_ptr(), retprobe, func_name_ptr)
+        };
+        let ptr = validate_bpf_ret(ptr).context("failed to attach kprobe")?;
+        // SAFETY: the pointer came from libbpf and has been checked for errors.
+        let link = unsafe { Link::new(ptr) };
+        Ok(link)
+    }
+
+    /// Attach this program to a [kernel
+    /// probe](https://www.kernel.org/doc/html/latest/trace/kprobetrace.html),
+    /// providing additional options.
+    pub fn attach_kprobe_with_opts<T: AsRef<str>>(
+        &self,
+        retprobe: bool,
+        func_name: T,
+        opts: KprobeOpts,
+    ) -> Result<Link> {
+        let func_name = util::str_to_cstring(func_name.as_ref())?;
+        let func_name_ptr = func_name.as_ptr();
+
+        let mut opts = libbpf_sys::bpf_kprobe_opts::from(opts);
+        opts.retprobe = retprobe;
+
+        let ptr = unsafe {
+            libbpf_sys::bpf_program__attach_kprobe_opts(
+                self.ptr.as_ptr(),
+                func_name_ptr,
+                &opts as *const _,
+            )
         };
         let ptr = validate_bpf_ret(ptr).context("failed to attach kprobe")?;
         // SAFETY: the pointer came from libbpf and has been checked for errors.
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -113,6 +113,32 @@ impl From<TracepointOpts> for libbpf_sys::bpf_tracepoint_opts {
     }
 }
 
+/// Options to optionally be provided when attaching to a raw tracepoint.
+#[derive(Clone, Debug, Default)]
+pub struct RawTracepointOpts {
+    /// Custom user-provided value accessible through `bpf_get_attach_cookie`.
+    pub cookie: u64,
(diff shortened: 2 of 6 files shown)


libbpf-rs/src/program.rs
```diff
@@ -113,6 +113,71 @@ impl From<TracepointOpts> for libbpf_sys::bpf_tracepoint_opts {
...
+        libbpf_sys::bpf_raw_tracepoint_opts {
+            sz: size_of::<Self>() as _,
+            cookie,
+            // bpf_raw_tracepoint_opts might have padding fields on some platform
+            ..Default::default()
+        }
+    }
+}
+
+/// Options to optionally be provided when attaching to a kprobes.
```
danielocfb, line 142:
```suggestion
/// Options to optionally be provided when attaching to a kprobe.
```

libbpf-rs/src/program.rs
```diff
@@ -864,6 +878,96 @@ impl<'obj> ProgramMut<'obj> {
         Ok(link)
     }
 
+    fn check_kprobe_multi_args<T: AsRef<str>>(symbols: &[T], cookies: &[u64]) -> Result<usize> {
+        if symbols.is_empty() {
+            return Err(Error::with_invalid_data("Symbols list cannot be empty"));
```
danielocfb, line 963: Let's introduce and use `Error::with_invalid_input` here and below. These are effectively user input issues.

atenart:
> In addition tests for raw tracepoints are added (for the existing support and for the new variant added), as well as basic tests for all kprobe variants (excluding kretprobes).

Looks like some of the added tests are failing while being triggered in Github actions. They worked locally and there might be some incompatibilities. I'll look at this later, I don't think that impacts the review of the rest as no existing test is failing AFAICT.

libbpf-rs/src/program.rs
```diff
@@ -864,6 +878,96 @@ impl<'obj> ProgramMut<'obj> {
...
+            mut cookies,
+            retprobe,
+            _non_exhaustive,
+        } = opts;
+
+        let cnt = Self::check_kprobe_multi_args(&symbols, &cookies)?;
+
+        let mut syms = symbols
+            .iter()
+            .map(|s| util::str_to_cstring(s.as_ref()).map(|s| s.into_raw()))
```
danielocfb, line 1031: Aren't we leaking the symbol name memory here?

danielocfb (commented):
Overall this seems great, thanks! Left a few mostly minor comments. As for the test failures, let's ignore them for the time being, as it seems GitHub Actions kernel is just not there yet or is compiled without support.
```diff
--- libbpf-rs/tests/test.rs
+++ libbpf-rs/tests/test.rs
@@ -1653,6 +1653,7 @@ fn test_object_kprobe_with_opts() {
 /// kprobe_multi.
 #[tag(root)]
 #[test]
+#[ignore = "requires kernel with kprobe support"]
 fn test_object_kprobe_multi() {
     let mut open_obj = open_test_object("kprobe.bpf.o");
     open_obj
```
Or so. I checked and they pass on my local system.

atenart:
> Overall this seems great, thanks! Left a few mostly minor comments. As for the test failures, let's ignore them for the time being, as it seems GitHub Actions kernel is just not there yet or is compiled without support.
> 
> ```diff
> --- libbpf-rs/tests/test.rs
> +++ libbpf-rs/tests/test.rs
> @@ -1653,6 +1653,7 @@ fn test_object_kprobe_with_opts() {
>  /// kprobe_multi.
>  #[tag(root)]
>  #[test]
> +#[ignore = "requires kernel with kprobe support"]
>  fn test_object_kprobe_multi() {
>      let mut open_obj = open_test_object("kprobe.bpf.o");
>      open_obj
> ```
> 
> Or so. I checked and they pass on my local system.

Thanks for the prompt review!

Yes, looks like the kernel on Github actions is not supporting kprobe multi and access to getting the BPF cookie from raw tp (that last part isn't surprising as it was only added a year ago, I just checked). As you suggested, I added an `ignore` attribute for those tests.

libbpf-rs/src/program.rs
```diff
@@ -864,6 +930,136 @@ impl<'obj> ProgramMut<'obj> {
...
+            retprobe,
+            // bpf_kprobe_multi_opts might have padding fields on some platform
+            ..Default::default()
+        };
+
+        let link = self.attach_kprobe_multi_impl(opts)?;
+
+        // Calling `CString::into_raw` requires the CString to be reconstituted
+        // using `CString::from_raw` to be properly deallocated.
+        let _ = syms.drain(..).map(|s| unsafe { CString::from_raw(s) });
```
chenhengqi, line 1058: I am confused about this line.

libbpf-rs/tests/bin/src/kprobe.bpf.c
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0
+
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+
+SEC("kprobe/bpf_fentry_test1")
+int handle__kprobe(void *ctx)
+{
+    return 0;
+}
```
chenhengqi, line 10: Use tab for indentation. I am preparing a patch to format bpf code.

libbpf-rs/src/program.rs
atenart: `CString::into_raw` transfers ownership and to avoid memory leak the above is required. This is explained in the doc, https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw.

libbpf-rs/tests/bin/src/kprobe.bpf.c
atenart: Oops, thanks for noticing.

libbpf-rs/src/program.rs
danielocfb: It makes sense, but it's kind of error prone. The next person may adjust the `attach` call above to return eagerly and we end up leaking again (even more easy to do because the `link` is actually a `Result<Link>` and not `Link` as seems to be implied). Personally I'd prefer we keep the `Vec<CString>` around and then another `Vec` comprised of a mapping using the non-consuming https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_ptr

I don't feel strongly, though.

danielocfb: Actually, we may leak even now: if part of
```rust
let mut syms = symbols
            .iter()
            .map(|s| util::str_to_cstring(s.as_ref()).map(|s| s.into_raw()))
```
fails midway through then I think we loose the first part that has already been converted to raw pointers, right? So we should probably fix that up.

libbpf-rs/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Unreleased
 - Added `OpenProgramMut::set_autoattach`
 - Adjusted `UprobeOpts::func_name` to be an `Option`
 - Implemented `Sync` for `Link`
+- Added kprobe multi support for attaching programs, with and without providing
+  additional options.
+- Allow to provide additional options when attaching programs to raw
+  tracepoints.
+- Allow to provide additional options when attaching programs to kprobes.
```
danielocfb, line 11: Nit: Let's order these to the top please. I try to order items by the some subjective "impact"/size of change and these are most certainly bigger additions than the few member additions above. Please also remove trailing periods `.` to stick with existing style.

danielocfb (commented):
Forgot to leave one more minor comment. Please take a look at the potential leak mentioned above. Then I think we can get this merged.

libbpf-rs/src/program.rs
atenart: Those are very good points; you're right choosing `into_raw` instead of `as_ptr` to avoid keeping a `Vec<CString>` around led to leak memory on error. Changing this as you suggested, keeping an extra `Vec` around.

danielocfb (approved):
Looks great, thanks!

PR #1116, opened 2025-03-07
libbpf-rs: Update docs in query.rs

syogaraj:


libbpf-rs/src/query.rs
```diff
@@ -122,36 +122,60 @@ impl From<&libbpf_sys::bpf_line_info> for LineInfo {
...
+    /// contents; used to detect changes in the program code.
     pub tag: Tag,
+    /// A unique identifier for the program instance.
     pub id: u32,
+    /// Pointer to the JIT-compiled instructions
     pub jited_prog_insns: Vec<u8>,
+    /// Pointer to the translated BPF instructions in an intermediate
     pub xlated_prog_insns: Vec<u8>,
-    /// Duration since system boot
+    /// Duration since system boot when the program was loaded.
```
chenhengqi, line 141: This should NOT be interpreted as duration. It should be an Instant instead.
This comment is completely misleading.

libbpf-rs/src/query.rs
```diff
@@ -122,36 +122,60 @@ impl From<&libbpf_sys::bpf_line_info> for LineInfo {
 #[repr(C)]
 pub struct Tag(pub [u8; 8]);
 
-/// Information about a BPF program
+/// Information about a BPF program. Translates to `struct bpf_prog_info` in kernel uapi.
```
chenhengqi, line 125: Translates to ?

libbpf-rs/src/query.rs
syogaraj: "Time (since system boot) at which the program was loaded". Is this correct? 
Also, should we change the `Duration::from_nanos(item.load_time)` logic to calculate the correct time? I think we should keep the value as it is.

libbpf-rs/src/query.rs
syogaraj: Referred tp other comments. Will update it to "Maps to".

libbpf-rs/src/query.rs
```diff
@@ -586,33 +624,37 @@ impl Iterator for BtfInfoIter {
...
 
+/// Information about a raw tracepoint.
 #[derive(Debug, Clone)]
-// TODO: Document members.
-#[allow(missing_docs)]
 pub struct RawTracepointLinkInfo {
+    /// The name of the raw tracepoint
     pub name: String,
+    /// Length (in bytes) of the raw tracepoint name
+    pub name_len: u32,
```
danielocfb, line 633: Why is this member needed? We do have the name and it has a length, no? This seems like useless duplication of information to me.

libbpf-rs/src/query.rs
```diff
@@ -436,24 +460,38 @@ impl Iterator for ProgInfoIter {
...
     pub id: u32,
+    /// Size (in bytes) of the keys stored in the map.
     pub key_size: u32,
+    /// Size (in bytes) of the values stored in the map.
     pub value_size: u32,
+    /// Maximum number of entries that the map can hold.
     pub max_entries: u32,
+    /// Map flags indicating specific properties. (e.g., `BPF_F_NO_PREALLOC`)
     pub map_flags: u32,
+    ///Network interface index if the map is associated with a specific device. Otherwise, this
```
danielocfb, line 480:
```suggestion
    /// Network interface index if the map is associated with a specific device. Otherwise, this
```

libbpf-rs/src/query.rs
danielocfb: All "documentation" I found says: "ns since boottime", which strikes me as very similar to what we had. Personally, I'd leave it the way it was, including keeping it as `Duration`. If it's misleading then it is misleading in the kernel as well, no?

libbpf-rs/src/query.rs
```diff
@@ -122,36 +122,60 @@ impl From<&libbpf_sys::bpf_line_info> for LineInfo {
...
+    /// A user-defined name for the BPF program (null-terminated string).
     pub name: CString,
+    /// The type of the program.
     pub ty: ProgramType,
+    /// An 8-byte hash (`BPF_TAG_SIZE`) computed from the program’s
+    /// contents; used to detect changes in the program code.
     pub tag: Tag,
+    /// A unique identifier for the program instance.
     pub id: u32,
+    /// Pointer to the JIT-compiled instructions
```
danielocfb, line 137: It's not really a pointer as it's the actual data, right? Similar below.

libbpf-rs/src/query.rs
```diff
@@ -122,36 +122,60 @@ impl From<&libbpf_sys::bpf_line_info> for LineInfo {
...
+    /// Pointer to line info records for the JIT-compiled code.
     pub jited_line_info: Vec<*const c_void>,
+    /// Size (in bytes) of each line info record.
     pub line_info_rec_size: u32,
+    /// Size (in bytes) of each record in the JITed line info array.
     pub jited_line_info_rec_size: u32,
+    /// Pointer to an array of program tags.
     pub prog_tags: Vec<Tag>,
+    /// Total accumulated run time (in nanoseconds) for the program’s execution.
     pub run_time_ns: u64,
```
danielocfb, line 177: Well, looking at it this should probably be a `Duration`. But of course you didn't change the type so it's fine to keep for now. Just pointing it out in case you want to fix it up subsequently.

danielocfb (commented):
Definitely a good improvement; thanks again! Left a few comments, please take another look.

libbpf-rs/src/query.rs
chenhengqi: You can use bpftool as a reference.

libbpf-rs/src/query.rs
syogaraj: Yes, didn't see the implementation. I referred `struct bpf_prog_info` and wrote the doc. I'll change it.

danielocfb (approved):
Looks good to me, thanks!

PR #1121, opened 2025-03-11
Make attach_kprobe_multi consistent with attach_kprobe

atenart:
The attach_kprobe helper has a kretprobe boolean as a first argument and then the target name. The attach_kprobe_multi helper has similar arguments, but in the inverse order. Reorder them to make the API consistent.

This shouldn't break production software as the kprobe multi support isn't part of an official release yet.


danielocfb (approved):
Looks good to me, thanks!

PR #1127, opened 2025-03-12
libbpf-rs: add missing link types

syogaraj:
We have moved to the libbpf-sys v1.5. Added sockmap link type now. 
Perf-event link type was missed in the previous pr.


danielocfb (approved):
The change looks good to me. Thanks again.

PR #1153, opened 2025-04-10
libbpf-rs: minor fixes suggested by linter

theihor:
clippy compalined here: https://github.com/libbpf/libbpf-rs/actions/runs/14391135733/job/40358223316

### DIFF
--- a/libbpf-rs/src/btf/mod.rs
+++ b/libbpf-rs/src/btf/mod.rs
@@ -630,7 +630,7 @@ impl<'btf> BtfType<'btf> {
                 let int = types::Int::try_from(skipped).unwrap();
                 Ok(Ord::min(
                     ptr_size,
-                    NonZeroUsize::new(((int.bits + 7) / 8).into()).unwrap(),
+                    NonZeroUsize::new(int.bits.div_ceil(8).into()).unwrap(),
                 ))
             }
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/btf/mod.rs
```diff
@@ -630,7 +630,7 @@ impl<'btf> BtfType<'btf> {
                 let int = types::Int::try_from(skipped).unwrap();
                 Ok(Ord::min(
                     ptr_size,
-                    NonZeroUsize::new(((int.bits + 7) / 8).into()).unwrap(),
+                    NonZeroUsize::new(usize::from(int.bits.div_ceil(8))).unwrap(),
```
anakryiko, line 633: `int.bits.div_ceil(8).into()` ?

theihor: It's explicit vs implicit cast, right? With into() the type is figured out by compiler?

Anyways...

anakryiko: yes, and I've seen Daniel using .into() everywhere, which is why I'm pointing this out, this seems to be more idiomatic

anakryiko:
thanks for quick fixes! back to green!

PR #1173, opened 2025-05-21
RingBuffer: Introduce consume_raw_n()

arighi:
libbpf provides the ring_buffer__consume_n() API to consume up to a certain amount of items from ringbuffers [1].

Add the new method consume_raw_n() also in RingBuffer to provide the same functionality.

This is needed by sched_ext to allow user-space schedulers (based on scx_rustland_core) to selectively consume task items from BPF ring buffer [2].

[1] https://lore.kernel.org/bpf/20240406092005.92399-1-andrea.righi@canonical.com/
[2] https://github.com/sched-ext/scx/blob/ea361886f714b968092a1ea14bdee8330beb4d5f/rust/scx_rustland_core/assets/bpf.rs#L228

### DIFF
--- a/libbpf-rs/tests/test.rs
+++ b/libbpf-rs/tests/test.rs
@@ -986,6 +986,23 @@ fn test_object_ringbuf_raw() {
     // Consume from a (potentially) empty ring buffer using poll()
     let ret = mgr.poll_raw(Duration::from_millis(100));
     assert!(ret >= 0);
+
+    // Call getpid multiple times, to refill the ring buffer.
+    for _ in 1..=10 {
+        unsafe { libc::getpid() };
+    }
+
+    // Consume exactly one item
+    let ret = mgr.consume_raw_n(1);
+    assert!(ret == 1);
+
+    // Consume two items
+    let ret = mgr.consume_raw_n(2);
+    assert!(ret == 2);
+
+    // Consume all the remaining items, but no more than 10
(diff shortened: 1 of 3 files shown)


libbpf-rs/tests/test.rs
```diff
@@ -986,6 +986,14 @@ fn test_object_ringbuf_raw() {
...
     let ret = mgr.poll_raw(Duration::from_millis(100));
     assert!(ret >= 0);
+
+    // Call getpid multiple times, to refill the ring buffer.
+    unsafe { libc::getpid() };
+    unsafe { libc::getpid() };
+
+    // Consume exactly one item
+    let ret = mgr.consume_raw_n(1);
+    assert!(ret == 1);
```
d-e-s-o, line 996: Should we also add a test making sure that with two events (or more) events available we also consume & report two? `1` really could mean so many things, so I'd think there is some value add.

d-e-s-o (approved):
The change looks good to me, thanks. Left one question/suggestion.

libbpf-rs/tests/test.rs
arighi: That's a good idea, I'll add a test for this case as well, thanks.

arighi:
Updated the `consume_raw_n()` selftest to cover more cases.

d-e-s-o:
Thanks!

PR #1174, opened 2025-05-21
Remove unused log dependency

d-e-s-o:
The `log` dependency is unused. Remove it.


d-e-s-o:
Grml. Not so unused after all...

PR #1176, opened 2025-05-22
examples: ringbuf_multi: include vmlinux.h than linux/bpf.h

SidongYang:
Using linux/bpf.h could make compile error. Fix this with using vmlinux.h


danielocfb:
Thanks!

PR #1178, opened 2025-05-22
libbpf-cargo: Relative anonymous type naming for struct fields

### LINKED ISSUE #1161: libbpf-cargo: Getters for struct fields
Hi!
Thank you for libbpf-rs, it's amazing!

Now, the access to struct fields is performed directly:
Consider we have the iphdr struct:
```rs
#[derive(Debug, Default, Copy, Clone)]
#[repr(C)]
pub struct iphdr {
    pub __pad_0: [u8; 1],
    pub tos: u8,
    pub tot_len: u16,
    pub id: u16,
    pub frag_off: u16,
    pub ttl: u8,
    pub protocol: u8,
    pub check: u16,
    pub __anon_6: __anon_6,
}
```

If we want to access any field, we will to do
```rs
let tos = iph.tos;
```

The main problem I see here is with anonymous structures. In my case, to access saddr field of my struct, I will do `iph.__anon_6.addrs.saddr`. But what if we change our BPF program a little bit, like add a new anonymous union before `iphdr`? Now, we will access `saddr` with `iph.__anon_7.addrs.saddr`. This also involves problems with readability.

I think the solution may be to simulate access to these fields and implement fields as getters in Rust:
```rs
impl iphdr {
    pub fn check(&self) -> u16 { self.check }
    pub fn protocol(&self) -> u8 { self.protocol }
    pub fn tot_len(&self) -> u16 { self.tot_len }
    pub fn id(&self) -> u16 { self.id }
    pub fn frag_off(&self) -> u16 { self.frag_off }
    pub fn tos(&self) -> u8 { self.tos }
    pub fn ttl(&self) -> u8 { self.ttl }
    pub unsafe fn daddr(&self) -> u32 { unsafe { self.__anon_7.addrs.daddr } }
    pub unsafe fn saddr(&self) -> u32 { unsafe { self.__anon_7.addrs.saddr } }
}
```

Also, the getter semantics will allow us to implement bitfield access in the future.

I have written some reference recursive implementation for generator:

```diff
diff --git a/libbpf-cargo/src/gen/btf.rs b/libbpf-cargo/src/gen/btf.rs
index 3be3cba..0ea22a4 100644
--- a/libbpf-cargo/src/gen/btf.rs
+++ b/libbpf-cargo/src/gen/btf.rs
@@ -681,6 +681,74 @@ impl<'s> GenBtf<'s> {
         self.type_definition_for_composites_with_opts(def, dependent_types, t, &opts)
     }
 
+    fn generate_explicit_field_tree(
+        &self,
+        t: types::Composite<'_>,
+        vc: &mut HashMap<String, String>,
+        opts: &TypeDeclOpts,
+        precontext: &String,
+        mut field_unsafe: bool,
+    ) -> Result<()> {
+        if !t.is_struct {
+            field_unsafe = true;
+        }
+
+        for member in t.iter() {
+            let member_offset = match member.attr {
+                MemberAttr::Normal { offset } => offset,
+                // Bitfields are tricky to get correct, if at all possible. For
+                // now we just skip them, which results in them being covered by
+                // padding bytes.
+                MemberAttr::BitField { .. } => continue,
+            };
+
+            let field_ty = self
+                .type_by_id::<BtfType<'_>>(member.ty)
+                .unwrap()
+                .skip_mods_and_typedefs();
+
+            let field_name = if let Some(name) = member.name {
+                escape_reserved_keyword(name.to_string_lossy())
+            } else {
+                // Only anonymous unnamed unions should ever have no name set.
+                // We just name them the same as their anonymous type. As there
+                // can only be one member of this very type, there can't be a
+                // conflict.
+                self.type_map.type_name_or_anon(&field_ty)
+            };
+            let field_ty_str = type_declaration_impl(field_ty, &self.type_map, opts)?;
+            let field_ty_str = if is_unsafe(field_ty) {
+                Cow::Owned(format!("std::mem::MaybeUninit<{field_ty_str}>"))
+            } else {
+                Cow::Borrowed(field_ty_str.as_str())
+            };
+
+            if let Ok(composite) = TryInto::<types::Composite<'_>>::try_into(field_ty) {
+                if composite.is_empty_union() {
+                    // Skip empty union field; we do not generate a type for them.
+                    continue;
+                }
+                if let None = field_ty.name() {
+                    // Anonymous struct or union
+                    self.generate_explicit_field_tree(composite, vc, opts, &format!("{precontext}.{field_name}"), field_unsafe)?;
+                    continue;
+                }
+            }
+ 
+            let mut st = String::new();
+
+            if field_unsafe {
+                st.push_str(format!(r#"    pub unsafe fn {field_name}(&self) -> {field_ty_str} {{ unsafe {{ self{precontext}.{field_name} }} }}"#).as_str()); 
+            } else {
+                st.push_str(format!(r#"    pub fn {field_name}(&self) -> {field_ty_str} {{ self{precontext}.{field_name} }}"#).as_str()); 
+            }
+
+            vc.insert(String::from(field_name), st);
+        }
+
+        Ok(())
+    }
+
     fn type_definition_for_composites_with_opts<'a>(
         &'a self,
         def: &mut String,
@@ -697,6 +765,7 @@ impl<'s> GenBtf<'s> {
 
         // fields in the aggregate
         let mut agg_content: Vec<String> = Vec::new();
+        let mut getter_content: HashMap<String, String> = HashMap::new();
 
         // structs with arrays > 32 length need to impl Default
         // rather than #[derive(Default)]
@@ -812,6 +881,8 @@ impl<'s> GenBtf<'s> {
             agg_content.push(format!(r#"    pub {field_name}: {field_ty_str},"#));
         }
 
+        self.generate_explicit_field_tree(t, &mut getter_content, opts, &String::from(""), false)?;
+
         if t.is_struct {
             let struct_size = t.size();
             let padding = required_padding(offset, struct_size, &t, packed)?;
@@ -850,6 +921,16 @@ impl<'s> GenBtf<'s> {
         }
         writeln!(def, "}}")?;
 
+        writeln!(
+            def, 
+            r#"impl {name} {{"#,
+            name = self.type_map.type_name_or_anon(&t)
+        )?;
+        for (field_name, st) in getter_content {
+            writeln!(def, "{st}")?;
+        }
+        writeln!(def, "}}")?;
+
         // if required write a Default implementation for this struct
         if gen_impl_default {
             writeln!(
```

Few opinionable problems I noticed here so far: 
- Mutability: We may want to duplicate getters with a `_mut` suffix to allow mutable members
- Return type: Should our getter return a reference to the field or just copy it?
- Name collisions: What if C structure has field default, like ::default() method?

Waujito:
Previously, anonymous types were named as __anon_{id} with id as a global counter. This approach caused cascading renaming of all subsequent anonymous definitions after any C anonymous struct order/amount change. This was a major design problem not for only development but even portability, since here may be the case when something was changed in vmlinux.h leading to compilation errors.

This PR addresses the issue by localizing cascading effects to the local struct layer rather than the global layer. While the solution is not fully ideal - cascading still occurs within individual structs - it eliminates the need for wide Rust code modifications when the order of anonymous struct definitions changes.

The PR is breaking and should be included in major version update.

The PR fixes #1161

While this approach is ok, I still think the better way will be to use getters, since this is breaking and less intuitive.

### DIFF
--- a/libbpf-cargo/src/gen/btf.rs
+++ b/libbpf-cargo/src/gen/btf.rs
@@ -369,10 +369,17 @@ fn escape_reserved_keyword(identifier: Cow<'_, str>) -> Cow<'_, str> {
 }
 
 #[derive(Debug, Clone)]
-pub struct BtfDependency {
-    pub name: Option<String>,
-    pub dep_id: i32,
-    pub child_counter: Rc<RefCell<i32>>,
+struct BtfDependency {
+    /// Name of the dependency parent
+    parent_name: Option<String>,
+
+    /// Dependency id relative to the parent's `child_counter`
+    dep_id: i32,
+
+    /// The `child_counter` for the dependency if it is intended to be
+    /// a parent itself.
+    /// For an anonymous unit this should be a pointer to the parent's `child_counter`
+    child_counter: Rc<RefCell<i32>>,
 }
 
 #[derive(Debug, Default)]
@@ -386,53 +393,47 @@ pub(crate) struct TypeMap {
     /// name already.
     names_count: RefCell<HashMap<String, u8>>,
 
-    dependencies: RefCell<HashMap<TypeId, BtfDependency>>,
+    /// Mapping from type to it's parent. Used in anonymous members naming
+    dependency_tree: RefCell<HashMap<TypeId, BtfDependency>>,
 }
 
 impl TypeMap {
-    pub fn derive_parent<'s>(&self, ty: &BtfType<'s>, parent: &BtfType<'s>) {
-        let mut deps = self.dependencies.borrow_mut();
+    fn register_parent<'s>(&self, ty: &BtfType<'s>, parent: &BtfType<'s>) {
+        let mut deps = self.dependency_tree.borrow_mut();
         if deps.get(&ty.type_id()).is_some() {
             return;
         }
 
-        let parent_dep = deps.get(&parent.type_id());
-        if let Some(pdep) = parent_dep {
-            let mut dep = pdep.clone();
+        let pdep = deps.entry(parent.type_id()).or_insert(BtfDependency {
+            parent_name: None,
+            dep_id: 0,
+            child_counter: Rc::new(RefCell::new(0)),
+        });
 
-            if let Some(n) = parent.name() {
-                dep.name = Some(n.to_string_lossy().to_string());
-            }
-            if ty.name().is_some() {
-                dep.child_counter = Rc::new(RefCell::new(0));
-            }
+        let mut dep = pdep.clone();
 
-            let parent_counter = Rc::<RefCell<i32>>::clone(&pdep.child_counter);
-            *parent_counter.borrow_mut() += 1;
-            dep.dep_id = *parent_counter.borrow();
-
-            deps.insert(ty.type_id(), dep);
-        } else {
-            let mut dep = BtfDependency {
-                name: None,
-                dep_id: 0,
-                child_counter: Rc::new(RefCell::new(1)),
-            };
-            deps.insert(parent.type_id(), dep.clone());
-
-            if let Some(n) = parent.name() {
-                dep.name = Some(n.to_string_lossy().to_string());
-            }
-            if ty.name().is_some() {
-                dep.child_counter = Rc::new(RefCell::new(0));
-            }
-            dep.dep_id = 1;
-            deps.insert(ty.type_id(), dep);
+        // If parent is named, derive it.
+        // Otherwise derive parent's parent
+        if let Some(n) = parent.name() {
+            dep.parent_name = Some(n.to_string_lossy().to_string());
         }
+
+        // If the current unit is named, self-assign the child_counter.
+        // Otherwise derive a parent's one
+        if ty.name().is_some() {
+            dep.child_counter = Rc::new(RefCell::new(0));
+        }
+
+        // Increment parent's `child_counter` and assign the new value to dep_id
+        let parent_counter = Rc::clone(&pdep.child_counter);
+        *parent_counter.borrow_mut() += 1;
+        dep.dep_id = *parent_counter.borrow();
+
+        deps.insert(ty.type_id(), dep);
     }
 
-    pub fn lookup_parent<'s>(&self, ty: &BtfType<'s>) -> Option<BtfDependency> {
-        self.dependencies.borrow().get(&ty.type_id()).cloned()
+    fn lookup_parent<'s>(&self, ty: &BtfType<'s>) -> Option<BtfDependency> {
+        self.dependency_tree.borrow().get(&ty.type_id()).cloned()
     }
 
     pub fn type_name_or_anon<'s>(&self, ty: &BtfType<'s>) -> Cow<'s, str> {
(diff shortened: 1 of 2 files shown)


danielocfb:
Haven't looked at it yet, but this is great. I was considering doing that for `0.25.0`, but didn't have the cycles. Will postpone the release until I've gotten a chance to take a look here. Please look into the CI failures, which all seem legit.

Waujito:
Here are still some plain __anon types like in `maps`. I'll try to catch more and may be add some defenders of name collisions.

libbpf-cargo/src/gen/btf.rs
```diff
@@ -377,15 +385,70 @@ pub(crate) struct TypeMap {
...
+        let parent_dep = deps.get(&parent.type_id());
+        if let Some(pdep) = parent_dep {
+            let mut dep = pdep.clone();
+
+            if let Some(n) = parent.name() {
+                dep.name = Some(n.to_string_lossy().to_string());
+            }
+            if ty.name().is_some() {
+                dep.child_counter = Rc::new(RefCell::new(0));
+            }
```
danielocfb, line 408: Why is this code needed?

libbpf-cargo/src/gen/btf.rs
```diff
@@ -377,15 +385,70 @@ pub(crate) struct TypeMap {
...
+            };
+            deps.insert(parent.type_id(), dep.clone());
+
+            if let Some(n) = parent.name() {
+                dep.name = Some(n.to_string_lossy().to_string());
+            }
+            if ty.name().is_some() {
+                dep.child_counter = Rc::new(RefCell::new(0));
+            }
+            dep.dep_id = 1;
```
danielocfb, line 429: Why is this value set here and not already during initialization?

libbpf-cargo/src/gen/btf.rs
```diff
@@ -377,15 +385,70 @@ pub(crate) struct TypeMap {
...
+            let mut dep = BtfDependency {
+                name: None,
+                dep_id: 0,
+                child_counter: Rc::new(RefCell::new(1)),
+            };
+            deps.insert(parent.type_id(), dep.clone());
+
+            if let Some(n) = parent.name() {
+                dep.name = Some(n.to_string_lossy().to_string());
+            }
```
danielocfb, line 425: Why is this value set here and not during initialization?

libbpf-cargo/src/gen/btf.rs
```diff
@@ -377,15 +385,70 @@ pub(crate) struct TypeMap {
     /// Mapping from type name to the number of times we have seen this
     /// name already.
     names_count: RefCell<HashMap<String, u8>>,
+
+    dependencies: RefCell<HashMap<TypeId, BtfDependency>>,
 }
 
 impl TypeMap {
+    pub fn derive_parent<'s>(&self, ty: &BtfType<'s>, parent: &BtfType<'s>) {
```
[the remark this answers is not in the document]
danielocfb: Is the name really fitting? Aren't you more registering a parent relationship rather than "deriving" anything?

libbpf-cargo/src/gen/btf.rs
```diff
@@ -367,6 +368,13 @@ fn escape_reserved_keyword(identifier: Cow<'_, str>) -> Cow<'_, str> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct BtfDependency {
+    pub name: Option<String>,
+    pub dep_id: i32,
+    pub child_counter: Rc<RefCell<i32>>,
```
danielocfb, line 375: Please document what these represent.

danielocfb (commented):
In general this seems okay to me, but the code seems under documented and under-tested in the sense that I can remove/change various lines and all tests keep passing. Similarly, I am not sure I understand some of the assignments done. Left some comments.

danielocfb:
> Here are still some plain __anon types like in `maps`. I'll try to catch more and may be add some defenders of name collisions.

What do you mean exactly? Where is "here"?

Waujito:
In [0e504e6](https://github.com/libbpf/libbpf-rs/pull/1178/commits/0e604e6b803c68dc613613657247c0129269637c) I simplified the code slightly, modified the global anon counting behavior, added documentation for new structure fields and included a test for depth-aligned anonymous structs.

The global anonymous struct counting now excludes relative anons to make its behavior more representative and easier to use. I also added some fallback protection on it for (im)possible case when global anonymous struct is assigned before a relative one: in such case we will use old type of naming.


> > Here are still some plain __anon types like in `maps`. I'll try to catch more and may be add some defenders of name collisions.
> 
> What do you mean exactly? Where is "here"?

Nvm. Those anonymous structs are at the global level and should be kept with old naming.

I think the code is ok now, but I'm aware of possible name collisions for relative types. 
I can't mind the case, but *what if*. Maybe I should implement dep id counting in a separate hashmap for parent names as strings, but not with that crazy `child_counter`? This solution will not be elegant, but more trustworthy...

libbpf-cargo/src/gen/btf.rs
Waujito: Added a comment

libbpf-cargo/src/gen/btf.rs
Waujito: Added a comment

libbpf-cargo/src/gen/btf.rs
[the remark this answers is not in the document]
Waujito: assign_parent?

danielocfb:
Thanks, it looks better now!

> I think the code is ok now, but I'm aware of possible name collisions for relative types.
> I can't mind the case, but _what if_.

Can you provide an example of such a case?

libbpf-cargo/src/gen/btf.rs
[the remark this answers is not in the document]
Waujito: Done

Waujito:
> Can you provide an example of such a case?

I'm almost sure here is no the case. I was just thinking about it

PR #1179, opened 2025-05-22
libbpf-cargo: Replace `CompilationOutput` type with `tracing` based solution

danielocfb:


danielocfb:
cc @JakeHillion

PR #1196, opened 2025-07-15
libbpf-cargo: use indoc in test strings

JakeHillion:
libbpf-cargo has a large number of test strings for both input and generated code. IMO this makes the test file very hard to read, especially given my vim has a really hard time syntax highlighting it - it's unclear whether the 0 indented code is test code or part of the test's Rust.

Use `indoc!` to allow us to indent this test text as we see fit, allowing them to look like part of a test case instead of jumping back to 0 indentation. It's purely stylistic.

Test plan:
- `cargo test -p libbpf-cargo` - passed before, passes after, with the same number of tests.
- CI

Signed-off-by: Jake Hillion <jake@hillion.co.uk>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -433,6 +433,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "indoc"
+version = "2.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


danielocfb:
Seems fine to me. PR is still marked as draft though?

PR #1197, opened 2025-07-17
libbpf-cargo: Place included object files in .bpf.objs section

JakeHillion:
libbpf-cargo generated skeletons include the blob of bytes of the .bpf.o
object file that's created. This is used by libbpf-rs to load the
program, but is otherwise not exposed.

Tools like veristat operate on BPF object files to give additional
information, such as how many instructions each program requires to
verify on the loaded kernel and whether they verify at all. The current
method to run veristat in CI runs for sched_ext is running `find` on the
Cargo target/ directory and finding all .bpf.o files. While this does
work, it's crude.

Use Rust's `link_section` attribute in generated skeletons to place the
contents of the object files in their own section, `.bpf.objs`. This
allows a simple script like [1] to easily extract all skeleton included
objects to a directory from the final binary. Features like LTO also
help here, as any unused object files are likely to have been removed at
link time.

Test plan:
- Added binary path output to `build_rust_project_from_bpf_c` and tested
  for symbol presence/correctness with Goblin.
- Built all sched_ext[2] Rust schedulers with this patch and used the
  aforementioned script[1] to extract the BPF object files. We could
  then run veristat on these object files as part of our CI pipelines.

[1] https://github.com/sched-ext/scx/blob/1c815584be36220e17a98d27811f9115ac5ec31c/scripts/extract_bpf_objects.sh
[2] https://github.com/sched-ext/scx

Signed-off-by: Jake Hillion <jake@hillion.co.uk>

### DIFF
--- a/libbpf-cargo/src/test.rs
+++ b/libbpf-cargo/src/test.rs
@@ -17,6 +17,7 @@ use libbpf_rs::btf::BtfType;
 use libbpf_rs::Btf;
 use memmap2::Mmap;
 use tempfile::tempdir;
+use tempfile::NamedTempFile;
 use tempfile::TempDir;
 use test_log::test;
 
@@ -371,8 +372,8 @@ fn test_make_workspace() {
         .exists());
 }
 
-fn build_rust_project_from_bpf_c_impl(bpf_c: &str, rust: &str, run: bool) {
-    let (_dir, proj_dir, cargo_toml) = setup_temp_project();
+fn build_rust_project_from_bpf_c_impl(bpf_c: &str, rust: &str, run: bool) -> (TempDir, PathBuf) {
+    let (dir, proj_dir, cargo_toml) = setup_temp_project();
 
     // Add prog dir
     create_dir(proj_dir.join("src/bpf")).expect("failed to create prog dir");
@@ -426,16 +427,18 @@ fn build_rust_project_from_bpf_c_impl(bpf_c: &str, rust: &str, run: bool) {
         .status()
         .expect("failed to run cargo");
     assert!(status.success());
+
+    (dir, proj_dir.join("target/debug/proj"))
 }
 
-fn build_rust_project_from_bpf_c(bpf_c: &str, rust: &str) {
+fn build_rust_project_from_bpf_c(bpf_c: &str, rust: &str) -> (TempDir, PathBuf) {
     let run = false;
     build_rust_project_from_bpf_c_impl(bpf_c, rust, run)
 }
 
 fn run_rust_project_from_bpf_c(bpf_c: &str, rust: &str) {
     let run = true;
-    build_rust_project_from_bpf_c_impl(bpf_c, rust, run)
+    build_rust_project_from_bpf_c_impl(bpf_c, rust, run);
 }
 
 #[test]
@@ -460,7 +463,7 @@ fn test_skeleton_empty_source() {
         }
     "#}
     .to_string();
-    let () = build_rust_project_from_bpf_c(&bpf_c, &rust);
+    build_rust_project_from_bpf_c(&bpf_c, &rust);
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


JakeHillion:
> Proposal makes sense to me; no objection from my side. Let's bump MSRV to `1.82.0` and then use `unsafe` attribute and remove `rustversion` dependency.

Thanks! Done.

> As for testing, perhaps use [`goblin`](https://crates.io/crates/goblin) or so to locate `.bpf.objs` section, get reference to contents, and then do basic testing akin to what [`test_object_build_from_memory`](https://github.com/libbpf/libbpf-rs/blob/6554564d1f5a4cee7ec61f5b9bd75911a48b21d5/libbpf-rs/tests/test.rs#L72) does? Easiest would probably be to put that into an example that we [run in CI](https://github.com/libbpf/libbpf-rs/blob/6554564d1f5a4cee7ec61f5b9bd75911a48b21d5/.github/workflows/test.yml#L68).

Turns out this was a bit easier than I thought, and it didn't take many goblin commands to get the data out. I adjusted `build_rust_project_from_bpf_c` to return the binary and it otherwise worked. If there's a preference for an example I can do that too, though I think anyone using this will have their own pipelines and constraints so not sure how useful it will be.

.github/workflows/test.yml
```diff
@@ -142,7 +142,7 @@ jobs:
         with:
           # Please adjust README and rust-version field in Cargo.toml files when
           # bumping version.
-          toolchain: 1.78.0
+          toolchain: 1.82.0
```
d-e-s-o, line 145: Can you please create a separate commit for the MSRV bump? Let's please also bump it in `libbpf-rs`'s README and add a CHANGELOG entry there as well (we haven't done that consistently but it makes the most sense; we don't test both in isolation).

libbpf-cargo/src/test.rs
```diff
@@ -518,7 +521,96 @@ fn test_skeleton_basic() {
...
+    let bpf_c = indoc! {r#"
+        #include "vmlinux.h"
+        #include <bpf/bpf_helpers.h>
+
+        struct {
+                __uint(type, BPF_MAP_TYPE_HASH);
+                __uint(max_entries, 1024);
+                __type(key, u32);
+                __type(value, u64);
+        } mymap SEC(".maps");
```
d-e-s-o, line 538: Let's stick with 4 spaces for indentation per level, please.
```suggestion
        struct {
            __uint(type, BPF_MAP_TYPE_HASH);
            __uint(max_entries, 1024);
            __type(key, u32);
            __type(value, u64);
        } mymap SEC(".maps");
```

libbpf-cargo/src/test.rs
```diff
@@ -518,7 +521,96 @@ fn test_skeleton_basic() {
...
+
+            // Check that we can grab handles to loaded maps/progs
+            let _map = &skel.maps.mymap;
+            let _prog = &skel.progs.this_is_my_prog;
+
+            // Check that attach() is generated
+            skel.attach().expect("failed to attach progs");
+
+            // Check that Option<Link> field is generated
+            let _mylink = skel.links.this_is_my_prog.unwrap();
```
d-e-s-o, line 577: Do we really need all these supposed checks here? Looks more like copy&paste remnants to me.
```suggestion
            let mut open_object = MaybeUninit::uninit();
            let open_skel = builder
                .open(&mut open_object)
                .expect("failed to open skel");
            let _skel = open_skel.load().expect("failed to load skel");
```

libbpf-cargo/src/test.rs
```diff
@@ -518,7 +521,96 @@ fn test_skeleton_basic() {
     "#}
     .to_string();
 
-    let () = build_rust_project_from_bpf_c(&bpf_c, &rust);
+    build_rust_project_from_bpf_c(&bpf_c, &rust);
+}
+
+#[test]
+fn test_skeleton_generate_bpf_objs_section() {
```
d-e-s-o, line 528:
```suggestion
/// Validate that `.bpf.objs` section contents are as expected.
#[test]
fn test_skeleton_generate_bpf_objs_section() {
```

d-e-s-o (commented):
Looks mostly good, thanks! Left a few more comments, please take another look and let me know what you think.

.github/workflows/test.yml
JakeHillion: Created a separate commit, let me know if you'd rather a separate PR too.

libbpf-cargo/src/test.rs
JakeHillion: Removed as suggested, and also removed most of the BPF-C program that isn't necessary for the test.

JakeHillion:
> will this work fine if you have two separate skeletons used from single application?

Yes, it's all good. It's the same mangled name as if they were all in .rodata, just in a different section.

anakryiko:
> > will this work fine if you have two separate skeletons used from single application?
> 
> Yes, it's all good. It's the same mangled name as if they were all in .rodata, just in a different section.

I'm asking more from the POV of recovering original object file or skeleton name. Is that possible?

JakeHillion:
> > Yes, it's all good. It's the same mangled name as if they were all in .rodata, just in a different section.
> 
> I'm asking more from the POV of recovering original object file or skeleton name. Is that possible?

I'll share an example of a sched_ext scheduler that uses multiple skeletons:

```
jake@merlin:/data/users/jake/repos/scx/ > ./scripts/extract_bpf_objects.sh target/release/scx_layered /tmp/nix-shell.lP3beS/tmp.OmXAB3yjRn
Extracting '_ZN14scx_bpf_compat8bpf_skel3imp4DATA17h36943cc0a407c2beE.bpf.o'...
Extracting '_ZN11scx_layered8bpf_skel3imp4DATA17h96c4c404bf2b1476E.bpf.o'...
Successfully extracted 2 BPF object files to /tmp/nix-shell.lP3beS/tmp.OmXAB3yjRn
jake@merlin:/data/users/jake/repos/scx/ > ls -l /tmp/nix-shell.lP3beS/tmp.OmXAB3yjRn
total 336
-rw-r--r-- 1 jake users 337712 Jul 19 20:32 _ZN11scx_layered8bpf_skel3imp4DATA17h96c4c404bf2b1476E.bpf.o
-rw-r--r-- 1 jake users   2024 Jul 19 20:32 _ZN14scx_bpf_compat8bpf_skel3imp4DATA17h36943cc0a407c2beE.bpf.o
jake@merlin:/data/users/jake/repos/scx/ > ~/.cargo/bin/rustfilt _ZN11scx_layered8bpf_skel3imp4DATA17h96c4c404bf2b1476E
scx_layered::bpf_skel::imp::DATA
jake@merlin:/data/users/jake/repos/scx/ > ~/.cargo/bin/rustfilt _ZN14scx_bpf_compat8bpf_skel3imp4DATA17h36943cc0a407c2beE
scx_bpf_compat::bpf_skel::imp::DATA
```

The crate path is there in the mangled name, but the rest is fixed. I'm not sure what it looks like if you have two `bpf_skel` in the same crate. The `bpf_skel` is defined by the Rust module path, but can be whatever you need, and this should be correctly uniqued. The temporary object file name is gone now though.

I think unless we remove the mangling it would be hard to make this any more specific, and I'm not sure how useful an extension that is to the API - the name already describes how it is used in the final binary which I believe is more useful.

PR #1213, opened 2025-08-12
query: Extract tracepoint & k(ret)probe from perf_event links

Jongy:
heavily inspired by matching bpftool code


danielocfb:
Are you still interested in bringing this over the finish line, @Jongy? Do you need help with something?

Jongy:
@danielocfb , I pushed my branch with fixes and tests, PTAL. The PR is closed and I seem to be unable to reopen it - I guess that once you do so, it will update with the new diff.

PR #1220, opened 2025-08-15
Fix TODO: use from_bytes_until_nul

frelon:
Fix TODO item which recommends using CStr::from_bytes_until_nul over CStr::from_bytes_with_nul and calculating the first nul index using an iterator.


danielocfb (approved):
Thank you!

PR #1230, opened 2025-09-05
Introduce Program::attach_iter_with_opts()

d-e-s-o:
Add the `Program::attach_iter_with_opts()` method to allow for iterator attachment with support for providing additional options.


tyrone-wu:
Heyo o/

This was how I ended up implementing the iter variants when I implemented link_info in aya.
https://github.com/tyrone-wu/aya/blob/aya/link-info/aya/src/links/info.rs#L118

It's kind old code (about a yearish ago) that I ended up not having the time & energy to get merged to upstream aya, and probly won't get around to since i'm kinda moving away from aya. 🫠 so it's been sitting in limbo ever since.

If it's still the same today, the `.target_name` should map to the `bpf_iter_link_info` variants:
- `"bpf_map_elem" | "bpf_sk_storage_map"` to `map`
- `"cgroup"` to `cgroup`
- `"task" | "task_file" | "task_vma"` to `task`

d-e-s-o:
> This was how I ended up implementing the iter variants when I implemented link_info in aya.
> https://github.com/tyrone-wu/aya/blob/aya/link-info/aya/src/links/info.rs#L118

Thanks for the pointer. Based on my understanding, this looks to be more about the querying part, right? Whereas this change is about the attachment configuration.

Be that as it may, the change at hand seems like a reasonable extension of what we already have and should fit in with the other attachment methods. That, and I got confirmation that it works for the intended use case, so I think we can finally move this along.

PR #1253, opened 2025-10-06
query: add u(ret)probe & perf_event to perf_event link info

### LINKED ISSUE #1251: Support for `bpf_link_info.perf_event`
Hi o/

I was wondering if anyone is currently working `bpf_link_info.perf_event`.

If not, I could implement this.
I've already implemented it when I was using aya, https://github.com/tyrone-wu/aya/blob/aya/link-info/aya/src/links/info.rs#L301, but now i've kinda been moving back to this library.

ver-nyan:
Add the remaining bits of `bpf_link_info.perf_event` for `.uprobe` and `.event`.

Fixes #1251.

### DIFF
--- a/examples/bpf_query/src/main.rs
+++ b/examples/bpf_query/src/main.rs
@@ -173,6 +173,17 @@ fn link() {
                         print!(" ref_ctr_offset={ref_ctr_offset}");
                     }
                 }
+                query::PerfEventType::Event {
+                    config,
+                    event_type,
+                    cookie,
+                } => {
+                    print!("    event");
+                    print!(" type={event_type} config={config}");
+                    if cookie != 0 {
+                        print!(" cookie={cookie}");
+                    }
+                }
                 query::PerfEventType::Unknown(ty) => {
                     println!("    unknown perf event type: {ty}");
                 }
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -773,8 +773,16 @@ pub enum PerfEventType {
         /// Offset of kernel reference counted USDT semaphore.
         ref_ctr_offset: u64,
     },
+    /// A perf event.
+    Event {
+        /// The specific event of the perf event type.
+        config: u64,
+        /// The perf event type.
+        event_type: u32,
+        /// Cookie value for the perf event program.
+        cookie: u64,
+    },
     /// An unknown or unsupported perf event type.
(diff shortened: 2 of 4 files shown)


tyrone-wu:
It looks like `libbpf-sys` is pinned to v1.5, and `.ref_ctr_offset` was added in v1.6. I'll comment it out for now with a `// TODO`, and once the bindings have caught up, it can be uncommented.

As for the `perf_event` tests, I had a suspicion that the GH runners wouldn't be able to support opening perf_events, and that looks to be the case based on the logs. I'll dig into the runners to see if it's possible to enable it.

d-e-s-o:
> It looks like `libbpf-sys` is pinned to v1.5, and `.ref_ctr_offset` was added in v1.6. I'll comment it out for now with a `// TODO`, and once the bindings have caught up, it can be uncommented.

We can probably bump the range. Let me look into it.

> As for the `perf_event` tests, I had a suspicion that the GH runners wouldn't be able to support opening perf_events, and that looks to be the case based on the logs. I'll dig into the runners to see if it's possible to enable it.

If not, you can ignore it: https://github.com/libbpf/libbpf-rs/blob/a953b11e81a66562a84c891617152e22bd9d59ca/libbpf-rs/tests/test.rs#L1795

tyrone-wu:
Ah, I just noticed the `.ref_ctr_offset` uprobe info field appears to be fairly recent, added in kernel v6.16. The `uprobe` info tests current leaves `ref_ctr_offset` 0/unset, so that explains why it was passing on my local machine. I haven't tested it, but I'm not sure if the runner will have that field too.

d-e-s-o:
> It looks like libbpf-sys is pinned to v1.5, and .ref_ctr_offset was added in v1.6.

I don't think we pin per se. But in CI we should be using 1.6.1, as that was what we had in our lock file. I am bumping it with https://github.com/libbpf/libbpf-rs/pull/1255, but it shouldn't make a difference for the change at hand, I'd think.

d-e-s-o:
To make CI pass I think you'll need to bump the minimum `libbpf-sys` version:
```patch
--- Cargo.toml
+++ Cargo.toml
@@ -36,7 +36,7 @@ dont-generate-test-files = []

 [dependencies]
 bitflags = "2.0"
-libbpf-sys = { version = "1.5.0", default-features = false }
+libbpf-sys = { version = "1.6.0", default-features = false }
 libc = "0.2"
 vsprintf = "2.0"

```

PR #1266, opened 2025-11-06
Introduce Program::attach_uprobe_multi{with_opts}()

sidchintamaneni:
Add Program::attach_uprobe_multi & Program::attach_uprobe_multi_with_opts
to enable attachment of multiple uprobes at once.

### DIFF
--- a/libbpf-rs/src/program.rs
+++ b/libbpf-rs/src/program.rs
@@ -65,6 +65,25 @@ pub struct UprobeOpts {
     pub _non_exhaustive: (),
 }
 
+/// Options to optionally be provided when attaching to a uprobe.
+#[derive(Clone, Debug, Default)]
+pub struct UprobeMultiOpts {
+    /// Optional, array of function symbols to attach to
+    pub syms: Vec<String>,
+    /// Optional, array of function addresses to attach to
+    pub offsets: Vec<usize>,
+    /// Optional, array of associated ref counter offsets
+    pub ref_ctr_offsets: Vec<usize>,
+    /// Optional, array of associated BPF cookies
+    pub cookies: Vec<u64>,
+    /// Create return uprobes
+    pub retprobe: bool,
+    /// Create session uprobes
+    pub session: bool,
+    #[doc(hidden)]
+    pub _non_exhaustive: (),
+}
+
 /// Options to optionally be provided when attaching to a USDT.
 #[derive(Clone, Debug, Default)]
 pub struct UsdtOpts {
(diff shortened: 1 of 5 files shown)


libbpf-rs/src/program.rs
```diff
@@ -65,6 +65,25 @@ pub struct UprobeOpts {
     pub _non_exhaustive: (),
 }
 
+/// Options to optionally be provided when attaching to a uprobe.
+#[derive(Clone, Debug, Default)]
+pub struct UprobeMultiOpts {
+    /// array of function symbols to attach to
```
danielocfb: Can you please start doc comments with upper case letters?

libbpf-rs/src/program.rs
```diff
@@ -65,6 +65,25 @@ pub struct UprobeOpts {
...
+    pub syms: Option<Vec<String>>,
+    /// array of function addresses to attach to
+    pub offsets: Option<Vec<usize>>,
+    /// optional, array of associated ref counter offsets
+    pub ref_ctr_offsets: Option<Vec<usize>>,
+    /// optional, array of associated BPF cookies
+    pub cookies: Option<Vec<u64>>,
+    /// create return uprobes
+    pub retprobe: bool,
+    /// create session kprobes
```
danielocfb: `uprobes`?

libbpf-rs/tests/test.rs
```diff
@@ -1830,11 +1831,61 @@ fn test_object_raw_tracepoint_with_opts() {
...
+#[test]
+fn test_object_uprobe_multi() {
+    let mut obj = get_test_object("uprobe_multi.bpf.o");
+    let prog = get_prog_mut(&mut obj, "handle__uprobe_multi");
+
+    let pid = unsafe { libc::getpid() };
+    let path = current_exe().expect("failed to find executable name");
+
+    let _link = prog
+        .attach_uprobe_multi(pid, path, "uprobe_multi_func_*", false, false)
```
danielocfb: Can you please use named argument to make this call more readable?
```rust
    let _link = prog
        .attach_uprobe_multi(pid, path, "uprobe_multi_func_*", retprobe, session)
```

libbpf-rs/src/program.rs
```diff
@@ -65,6 +65,25 @@ pub struct UprobeOpts {
     pub _non_exhaustive: (),
 }
 
+/// Options to optionally be provided when attaching to a uprobe.
+#[derive(Clone, Debug, Default)]
+pub struct UprobeMultiOpts {
+    /// array of function symbols to attach to
+    pub syms: Option<Vec<String>>,
+    /// array of function addresses to attach to
+    pub offsets: Option<Vec<usize>>,
```
danielocfb: I am admittedly a bit torn on the `Option` usage. A `Vec` is implicitly optional in the sense that it can be empty. I feel like just relying on that and perhaps documenting that it can be empty (which you basically already do if you state that it's optional) should suffice. What do you think?

danielocfb (commented):
Thanks for working on this feature! Overall this looks good. Left a few mostly stylistic comments.

The main gripe I have is that the testing should probably include actually triggering the uprobes, or we don't actually know what we are testing. That's what is being done in, say, `test_object_uprobe_with_opts`. Can you check whether that's possible?

Please also make sure to format properly from the start. No "fixup" commits. So just merge the two, please.

libbpf-rs/src/program.rs
sidchintamaneni: make sense, where do you want to document it? I think you mean something like 
```
/// Optional, array of function symbols to attach to
pub syms: Vec<String>,
```

danielocfb: Yeah, that sounds good.

libbpf-rs/src/program.rs
```diff
@@ -1034,6 +1053,120 @@ impl<'obj> ProgramMut<'obj> {
...
+            retprobe,
+            session,
+            _non_exhaustive,
+        } = opts;
+
+        let pattern = util::str_to_cstring(func_pattern.as_ref())?;
+        let pattern_ptr = pattern.as_ptr();
+
+        let syms_cstrings = if !syms.is_empty() {
+            Some(
```
d-e-s-o: Why do we need an `Option` here? It seems we should be able to use the same pattern we use for all other vectors, no?

libbpf-rs/tests/test.rs
```diff
@@ -1830,11 +1831,107 @@ fn test_object_raw_tracepoint_with_opts() {
...
+fn test_object_uprobe_multi_with_opts() {
+    let mut obj = get_test_object("uprobe_multi.bpf.o");
+    let prog = get_prog_mut(&mut obj, "handle__uprobe_multi_with_opts");
+    let func_pattern = "multi_uprobe_func_*";
+
+    let pid = unsafe { libc::getpid() };
+    let path = current_exe().expect("failed to find executable name");
+    let opts = UprobeMultiOpts {
+        ..Default::default()
+    };
```
d-e-s-o:
```suggestion
    let opts = UprobeMultiOpts::default();
```

d-e-s-o:
> Thanks! this was fun. Let me know if there are more fun things to work on.

Let me do some digging. I will ping you once I found something.

PR #1274, opened 2025-11-19
program: Add OpenProgram.autoload()

Jongy:


Jongy:
> Hi @Jongy, thanks for the pull request. would it be possible to add a unit test for this functionality?

Sure! I pushed something to cover autoload/set_autoload on OpenProgram and Program.

PR #1284, opened 2025-12-09
Introduce BPF Streams abstraction

etsal:
Add libbpf-rs support for interacting with BPF streams from Rust code. The interface provides the option to either directly specify the prog fd/stream id, or to hide the details behind a Stream abstraction.

### DIFF
--- a/libbpf-rs/tests/bin/src/stream.bpf.c
+++ b/libbpf-rs/tests/bin/src/stream.bpf.c
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright (c) 2025 Meta Platforms, Inc. and affiliates. */
+
+#include "vmlinux.h"
+#include <bpf/bpf_helpers.h>
+
+/* Definition of can_loop taken from bpf_experimental.h. */
+#ifdef __BPF_FEATURE_MAY_GOTO
+#define can_loop                                                               \
+    ({                                                                         \
+        __label__ l_break, l_continue;                                         \
+        bool ret = true;                                                       \
+        asm volatile goto("may_goto %l[l_break]" :: ::l_break);                \
+        goto l_continue;                                                       \
+    l_break:                                                                   \
+        ret = false;                                                           \
+    l_continue:;                                                               \
+        ret;                                                                   \
+    })
+#else
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#define can_loop                                                               \
+    ({                                                                         \
+        __label__ l_break, l_continue;                                         \
+        bool ret = true;                                                       \
+        asm volatile goto("1:.byte 0xe5;		\
+		      .byte 0;				\
+		      .long ((%l[l_break] - 1b - 8) / 8) & 0xffff;	\
+		      .short 0" :: ::l_break);                                 \
+        goto l_continue;                                                       \
+    l_break:                                                                   \
+        ret = false;                                                           \
+    l_continue:;                                                               \
+        ret;                                                                   \
+    })
+#else
+#define can_loop                                                               \
+    ({                                                                         \
+        __label__ l_break, l_continue;                                         \
+        bool ret = true;                                                       \
+        asm volatile goto("1:.byte 0xe5;		\
+		      .byte 0;				\
+		      .long (((%l[l_break] - 1b - 8) / 8) & 0xffff) << 16;	\
+		      .short 0" :: ::l_break);                                 \
+        goto l_continue;                                                       \
+    l_break:                                                                   \
+        ret = false;                                                           \
+    l_continue:;                                                               \
+        ret;                                                                   \
+    })
+#endif
+#endif
+
+volatile u64 i;
+
+/*
+ * Trigger a may_goto timeout to emit a streams error. As of 6.19 the only way
+ * to trigger streams output is by causing an error condition in the program.
+ * One of these is a loop timeout: The may_goto macro allows for loops that
+ * cannot be verified by embedding a timer that is guaranteed to expire in the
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/streams.rs
```diff
@@ -0,0 +1,105 @@
...
+            bytes_per_read,
+        }
+    }
+
+    /// Get a default `StreamReadOpts` struct, since currently the struct
+    /// has no actual data and requiring the user to provide it is
+    /// unnecessarily complicated..
+    fn opts_dfl() -> StreamReadOpts {
+        StreamReadOpts { sz: 0 }
+    }
```
danielocfb: This constructor should probably be replaced by a [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) impl?

libbpf-rs/src/streams.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Default BPF stdout stream id.
+pub const BPF_STDOUT: u32 = 1;
+/// Default BPF stderr stream id.
+pub const BPF_STDERR: u32 = 2;
+
+/// Config options, currently empty and here for forward compatibility.
+struct StreamReadOpts {
+    sz: usize,
+}
```
danielocfb: I kind of fail to see the purpose of this type. It's not public, so users can't use it. So we don't need forward compat. And if it's empty, then we don't need it at all?

libbpf-rs/src/streams.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl Stream {
+    /// Default read batch for when the user doesn't specify one.
+    const BYTES_PER_READ_DFL: usize = 4096;
+
+    /// Create a new Stream instance.
+    pub fn new(fd: i32, stream_id: u32) -> Self {
```
danielocfb: So what is a "stream_id" and what is its purpose? It's unclear what relevance it has to the user. I don't see it mentioned in the only "documentation" that I can find on the topic, which is https://lwn.net/Articles/1026660/ (and its reincarnations)

danielocfb (changes requested):
Hi @etsal, thanks for the pull request. Left a few comments with suggestions for improvement/clarifying questions. Can you please make sure to add tests for the functionality?

In general it feels as if `Stream` should likely implement [`io::Read`](https://doc.rust-lang.org/std/io/trait.Read.html) to enable proper interop with other Rust components.

libbpf-rs/src/streams.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::Error;
+use crate::Result;
+
+/// [`Stream`] instances.
+///
+/// `streams`s are BPF descriptors that provide a character output
+/// interface from the BPF program to userspace.
+#[derive(Debug)]
+pub struct Stream {
+    fd: i32,
```
danielocfb: Rust generally uses file descriptor abstractions that encode ownership semantics see https://doc.rust-lang.org/stable/std/os/fd/index.html

etsal:
Thank you for the review, I will update the PR accordingly and post here to confirm it's ready for review.  Ack to all comments, though I have a couple  followup questions (inline).

libbpf-rs/src/streams.rs
etsal: The fd in question is the fd of the program whose stream we are reading. AFAICT the real fd is owned by the ProgramMut. Would it work to just embed the `Stream` into `ProgramMut`, then turn this fd into a BorrowedFd? I avoided doing so because `ProgramMut` is a pretty thin wrapper around the underlying C structs, but I think that would be the most natural place for `Stream`.

etsal: Actually, combining this with the comment about stdout() above: We don't need to embed anything, we can just add a method for a ProgramMut that creates a Stream object for its stdout or stderr. That would address both comments.

etsal:
Updated the code and added tests. The interface now presents the streams only indirectly and without exposing the type itself unless the user opts to create an instance directly. I think the code now is more idiomatic and looks a lot less like C.

libbpf-rs/tests/test_streams.rs
```diff
@@ -0,0 +1,66 @@
...
+    let mut obj = get_test_object("stream.bpf.o");
+    let prog = get_prog_mut(&mut obj, "trigger_streams");
+
+    let input = ProgramInput::default();
+    let _output = prog.test_run(input).unwrap();
+
+    let mut stdout = prog.stdout();
+    let mut buf = [0u8; 1024];
+
+    // The read itself should succeed and return 0 bytes
```
danielocfb: Am I misreading or does the comment state the opposite of what is being asserted?

libbpf-rs/tests/bin/src/stream.bpf.c
```diff
@@ -0,0 +1,65 @@
...
+        ret = false;                                                           \
+    l_continue:;                                                               \
+        ret;                                                                   \
+    })
+#endif
+#endif
+
+volatile u64 i;
+
+/* Trigger a may_goto timeout to emit a streams error. */
```
danielocfb: So this `may_goto` timeout is a way to force the kernel to emit data to stderr, is that how it works? And there is no way to write to a stream directly? And what are the requirements for `may_goto` to be available? Is it just a certain LLVM version? Or kernel config? It's also not exactly clear what happens if it's not available. Will the test still work?

If there is no easier way to trigger a stream write, can you please expand the comment a bit to make it clear who does what and in response to what exactly?

etsal:
Not sure why clippy is hitting these errors, they are for pre-existing code.

etsal:
Thank you for the detailed feedback!

PR #1286, opened 2025-12-15
Reorganize test code

danielocfb:
Reorganize the test code a bit by having only a single binary for integration tests. That's beneficial for compile and test times, but most importantly eliminates the need for each integration test to import the 'common' module, which can be confusing.


danielocfb:
I ran
> cargo test -- --test-threads=1 | grep '\.\.\. ok' | wc -l

before and after the change and both times `135` was reported.

Issue #1285, opened 2025-12-13
Generated structures miss padding


goofacz:
The `libbpf-cargo` generates structure with incorrect padding, if the last field it a bit-field.
```
    pub mod types {
        #[allow(unused_imports)]
        use super::*;
        #[derive(Debug, Default, Copy, Clone)]
        #[repr(C)]
        pub struct Foo {
            pub g: i8,
        }
        #[derive(Debug, Default, Copy, Clone)]
        #[repr(C)]
        pub struct Bar {
            pub g: i8,
        }
        // skipped
    }
```

is generated for C code:
```
#include <inttypes.h>

struct Foo {
    char     g;
    uint32_t h : 23;
};

struct Bar {
    char     g;
    uint32_t h : 23;
} __attribute__((__packed__));

const struct Foo f = {
    .g = 'K',
    .h = 0x1,
};

const struct Bar v = {
    .g = 'K',
    .h = 0x1,
};
```

BTF:
```
[1] CONST '(anon)' type_id=2
[2] STRUCT 'Foo' size=4 vlen=2
	'g' type_id=3 bits_offset=0
	'h' type_id=4 bits_offset=8 bitfield_size=23
[3] INT 'char' size=1 bits_offset=0 nr_bits=8 encoding=SIGNED
[4] TYPEDEF 'uint32_t' type_id=5
[5] TYPEDEF '__uint32_t' type_id=6
[6] INT 'unsigned int' size=4 bits_offset=0 nr_bits=32 encoding=(none)
[7] VAR 'f' type_id=1, linkage=global
[8] CONST '(anon)' type_id=9
[9] STRUCT 'Bar' size=4 vlen=2
	'g' type_id=3 bits_offset=0
	'h' type_id=4 bits_offset=8 bitfield_size=23
[10] VAR 'v' type_id=8, linkage=global
[11] DATASEC '.rodata' size=0 vlen=2
	type_id=7 offset=0 size=4 (VAR 'f')
	type_id=10 offset=0 size=4 (VAR 'v')
```

`pahole` output
```
struct Foo {
	char                       g;                    /*     0     1 */

	/* Bitfield combined with previous fields */

	uint32_t                   h:23;                 /*     0: 8  4 */

	/* size: 4, cachelines: 1, members: 2 */
	/* bit_padding: 1 bits */
	/* last cacheline: 4 bytes */
};
struct Bar {
	char                       g;                    /*     0     1 */

	/* Bitfield combined with previous fields */

[...]

PR #1288, opened 2025-12-15
libbpf-cargo: Improve handling of trailing bitfields

danielocfb:
Improve the handling of trailing bitfields in struct types, for which we missed the emitting of padding bytes.

Closes: #1285


danielocfb:
Not really sure whether that's a proper way to go about it. @anakryiko , do you have any opinions?

PR #1297, opened 2025-12-28
map: Add lookup_into() method

Jongy:
Allows looking up into a preallocated buffer.

This API is just a proposal - lmk what you think?

### DIFF
--- a/libbpf-rs/src/map.rs
+++ b/libbpf-rs/src/map.rs
@@ -311,8 +311,14 @@ where
     key.as_ptr() as *const c_void
 }
 
-/// Internal function to return a value from a map into a buffer of the given size.
-fn lookup_raw<M>(map: &M, key: &[u8], flags: MapFlags, out_size: usize) -> Result<Option<Vec<u8>>>
+/// Internal function to perform a map lookup and write the value into raw pointer.
+/// Returns `Ok(true)` if the key was found, `Ok(false)` if not found, or an error.
+fn lookup_raw<M>(
+    map: &M,
+    key: &[u8],
+    value: &mut [mem::MaybeUninit<u8>],
+    flags: MapFlags,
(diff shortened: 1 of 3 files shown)


libbpf-rs/src/map.rs
```diff
@@ -483,6 +507,32 @@ pub trait MapCore: Debug + AsFd + private::Sealed {
...
+    ///
+    /// Returns `Ok(true)` if the key was found and the buffer was filled,
+    /// `Ok(false)` if the key was not found, or an error.
+    ///
+    /// If the map is one of the per-cpu data structures, this function cannot be used.
+    /// If the map is of type `bloom_filter`, this function cannot be used.
+    fn lookup_into(&self, key: &[u8], value: &mut [u8], flags: MapFlags) -> Result<bool> {
+        check_not_bloom_or_percpu(self)?;
+
+        if value.len() != self.value_size() as usize {
```
Jongy: size check is done here and not in `lookup_raw_ptr` since `lookup_raw` does not require it

d-e-s-o:
> I'm not sure why the tests fail on nightly/dev

It's the debug assertion you added. It doesn't work for the per-CPU case.

Jongy:
> > I'm not sure why the tests fail on nightly/dev
> 
> It's the debug assertion you added. It doesn't work for the per-CPU case.

Haha nice. Will run the full test suite next time.
I'll find another place for this assert then

Jongy:
I removed the SAFETY from the non-unsafe function. I did keep the `debug_assert_eq`, just fixed the condition, because I think it's a helpful guard for this internal function, and it has zero cost in release builds. Practically I don't expect it to hit any of our users (in their debug builds) but if we're unsure, we can gate it with `cfg(test)`?

PR #1314, opened 2026-01-28
libbpf-rs: BTF custom path setter in ObjectBuilder

javierhonduco:
Allowing to set a custom BTF path in the object builder will reduce the boilerplate that's currently necessary. Making libbpf use a separate BTF file is incredibly important in systems that don't ship it. Looking at you, Nvidia and Raspberry Pi official kernels.

It also can be very helpeful when testing for ABI changes in kernels and it would have been quite helpful for a few investigations where kernel changes have prevented the loading of BPF programs as they either exposed bugs in our code.

Several users libbpf-rs that set this option currently leak memory (which might not be a big deal since it's a one-off and not a lot of memory) but having this as a blessed, leak-free way to do it would be great.

Test Plan
=========

Updated https://github.com/javierhonduco/lightswitch/pull/367/commits with these changes and verified test pass (which exercise setting an alternative BTF path) and that ASAN reports no issues.

### DIFF
--- a/libbpf-rs/CHANGELOG.md
+++ b/libbpf-rs/CHANGELOG.md
@@ -17,6 +17,7 @@ Unreleased
   buffers
 - Added `verified_insns` attribute to `query::ProgramInfo` type
 - Made `query::ProgramInfo` non-exhaustive
+- Added `ObjectBuilder::btf_custom_path` setter
 
 
 0.26.0-beta.0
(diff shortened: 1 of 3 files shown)


d-e-s-o (commented):
This seems fine to me, thanks, but can you please add a test for this functionality?

PR #1322, opened 2026-02-03
Add verified_insns attribute to query::ProgramInfo

gadol21:
Add verified_insns attribute, and populate it from the bpf_prog_info struct field.


d-e-s-o (approved):
Seems fine, thanks!

PR #1380, opened 2026-04-15
query: add kprobe_multi & uprobe_multi link info fields that require second info call

ver-nyan:
Added `KprobeMultiLinkInfo` & `UprobeMultiLinkInfo` fields that required a second info call.

### DIFF
--- a/libbpf-rs/src/query.rs
+++ b/libbpf-rs/src/query.rs
@@ -13,6 +13,7 @@
 use std::ffi::c_void;
 use std::ffi::CStr;
 use std::ffi::CString;
+use std::ffi::OsStr;
 use std::io;
 use std::mem::size_of_val;
 use std::mem::zeroed;
@@ -22,6 +23,8 @@ use std::os::fd::BorrowedFd;
 use std::os::fd::FromRawFd;
 use std::os::fd::OwnedFd;
 use std::os::raw::c_char;
+use std::os::unix::ffi::OsStrExt;
+use std::path::PathBuf;
 use std::ptr;
 use std::time::Duration;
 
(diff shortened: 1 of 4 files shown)


d-e-s-o:
Sorry for the delay, will try and get to it tomorrow.

ver-nyan:
No worries! this feature isn't urgent me.

libbpf-rs/src/query.rs
```diff
@@ -940,18 +952,70 @@ impl LinkInfo {
...
+                s.__bindgen_anon_1.uprobe_multi.offsets = offsets.as_mut_ptr() as u64;
+                s.__bindgen_anon_1.uprobe_multi.ref_ctr_offsets =
+                    ref_ctr_offsets.as_mut_ptr() as u64;
+                s.__bindgen_anon_1.uprobe_multi.cookies = cookies.as_mut_ptr() as u64;
+                let item_ptr: *mut libbpf_sys::bpf_link_info = &mut s;
+                let mut len = size_of_val(&s) as u32;
+                let ret = unsafe {
+                    libbpf_sys::bpf_obj_get_info_by_fd(
+                        fd.as_raw_fd(),
+                        item_ptr as *mut c_void,
```
d-e-s-o, line 1000:
```suggestion
                        item_ptr.cast(),
```

d-e-s-o: And can you please rebase on top of `master`?

libbpf-rs/src/query.rs
```diff
@@ -731,12 +731,20 @@ pub struct KprobeMultiLinkInfo {
 pub struct UprobeMultiLinkInfo {
     /// Size of the path.
     pub path_size: u32,
+    /// The absolute file path of the binary being probed.
+    pub path: Option<CString>,
```
d-e-s-o, line 735: If it's a path should we make it a `PathBuf`?

libbpf-rs/src/query.rs
```diff
@@ -972,11 +980,42 @@ impl LinkInfo {
...
+                };
+                if ret != 0 {
+                    return None;
+                }
+
+                let path = unsafe { s.__bindgen_anon_1.uprobe_multi.path };
                 LinkTypeInfo::UprobeMulti(UprobeMultiLinkInfo {
                     path_size: unsafe { s.__bindgen_anon_1.uprobe_multi.path_size },
+                    path: (path != 0)
+                        .then_some(unsafe { CStr::from_ptr(path as *const c_char).to_owned() }),
```
d-e-s-o, line 1012: I would assume that if there is a `path_size` field that the string doesn't have to be NUL terminated but length determined by the size, no?

d-e-s-o (commented):
Overall seems reasonable to me. Few minor comments.

libbpf-rs/src/query.rs
ver-nyan: Ah oops, that was meant to be path_size != 0.

libbpf-rs/src/query.rs
ver-nyan: Yea, I think `PathBuf` makes sense. At least for my usecase i'm only displaying it, so it would be more convenient since `PathBuf` has display impl but not `CString`.

PR #1391, opened 2026-05-11
Add ProgramHandle for lifetime-independent program fd ownership 

### LINKED ISSUE #920: Introduce `ProgramHandle` abstraction
We should think about introducing a `ProgramHandle` type, similar to what we have with [`MapHandle`](https://docs.rs/libbpf-rs/0.24.2/libbpf_rs/struct.MapHandle.html). We seem to have a bunch of somewhat confusing APIs that work on file descriptors, when really I think they should be abstracted behind such a type.

cong-or:
## Summary                                                                                                                                                                                  
                                                                
  I needed a way to hold onto a program fd after the Object goes out of scope. MapHandle already solves this for maps — ProgramHandle does the same thing for programs.                       
                                                                                                                                                                                              
  The implementation queries bpf_prog_info via bpf_obj_get_info_by_fd and caches name, type, tag, and id. Constructors cover the three ways you'd typically get a program: by kernel id, from 
  a pinned bpffs path, or by converting from an existing Program reference. Pin/unpin and AsFd round it out.

  Kept it scoped — no ProgramCore trait, no attach methods (those need the libbpf bpf_program pointer), no ProgramInfo renames. Those can come separately if wanted.

  Closes #920

### DIFF
--- a/libbpf-rs/src/lib.rs
+++ b/libbpf-rs/src/lib.rs
@@ -147,6 +147,7 @@ pub use crate::program::Output as ProgramOutput;
 pub use crate::program::PerfEventOpts;
 pub use crate::program::Program;
 pub use crate::program::ProgramAttachType;
+pub use crate::program::ProgramHandle;
 pub use crate::program::ProgramImpl;
 pub use crate::program::ProgramMut;
 pub use crate::program::ProgramType;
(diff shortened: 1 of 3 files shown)


danielocfb (approved):
Seems fine to me, thank you!

PR #1402, opened 2026-05-26
example: fix compilation of task_longrun

weriomat:
The following [commit](https://github.com/torvalds/linux/commit/86dd6c04ef9f213e14d60c9f64bce1cc019f816e) renamed the `scheduler_tick` function to `sched_tick`. We should probably update the example accordingly.


d-e-s-o (approved):
Looks good to me, thanks!

PR #1417, opened 2026-06-29
Support opening read-only maps

jacob-pro:
## What

Adds a new method `MapHandle::from_pinned_path_with_file_flags(path, file_flags)` that uses `bpf_obj_get_opts()`.

## Why

It is currently impossible to open a pinned map that has read-only file permissions because `bpf_obj_get()` uses the default file flags.

```
Permission denied (os error 13)
```

After this change it is possible to use:

```rust
MapHandle::from_pinned_path_with_file_flags(path, libbpf_sys::BPF_F_RDONLY)
```


d-e-s-o (approved):
Looks good to me, thanks.

PR #1447, opened 2026-08-17
examples: avoid struct sample name collision

anakryiko:
Linux has `struct sample` defined in drivers/cpufreq/intel_pstate.c, which seems to be making its way into vmlinux.h now, so to avoid name collision, rename our struct sample into struct rb_sample.

Also, while at it, add .output build directories to .gitignore under examples/ subdir.


d-e-s-o (approved):
Looks good to me.

PR #1454, opened 2026-09-03
Add doc aliases for C library types

d-e-s-o:
As of some time ago, rustdoc supports the addition of doc aliases that allow users to discover functionality via developer-defined aliases used in the search index [1]. For us this seems like very relevant functionality because in many cases users come from a C background and want to understand what Rust functionality wraps a certain concept.

Sprinkle a bunch of `#[doc(alias = "..."]` annotations for the corresponding C functionality on our types to make them discoverable more easily via the C names.

[1] https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#alias


d-e-s-o:
I don't remember where this was proposed/hinted at (but I recall it being somewhere in this repo), so I can't credit anyone anymore for the pointer

